PR #13, opened 2024-03-11
[AIDAPP-17]: Extend knowledge portal to include service request management

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-17

### Technical Description

> Introduces the ability for contacts to self service (create service requests) within the Knowledge Management Portal.

> No


app-modules/service-management/database/migrations/2023_09_01_143524_create_service_request_types_table.php
```diff
@@ -45,6 +45,8 @@ public function up(): void
             $table->uuid('id')->primary();
 
             $table->string('name');
+            $table->text('description')->nullable();
+            $table->string('icon')->nullable();
```
Orrison, line 49: This will need to be added in a separate migration. Including a `down` method and we will need to ensure that the app will not break if the code is deployed before the migration is run.

app-modules/service-management/database/migrations/2023_09_01_172133_create_service_request_form_submissions_table.php
```diff
@@ -48,6 +48,7 @@ public function up(): void
             $table->foreignUuid('service_request_priority_id')->nullable()->constrained('service_request_priorities');
             $table->string('author_id')->nullable();
             $table->string('author_type')->nullable();
+            $table->text('description')->nullable();
```
Orrison, line 51: This will need to be added in a separate migration. Including a `down` method and we will need to ensure that the app will not break if the code is deployed before the migration is run.

app-modules/service-management/database/migrations/2023_09_05_144251_create_service_requests_table.php
```diff
@@ -45,13 +45,14 @@ public function up(): void
...
             $table->string('service_request_number')->unique();
+            $table->string('title')->nullable();
             $table->string('respondent_type');
             $table->string('respondent_id');
             $table->longText('close_details')->nullable();
             $table->longText('res_details')->nullable();
 
             $table->foreignUuid('service_request_form_submission_id')->nullable()->constrained('service_request_form_submissions');
-            $table->foreignUuid('division_id')->constrained('divisions');
+            $table->foreignUuid('division_id')->nullable()->constrained('divisions');
```
Orrison, line 55: This will need to be added in a separate migration. Including a `down` method and we will need to ensure that the app will not break if the code is deployed before the migration is run.

database/seeders/NewTenantSeeder.php
```diff
@@ -93,6 +92,8 @@ public function run(): void
             // Change Request
             ChangeRequestTypeSeeder::class,
             ChangeRequestStatusSeeder::class,
+
+            ContactSeeder::class,
```
Orrison, line 96: This should be added to just the regular database seeder I think. This seeder is run when a new tenant is created in every environment including production.

Orrison (changes requested):
This looks great! We can't merge in change to existing migrations though. So any migration changes will need to be split out into their own migrations.

We also need to ensure that the app can run with/without the migration changes having happened. See Dan's most recent PRs for how he does so using Feature flagging and then a subsequent PR to remove the feature flagging later.

We could possibly get away with this last one though and just let the migrations run knowing KB portal will be borked for a moment.

database/seeders/NewTenantSeeder.php
dgoetzit: If that is the case we'll need to make some other removals from this file in Aiding App and Advising App, as there are non metadata seeders included in each.

Orrison (approved):

PR #39, opened 2024-04-09
[Chore] Check environment when syncing roles and permissions

dgoetzit:
### Ticket(s) or GitHub Issue

n/a

### Technical Description

> Because roles are not tested, we have no need to sync roles and permissions when running tests

> No


Orrison (changes requested):
Looks like there are still some test failures here.

mxm1070 (approved):

danharrin (approved):

PR #59, opened 2024-04-23
[AIDAPP-55] Enhance the way Change Management works

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-55

### Technical Description

> Change Management is now under Add-ons in global settings which will visible based on toggle is on or off.


database/settings/2024_04_23_111934_add_change_management_in_license_add_ons.php
```diff
@@ -0,0 +1,58 @@
...
+        $data['addons']['change_management'] = true;
+
+        /** Deleting existing data group */
+        $this->migrator->delete('license.data');
+
+        /** Creating License data group again with new propery with existing data */
+        $this->migrator->addEncrypted(
+            'license.data',
+            $data
+        );
```
danharrin, line 56: I believe that this can use `updateEncrypted()`

danharrin (changes requested):

database/settings/2024_04_23_111934_add_change_management_in_license_add_ons.php
mxm1070: Actually I think you can just save the value
```php
$licenseSettings = app(LicenseSettings::class);
$licenseSettings->data->addons->changeManagement = true;
$licenseSettings->save();
```

app/DataTransferObjects/LicenseManagement/LicenseAddonsData.php
```diff
@@ -50,6 +50,7 @@ public function __construct(
         public bool $knowledgeManagement,
         public bool $realtimeChat,
         public bool $mobileApps,
+        public bool $changeManagement = true,
```
mxm1070, line 53: Should this be true by default?

mxm1070 (changes requested):
Do we need a data flag for this or is it fine because the underlying functionality already exists?

app/DataTransferObjects/LicenseManagement/LicenseAddonsData.php
danharrin: I don't think you should interact with the settings class in a migration, that kinda breaks our guidelines

database/settings/2024_04_23_111934_add_change_management_in_license_add_ons.php
danharrin: I don't think you should interact with the settings class in a migration, that kinda breaks our guidelines

mxm1070: Ah, if this is intended to be data migration you think something like this?
```php
$this->migrator->updateEncrypted([
  'license.data.addons.change_management' => true,
]);
```

ankit-canyon: > Ah, if this is intended to be data migration you think something like this?
> 
> ```
> $this->migrator->updateEncrypted([
>   'license.data.addons.change_management' => true,
> ]);
> ```

HI @mxm1070 

If we update like this then it will be give error 

Spatie\LaravelSettings\Migrations\SettingsMigrator::updateEncrypted(): Argument #1 ($property) must be of type string, array given, called in /var/www/html/database/settings/2024_04_24_082011_license_settings.php on line 8

if we follow the syntax from the doc 

```
public function up(): void
{
    $this->migrator->updateEncrypted(
        'general.site_name', 
        fn(string $siteName) => return 'Space'
    );
}
```

and do something like that 

```
 $this->migrator->updateEncrypted(
            'license.data.addons.test_ankit',
            fn () => true
        );
```
        
It will also throw the same error, hence I need to use that way.

app/DataTransferObjects/LicenseManagement/LicenseAddonsData.php
ankit-canyon: > Should this be true by default?

In Ticket there is no default behave is defined for this, hence I added true but now changed it to false.

ankit-canyon:
> Do we need a data flag for this or is it fine because the underlying functionality already exists?

Yep that works ! thanks !

```
$licenseSettings = app(LicenseSettings::class);
$licenseSettings->data->addons->change_management = true;
$licenseSettings->save();
```

database/settings/2024_04_23_111934_add_change_management_in_license_add_ons.php
Orrison, line 1: I think this migration may be unneeded and can be removed.

We prefer features to be off by default unless specifically enabled. So, leaving it in its default `false` state is sufficient. Because the DTO defaults it to `false,` it will be false even if not present in the database, and it will be updated to the correct value on the next save/sync.

@danharrin and @mxm1070 sanity check me on this if you please.

Orrison (changes requested):
Looks like this is causing some test failures in CI. I assume because the tests now need to enable the feature during the test?

database/settings/2024_04_23_111934_add_change_management_in_license_add_ons.php
mxm1070: @Orrison If you don't want to enable it by default, then yeah. It will default to false, then automatically be created if enabled in the menu.

ankit-canyon:
> Looks like this is causing some test failures in CI. I assume because the tests now need to enable the feature during the test?

Yep actually what i was doing before that adding property without default value in that class hence getting the error for required parameter.  Once I set the default value it's working fine. Thanks @Orrison !

app-modules/service-management/tests/ChangeRequest/ListChangeRequestTest.php
```diff
@@ -41,5 +41,6 @@
 testResourceRequiresPermissionForAccess(
     resource: ChangeRequestResource::class,
     permissions: 'change_request.view-any',
-    method: 'index'
+    method: 'index',
+    featureProperty : 'changeManagement'
```
Orrison, line 45:
```suggestion
    featureProperty: 'changeManagement'
```

app-modules/service-management/tests/ChangeRequestStatus/ListChangeRequestStatusTest.php
```diff
@@ -41,5 +41,6 @@
 testResourceRequiresPermissionForAccess(
     resource: ChangeRequestStatusResource::class,
     permissions: 'change_request_status.view-any',
-    method: 'index'
+    method: 'index',
+    featureProperty : 'changeManagement'
```
Orrison, line 45:
```suggestion
    featureProperty: 'changeManagement'
```

app-modules/service-management/tests/ChangeRequestType/ListChangeRequestTypeTest.php
```diff
@@ -41,5 +41,6 @@
 testResourceRequiresPermissionForAccess(
     resource: ChangeRequestTypeResource::class,
     permissions: 'change_request_type.view-any',
-    method: 'index'
+    method: 'index',
+    featureProperty : 'changeManagement'
```
Orrison, line 45:
```suggestion
    featureProperty: 'changeManagement'
```

tests/Helpers/AccessControl.php
```diff
@@ -38,21 +38,35 @@
...
 
         actingAs($user)
             ->get(
                 $resource::getUrl($method)
             )->assertForbidden();
 
+        if (! is_null($featureProperty)) {
+            $settings->data->addons->$featureProperty = true;
+            $settings->save();
+        }
```
Orrison, line 68: There should be no need to reset here; correct me if I am wrong, but it should auto-roll back on test completion.

tests/Helpers/AccessControl.php
```diff
@@ -38,21 +38,35 @@
...
 use App\Models\User;
 use Illuminate\Support\Arr;
+use App\Settings\LicenseSettings;
 
 use function Pest\Laravel\actingAs;
 
 use AidingApp\Authorization\Enums\LicenseType;
 
-function testResourceRequiresPermissionForAccess(string $resource, string|array $permissions, string $method)
+function testResourceRequiresPermissionForAccess(string $resource, string|array $permissions, string $method, $featureProperty = null)
```
Orrison, line 47: How about calling this new argument `addons` instead and typing it to `?LicenseData`.

If it is not `null` then just set `$settings->data->addons` equal to it and save. Giving the dev the ability to configure all the addons at once and providing type safety.

Orrison (changes requested):

PR #61, opened 2024-04-30
[AIDAPP-58] Unable to search service requests

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-58

### Technical Description

> Unable to search in service requests module due to some casting issue of column which fixed by implicit migration.


Orrison (approved):
Though this does fix the error I think we should fix this by modifying the `respondent_id` column of the `service_requests` table to be a `uuid` data type.

I am pretty sure we can do this without issue and loss of data, but please confirm.

When we copied over from AdvisingApp we never changed this, so it was a VARCHAR to support additional models that did NOT have a `uuid` based primary key.

Because we are just using `Contact` now, which is a `uuid` based primary key we should u

PR #82, opened 2024-05-15
[Hotfix] Update tag permissions for new system.

mxm1070:


database/migrations/Concerns/CanModifyPermissions.php
```diff
@@ -75,15 +75,29 @@ public function createPermissions(array $names, string $guardName): void
...
+                    ->insert(array_map(function (string $name, string $groupName) use ($groups, $guardName): array {
+                        return [
+                            'id' => (string) Str::orderedUuid(),
+                            'group_id' => $groups[$groupName],
+                            'guard_name' => $guardName,
+                            'name' => $name,
+                            'created_at' => now(),
+                        ];
+                    }, array_keys($names), array_values($names)));
+            });
```
Orrison, line 90: Though I understand the intention here from our conversation that these would not be run multiple times, let's add protections here to filter out permissions that already exist to prevent unique constraint violations in case something weird happens.

Orrison (changes requested):
Let's revert the changes to the Trait and write the `seed_permissions_` in the current iteration for now

Orrison (approved):

PR #89, opened 2024-05-17
[AIDAPP-90] Review Open Service Requests query logic for dashboard

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-90

### Technical Description

> Added more status (except closed) for open service requests count in widget. Fix height issue in Line chart, Donut chart and they have now equal height.
______________________________________________

### DIFF
--- a/app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
+++ b/app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
@@ -1,0 +1,51 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Aiding App™ is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Aiding App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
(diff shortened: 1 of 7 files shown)


app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
```diff
@@ -0,0 +1,51 @@
...
+*/
+
+namespace AidingApp\ServiceManagement\Models\Scopes;
+
+use Illuminate\Database\Eloquent\Builder;
+
+class ClassifiedIn
+{
+    public function __construct(
+        protected array $classification,
```
Orrison, line 44: Let's make this plural, `classifications`

app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
```diff
@@ -0,0 +1,51 @@
...
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Models\Scopes;
+
+use Illuminate\Database\Eloquent\Builder;
+
+class ClassifiedIn
+{
+    public function __construct(
```
Orrison, line 43: Are we able to add a DOCBLOCK here to specify the `$classifications` parameter should be an array of classifications?

Right now I think we have two classification ENUMs `SystemChangeRequestClassification` and `SystemServiceRequestClassification`.

So we can either make the DOCBLOCK typehint something like `SystemServiceRequestClassification[]|SystemServiceRequestClassification[].
Or better yet, we make a `Classification` interface, apply it to both classification classes and then make the typehint `Classification[]`

app/Filament/Widgets/ServiceRequestWidget.php
```diff
@@ -80,7 +80,12 @@ protected function getStats(): array
 
     private function calculateOpenServiceRequestStats(Carbon $intervalStart): array
     {
-        $openStatusIds = ServiceRequestStatus::tap(new ClassifiedAs(SystemServiceRequestClassification::Open))->pluck('id');
+        $openStatusIds = ServiceRequestStatus::tap(new ClassifiedIn([
+            SystemServiceRequestClassification::Open,
+            SystemServiceRequestClassification::InProgress,
+            SystemServiceRequestClassification::Waiting,
+            SystemServiceRequestClassification::Custom,
```
Orrison, line 87: What do you think about returning this list from a static method of `SystemServiceRequestClassification` like `getUnclosedClassifications`.

That way if we ever add a new classification that would technically be a part of this list, we don't have to track down everywhere we put it.

Orrison (changes requested):
Looks great, just a few comments and notes

app/Filament/Widgets/ServiceRequestWidget.php
ankit-canyon: Yep that's been added !

Thank you.

app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
ankit-canyon: Yep that's been added !

Thank you.

app-modules/service-management/src/Models/Scopes/ClassifiedIn.php
ankit-canyon: Yep that's been added !

Thank you.

danharrin (approved):

Orrison (approved):

PR #94, opened 2024-05-28
[AIDAPP-89]: Enable file upload feature for service requests

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-89

### Technical Description

Introduced file uploads and management for service requests.

There's a couple outstanding issues:
 
A minor nit, but for the service request view page I wasn't sure how to handle files that don't generate a preview so I went with using an `IconEntry` instead. I don't love this because its looks a bit meh, but I'm not sure how else to handle this.
 
A bigger problem is warping the controller create method in a database transaction just doesn't work for some reason. If you look here: https://github.com/canyongbs/aidingapp/blob/27055434f09e242c13fe87987fcacda87f3ce57b/app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/CreateServiceRequestController.php#L108 you can see where I commented it out. Previously it worked when I wrapped it in a DB::transaction callback. But now I want to handle and return the error message instead of silently discarding it like we were doing previously it doesn't work. It doesn't throw an error, or log an error; it just does nothing.


Orrison (changes requested):
Looks great! Just a couple of comments and a few other items:
- I fixed the transaction issue. The problem was that the code returns in 2 places and you were only committing at the end of the try/catch block. First of all, commits should be within the try/catch, and you do need to make sure you commit before any return.
- Currently if I have uploaded some files to the input, then click again to upload some more files, it completely clears out what I had uploaded before. Can we change this so it just adds to it? Of course only allowing the max amount of files.
- When you have multiple files in the input how hard would it be to add an x for each one to be able to remove individual files? We can probably break that off into a different ticket but just want your opinion.

PR #101, opened 2024-06-03
Create header logo in portal page

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-108

### Technical Description

> In portal section add an media field to upload header logo.
> Create migration for field and also doe pennant/feature.

### DIFF
--- a/app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
+++ b/app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
@@ -38,7 +38,6 @@ namespace AidingApp\Portal\Filament\Pages;
 
 use App\Models\User;
 use App\Enums\Feature;
-use Laravel\Pennant\Feature as PennantFeature;
 use Filament\Forms\Get;
 use Filament\Forms\Form;
 use Filament\Pages\SettingsPage;
--- a/app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
+++ b/app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
@@ -188,7 +188,6 @@ class ManagePortalSettings extends SettingsPage
                     Section::make('Header')
                     ->schema([
                         SpatieMediaLibraryFileUpload::make('logo')
-                            ->disk('s3')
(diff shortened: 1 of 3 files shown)


app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
```diff
@@ -170,7 +173,19 @@ public function form(Form $form): Form
                             )
                             ->columnSpanFull(),
                     ])->columns(2),
-
+                Section::make('Header')
+                    ->schema([
+                        SpatieMediaLibraryFileUpload::make('logo')
+                            ->disk('s3')
```
mxm1070, line 179: I know we've done this in a bunch of places previously, but to help future clean up please remove setting the disk here so we can rely on the `FILAMENT_FILESYSTEM_DISK` env key to handle the disk instead.

mxm1070 (changes requested):
Are you intending to do the frontend work in a separate pr?

Orrison (changes requested):
Looks good, let's move forward with @mxm1070's suggestions before merging though.

app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
jayushi-canyon: @mxm1070 Correct me if I am wrong, so you are expecting changes as below?
->disk(env(FILAMENT_FILESYSTEM_DISK), 's3' )
And in env set FILAMENT_FILESYSTEM_DISK=s3

Orrison: @jayushi-canyon, no. You can just remove the `disk` method here. As Filament will just default to the default system disk. Which we currently have set to `s3`.

jayushi-canyon: @mxm1070 @Orrison Made the required changes.

Orrison (approved):

PR #110, opened 2024-06-05
Create header on portal page - frontend

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-108

### Technical Description

> Create vue component to create a header for portal frontend page. Also add logo and sign in button in the header.

### DIFF
--- a/app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalController.php
+++ b/app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalController.php
@@ -42,15 +42,23 @@ use Illuminate\Support\Facades\URL;
 use App\Http\Controllers\Controller;
 use AidingApp\Portal\Enums\PortalLayout;
 use AidingApp\Portal\Settings\PortalSettings;
+use App\Models\SettingsProperty;
 
 class KnowledgeManagementPortalController extends Controller
 {
     public function show(): JsonResponse
     {
         $settings = resolve(PortalSettings::class);
-        // dd($settings);
+        $settingsProperty = SettingsProperty::getInstance('portal.logo');
+        $logo = $settingsProperty->getFirstMedia('logo');
+
         return response()->json([
             'layout' => $settings->knowledge_management_portal_layout ?? PortalLayout::Full,
+            'header_logo' => $logo ?$logo->getTemporaryUrl(
+                expiration: now()->addMinutes(5),
+                conversionName: 'logo-height-250px',
+            ) : null,
+            'app_name' => config('app.name') ,
             'primary_color' => Color::all()[$settings->knowledge_management_portal_primary_color ?? 'blue'],
             'rounding' => $settings->knowledge_management_portal_rounding,
             'requires_authentication' => $settings->knowledge_management_portal_requires_authentication,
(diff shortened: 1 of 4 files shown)


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalController.php
```diff
@@ -48,9 +49,16 @@ class KnowledgeManagementPortalController extends Controller
     public function show(): JsonResponse
     {
         $settings = resolve(PortalSettings::class);
+        $settingsProperty = SettingsProperty::getInstance('portal.logo');
```
Orrison, line 52: Is there a particular reason we are  calling the `SettingsProperty` class here instead of putting from `$settings` above?

portals/knowledge-management/src/App.vue
```diff
@@ -85,7 +86,9 @@
     const portalRounding = ref('');
     const categories = ref({});
     const serviceRequests = ref({});
+    const headerLogo = ref({});
```
Orrison, line 89: I believe the default value here should be `''` not `{}` as this is a string not an object.

portals/knowledge-management/src/App.vue
```diff
@@ -85,7 +86,9 @@
...
     const tags = ref({});
+    const appName = ref({});
```
Orrison, line 91: I believe the default value here should be `''` not `{}` as this is a string not an object.

portals/knowledge-management/src/Components/Header.vue
```diff
@@ -0,0 +1,96 @@
...
+
+            removeToken();
+            window.location.href = response.data.redirect_url;
+        });
+    };
+</script>
+
+<template>
+    <div class="header">
+        <div v-if="requiresAuthentication || hasServiceManagement" class="columns-2 mb-1">
```
Orrison, line 76: I think we may want to apply this `v-if` only to the login/logout buttons, perhaps surrounding them in a dev applying this check.

Even if the Portal does not require auth and they do not have the service management feature I think we still want to display the logo, right?

Orrison (changes requested):

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalController.php
jayushi-canyon: @Orrison Actually I took refrenece from logo.blade file
Screenshot 2024-06-06 231540

Orrison: Got it, if you can try pulling from the `$settings` variable already set here and see if that works.

jayushi-canyon: @Orrison I tried but that gave error.

portals/knowledge-management/src/App.vue
jayushi-canyon: Done

portals/knowledge-management/src/App.vue
jayushi-canyon: Done

portals/knowledge-management/src/Components/Header.vue
jayushi-canyon: Made the required changes

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalController.php
Orrison: Okay We can leave this for now then

Orrison (approved):

PR #125, opened 2024-06-11
[AIDAPP-122]: Fix DeliveryEngagements Job

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-122

### Technical Description

> Delays the release of the DeliverEngagements job back to the queue to prevent `MaxAttemptsExceededException`

> No

### DIFF
--- a/app-modules/engagement/src/Actions/DeliverEngagements.php
+++ b/app-modules/engagement/src/Actions/DeliverEngagements.php
@@ -68,6 +68,6 @@ class DeliverEngagements implements ShouldQueue
 
     public function middleware(): array
     {
-        return [(new WithoutOverlapping(Tenant::current()->id))->releaseAfter(30)];
+        return [(new WithoutOverlapping(Tenant::current()->id))->dontRelease()->expireAfter(180)];
     }
 }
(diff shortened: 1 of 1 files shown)


mxm1070 (approved):

app-modules/engagement/src/Actions/DeliverEngagements.php
```diff
@@ -68,6 +68,6 @@ public function handle(): void
 
     public function middleware(): array
     {
-        return [new WithoutOverlapping(Tenant::current()->id)];
+        return [(new WithoutOverlapping(Tenant::current()->id))->releaseAfter(30)];
```
Orrison, line 71: I don't think this is what we want to do here. This job is dispatched onto the queue by the scheduler every minute and we don't really need a job that encounters a lock to be retried because of this.

So I propose we remove this `releaseAfter` and add the following:
- `expireAfter` with a timeout of something like 3 minutes to prevent issues if the previous job did not clear the lock.
- `dontRelease` so that a job that encounters a lock does not go back onto the queue.

What do you think?

Orrison (changes requested):

app-modules/engagement/src/Actions/DeliverEngagements.php
dgoetzit: Makes sense - this has been adjusted.

Orrison (approved):

PR #131, opened 2024-06-14
[AIDAPP-126]: Improve the boxed portal view to be more prominent

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-126

### Technical Description

>please test with hard refresh or in private window.
> Please ensure that at least one correct change type label is added to the PR. For example, if you are adding a new feature, please add the `Change Type | New Feature` label.

> A deployment step could be a command that needs to be executed or an ENV key that needs to be added, for example.
>
> If yes, please describe the deployment steps required below and apply the `Deployment Steps` label to the PR.

### DIFF
--- a/app-modules/portal/resources/views/livewire/render-knowledge-management-portal.blade.php
+++ b/app-modules/portal/resources/views/livewire/render-knowledge-management-portal.blade.php
@@ -37,7 +37,7 @@
 @endphp
 
 <div class="flex items-center justify-center">
-    <div class="w-full max-w-6xl">
+    <div class="w-full max-w-full">
         {!! resolve(GeneratePortalEmbedCode::class)->handle(PortalType::KnowledgeManagement) !!}
     </div>
 </div>
--- a/app-modules/portal/resources/views/livewire/render-knowledge-management-portal.blade.php
+++ b/app-modules/portal/resources/views/livewire/render-knowledge-management-portal.blade.php
@@ -37,7 +37,7 @@
 @endphp
 
 <div class="flex items-center justify-center">
-    <div class="w-full max-w-full">
+    <div class="w-full max-w-6xl">
         {!! resolve(GeneratePortalEmbedCode::class)->handle(PortalType::KnowledgeManagement) !!}
     </div>
 </div>
--- a/portals/knowledge-management/src/App.vue
+++ b/portals/knowledge-management/src/App.vue
@@ -362,7 +362,7 @@
 <template>
     <div
         class="font-sans bg-gray-50 min-h-screen"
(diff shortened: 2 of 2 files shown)


app-modules/portal/resources/views/livewire/render-knowledge-management-portal.blade.php
```diff
@@ -37,7 +37,7 @@
 @endphp
 
 <div class="flex items-center justify-center">
-    <div class="w-full max-w-full">
+    <div class="w-full max-w-6xl">
```
Orrison, line 40: Sorry, to be more clear about the ticket.

This should only be applied if the boxed portal view setting is the one selected. It is passed in through the API and currently applies some padding in the Vue component if it is selected I believe.

So we need to do this in the Vue component instead of what we are doing in there with padding now

Orrison (changes requested):

portals/knowledge-management/src/App.vue
```diff
@@ -362,7 +362,7 @@
 <template>
     <div
         class="font-sans bg-gray-50 min-h-screen"
-        :class="{ 'mx-0 md:mx-10 lg:mx-24': portalLayout === 'fixed' }"
+        :class="{ 'mx-0 md:mx-10 lg:mx-40': portalLayout === 'fixed' }"
```
dgoetzit, line 365: @ketan-canyon Let's make sure this is setting a max width instead of the margin. We can remove the other margins as well.

dgoetzit (changes requested):

Orrison (changes requested):
Is there still a `max-w-6xl` on the blade file? If so I think we can remove that now as these changes should handle everything we need for the max size now.

ketan-canyon:
> .



> Is there still a `max-w-6xl` on the blade file? If so I think we can remove that now as these changes should handle everything we need for the max size now.

No there is only max-w-full see below screen shot.
image

Orrison (commented):
I think this looks good, let's just get @joelicatajr's opinion.

Current we set the max-width to `4xl` in tailwind, which equates to ~`896px`. So on mobile it's full screen, tablet mostly so depending on the tablet size, then on laptops and desktops it is never more than that max width.

Below is what it looks like on my screen which is a 1920 x 1080 resolution. What do you think Joe? Good? Or bigger or smaller?

Screenshot 2024-06-18 at 12 26 39 PM

dgoetzit (approved):

Orrison (approved):

PR #134, opened 2024-06-18
[AIDAPP-130]: Improve the display of the article display page

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-130

### Technical Description

> Move the label under the title of the article.
> Include tags after the question on the Article View page.


portals/knowledge-management/src/Components/Article.vue
```diff
@@ -54,14 +54,13 @@
...
     >
-        <Tags :tags="article.tags" />
-
         <div class="flex">
             <span> {{ article.name }} </span>
             <ChevronRightIcon
                 class="opacity-0 h-5 w-5 text-primary-600 transition-all group-hover:translate-x-2 group-hover:opacity-100"
             />
         </div>
+        <Tags :tags="article.tags" class="text-xs font-medium me-2 py-0.5 rounded" />
```
Orrison, line 63: Is there a particular reason we chose to put the class changes here and not in the `Tags` component itself?

portals/knowledge-management/src/Pages/ViewArticle.vue
```diff
@@ -108,10 +108,10 @@
                         ></Breadcrumbs>
 
                         <div class="grid space-y-2">
-                            <Tags :tags="article.tags" />
                             <div class="flex flex-col gap-3">
                                 <div class="prose max-w-none">
                                     <h1>{{ article.name }}</h1>
+                                    <Tags :tags="article.tags" class="text-xs font-medium me-2 py-0.5 rounded" />
```
Orrison, line 114: Same question as above.

Orrison (changes requested):

portals/knowledge-management/src/Components/Article.vue
payal-canyon: Hi, 
Fixed.
Thank You!!

portals/knowledge-management/src/Pages/ViewArticle.vue
payal-canyon: Hi,
Fixed.
Thank you!!

portals/knowledge-management/src/Components/Tags.vue
```diff
@@ -44,7 +44,7 @@
 </script>
 
 <template>
-    <div class="flex flex-wrap gap-2">
+    <div class="flex flex-wrap gap-2 text-xs font-medium me-2 py-0.5 rounded">
         <Badge v-for="tag in tags" :key="tag.id" :value="tag.name" />
     </div>
```
Orrison, line 49: I think we may actually want to make this change in the `Badge` component right?

This `Tags` component is just to pass an array and make a bunch of tag Badges.

Orrison (changes requested):

portals/knowledge-management/src/Components/Tags.vue
payal-canyon: Hi,
Done.
Thank You!!

Orrison (commented):
This looks good, but I think we need some clarification from Product (Joe) on the rounding of the tags. Right now because we override the `rounded` class this doesn't match the Flowbite design for rounding, but instead takes on what ever rounding is defined in the Portal Settings.

Need direction on whether or not this is wanted.

Orrison:
> This looks good, but I think we need some clarification from Product (Joe) on the rounding of the tags. Right now because we override the `rounded` class this doesn't match the Flowbite design for rounding, but instead takes on what ever rounding is defined in the Portal Settings.
> 
> Need direction on whether or not this is wanted.

Resolved by Product, it is fine that the tag matched the rounding defined in the settings. This is good to go.

Orrison (approved):

PR #135, opened 2024-06-18
[AIDAPP - 128]: Improve the portal display background and sign in button

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-

### Technical Description

> Change portal header background colour and sign in button colour.


Orrison (changes requested):
I am seeing the colors come through into the portal for the button. But it seems like the rounding is not being applied to the button.

jayushi-canyon:
> I am seeing the colors come through into the portal for the button. But it seems like the rounding is not being applied to the button.

@Orrison This branch work depended on T 127, that's why I didn't share the PR for review. As I was waiting T-127 to merge.

PR #139, opened 2024-06-20
[AIDAPP-135]: Add tag for dependencies to avoid caching of assets as they change from release to release

dgoetzit:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-135

### Technical Description

> Takes the same approach as was introduced in Advising App to append the commit hash as a query parameter to the portal script.

> No


Orrison (changes requested):
This looks good, but is there anywhere else we should be applying this? Do any of the other widgets or anywhere else we output JS need it?

Orrison (approved):

PR #184, opened 2024-07-25
[AIDAPP-200]: Improve the display of charts in the Service Requests report in the Report Library & [AIDAPP-208]: Enhance and standardize the maximum height display for line charts in dashboards

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-200

### Technical Description

> Set the Donut chart and Line chart to have a maximum height of 200px and set them to full width.

> No

> No

### DIFF
--- a/app-modules/report/src/Filament/Widgets/ServiceRequestsOverTimeLineChart.php
+++ b/app-modules/report/src/Filament/Widgets/ServiceRequestsOverTimeLineChart.php
@@ -44,13 +44,9 @@ class ServiceRequestsOverTimeLineChart extends ChartReportWidget
 {
     protected static ?string $heading = 'Requests Over Time';
 
-    protected static ?string $maxHeight = '250px';
+    protected static ?string $maxHeight = '200px';
 
-    protected int | string | array $columnSpan = [
-        'sm' => 1,
-        'md' => 2,
-        'lg' => 2,
-    ];
+    protected int | string | array $columnSpan = 'full';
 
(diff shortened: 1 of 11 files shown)


app-modules/report/src/Filament/Widgets/ServiceRequestsOverTimeLineChart.php
```diff
@@ -44,13 +44,9 @@ class ServiceRequestsOverTimeLineChart extends ChartReportWidget
 {
     protected static ?string $heading = 'Requests Over Time';
 
-    protected static ?string $maxHeight = '250px';
+    protected static ?string $maxHeight = '200px';
```
Orrison, line 47: Let's copy what we did for linecharts in AdvisingApp here.

In AdvisingApp we created an abstract class specifically for LineCharts that applied the max height and set the type. Then all LineCharts in the `report` module extended it.

Orrison (changes requested):

app-modules/report/src/Filament/Widgets/ServiceRequestsOverTimeLineChart.php
Orrison, line 1: This looks great! But one final change is that you should be able to remove the `getType` method in this class. Since it is take care of, and the same, in the parent.

Orrison (changes requested):

Orrison (approved):

PR #186, opened 2024-07-26
[AIDAPP-194]: Enhance the display of assets by better utilizing the screen real estate

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-194

### Technical Description

change asset view page in two columns.

No

No

### DIFF
--- a/app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ViewAsset.php
+++ b/app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ViewAsset.php
@@ -39,6 +39,7 @@ namespace AidingApp\InventoryManagement\Filament\Resources\AssetResource\Pages;
 use Filament\Actions\EditAction;
 use Filament\Infolists\Infolist;
 use Filament\Resources\Pages\ViewRecord;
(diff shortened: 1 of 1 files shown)


app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ViewAsset.php
```diff
@@ -54,21 +53,18 @@ public function infolist(Infolist $infolist): Infolist
     {
         return $infolist
             ->schema([
-                Section::make()
```
Orrison, line 57: We don't want to remove the `Section` here as that provides the nice box section around the content.

We just need to make this section have two columns.

Orrison (changes requested):

app-modules/inventory-management/src/Filament/Resources/AssetResource/Pages/ViewAsset.php
ketan-canyon: Done

Orrison (approved):

PR #194, opened 2024-07-31
[AIDAPP-207] Enhance the knowledge management experience by improving the article listing and article properties pages

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-207

### Technical Description

> Added created_at, updated_at column in Article properties view, Change the column orders in List Knowledge Base Items table. 

> NA

> NA


app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ListKnowledgeBaseItems.php
```diff
@@ -79,26 +81,37 @@ public function table(Table $table): Table
...
+                    ->toggleable()
                     ->sortable()
                     ->formatStateUsing(fn (bool $state): string => $state ? 'Yes' : 'No'),
-                TextColumn::make('category.name')
-                    ->label('Category')
-                    ->translateLabel()
-                    ->sortable(),
-                TextColumn::make('tags.name')
-                    ->badge()
-                    ->visible(fn (): bool => Feature::active('tags')),
```
Orrison, line 101: This column was removed here, but in the acceptance criteria it is still present, just modified and moved to be the last item in the list.

Orrison (changes requested):

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ListKnowledgeBaseItems.php
ankit-canyon: Hi @Orrison ,

Tags (not default/not in filament column selector/not sortable)

But in Acceptance criteria it's mentioned as above. So by understanding that point isn't column should be removed from the list?

Orrison: Ah, that is a good point. Let me get clarity as to whether or not it should be removed.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ListKnowledgeBaseItems.php
```diff
@@ -79,25 +81,40 @@ public function table(Table $table): Table
                     ->translateLabel()
                     ->searchable()
                     ->sortable(),
+                TextColumn::make('category.name')
+                    ->label('Category')
+                    ->translateLabel()
```
Orrison, line 86: Let's remove the `translateLabel` here and everywhere else in this page. We are getting rid of that as it is not needed.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItemResource/Pages/ListKnowledgeBaseItems.php
```diff
@@ -114,6 +131,34 @@ public function table(Table $table): Table
...
+                            ->format('d-m-Y'),
+                    ])
+                    ->query(function (Builder $query, array $data): Builder {
+                        return $query
+                            ->when(
+                                $data['created_after'],
+                                fn (Builder $query, $date): Builder => $query->whereDate('created_at', '>=', $date),
+                            );
+                    }),
+                Filter::make('updated_at')
```
Orrison, line 148: Let's add a test to `app-modules/knowledge-base/tests/KnowledgeBaseArticle/ListKnowledgeBaseArticlesTest.php` for each of these filters.

Creating KB articles, triggering the filter, then confirming that you do see that ones you would expect and not the ones that you would expect to not be listed.

Orrison (changes requested):

PR #196, opened 2024-08-01
[AIDAPP-206]: Introduce a badge that displays public or internal for a knowledge base article.

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-206

### Technical Description

Introduce a badge that displays public or internal for a knowledge base article.

No

No

### DIFF
--- a/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
+++ b/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
@@ -39,9 +39,9 @@
     </x-filament::badge>
   @endif
     <x-filament::badge>
-      {{ $getRecord()?->public ?? false ? 'Public' : 'Internal' }}
+      {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
     </x-filament::badge>
-  @if(!empty($getRecord()->public))
+  @if(!empty($getRecord()['public']))
   <x-filament::badge icon="heroicon-m-clipboard">
     <button type="button" x-data x-on:click="
         window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show').'/categories/'.$getRecord()->category_id.'/articles/'.$getRecord()->id)).then(() => {
--- a/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
+++ b/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
@@ -35,20 +35,20 @@
 <div class="flex gap-2">
     @if ($getRecord()->status)
         <x-filament::badge>
-            {{ $getRecord()->status->name }}
+            {{ $getRecord()?->status?->name ?? '' }}
         </x-filament::badge>
     @endif
     <x-filament::badge>
-        {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
+        {{ ! empty($getRecord()['public']) ? 'Public' : 'Internal' }}
     </x-filament::badge>
-    @if (!empty($getRecord()['public']))
+    @if (!empty($getRecord()['public']) && !empty($getRecord()['category_id']))
         <x-filament::badge icon="heroicon-m-clipboard">
             <button
                 type="button"
                 x-data
                 x-on:click="
         window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show') . '/categories/' . $getRecord()->category_id . '/articles/' . $getRecord()->id)).then(() => {
-            $tooltip('Text copied to clipboard!', {
+            $tooltip('Copied!', {
                 theme: $store.theme,
                 timeout: 2000,
             })
--- a/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
+++ b/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
@@ -33,7 +33,30 @@
 --}}
 
 <div class="flex gap-2">
+  @if($getRecord()->status)
     <x-filament::badge>
-        {{ $getRecord()?->status?->name ?? '' }}
+        {{ $getRecord()->status->name }}
     </x-filament::badge>
+  @endif
+    <x-filament::badge>
+        {{ $getRecord()?->public ?'Public':'Internal' }}
+    </x-filament::badge>
+  @if(!empty($getRecord()->public))
+  <x-filament::badge icon="heroicon-m-clipboard">
+    <button type="button" x-data x-on:click="
+        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show').'/categories/'.$getRecord()->category_id.'/articles/'.$getRecord()->id)).then(() => {
+            $tooltip('Text copied to clipboard!', {
+                theme: $store.theme,
+                timeout: 2000,
+            })
+        }).catch(err => {
+            console.error('Failed to copy text: ', err);
+        })
+    ">
+        Copy
+    </button>
+</x-filament::badge>
+
+  @endif
 </div>
+
--- a/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
+++ b/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
@@ -33,18 +33,21 @@
 --}}
 
 <div class="flex gap-2">
-  @if($getRecord()->status)
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()->status->name }}
+        </x-filament::badge>
+    @endif
     <x-filament::badge>
-        {{ $getRecord()->status->name }}
+        {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
     </x-filament::badge>
-  @endif
-    <x-filament::badge>
-      {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
-    </x-filament::badge>
-  @if(!empty($getRecord()['public']))
-  <x-filament::badge icon="heroicon-m-clipboard">
-    <button type="button" x-data x-on:click="
-        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show').'/categories/'.$getRecord()->category_id.'/articles/'.$getRecord()->id)).then(() => {
+    @if (!empty($getRecord()['public']))
+        <x-filament::badge icon="heroicon-m-clipboard">
+            <button
+                type="button"
+                x-data
+                x-on:click="
+        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show') . '/categories/' . $getRecord()->category_id . '/articles/' . $getRecord()->id)).then(() => {
             $tooltip('Text copied to clipboard!', {
                 theme: $store.theme,
                 timeout: 2000,
@@ -52,11 +55,10 @@
         }).catch(err => {
             console.error('Failed to copy text: ', err);
         })
-    ">
-        Copy
-    </button>
-</x-filament::badge>
-
-  @endif
+    "
+            >
+                Copy
+            </button>
+        </x-filament::badge>
+    @endif
 </div>
-
--- a/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
+++ b/app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
@@ -39,7 +39,7 @@
     </x-filament::badge>
   @endif
     <x-filament::badge>
-        {{ $getRecord()?->public ?'Public':'Internal' }}
+      {{ $getRecord()?->public ?? false ? 'Public' : 'Internal' }}
     </x-filament::badge>
   @if(!empty($getRecord()->public))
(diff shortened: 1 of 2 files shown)


app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
...
 
 <div class="flex gap-2">
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()->status->name }}
+        </x-filament::badge>
+    @endif
     <x-filament::badge>
-        {{ $getRecord()?->status?->name ?? '' }}
+        {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
```
Orrison, line 42:
```suggestion
        {{ ! empty($getRecord()['public']) ? 'Public' : 'Internal' }}
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
 --}}
 
 <div class="flex gap-2">
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()->status->name }}
```
Orrison, line 38: Is status an option field right now? If so, we will probably want to keep the `$getRecord()?->status?->name ?? ''` pattern you had before to prevent an error being thrown when visiting an article that does not have a status set.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
...
+        {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
     </x-filament::badge>
+    @if (!empty($getRecord()['public']))
+        <x-filament::badge icon="heroicon-m-clipboard">
+            <button
+                type="button"
+                x-data
+                x-on:click="
+        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show') . '/categories/' . $getRecord()->category_id . '/articles/' . $getRecord()->id)).then(() => {
+            $tooltip('Text copied to clipboard!', {
```
Orrison, line 51:
```suggestion
            $tooltip('Copied!', {
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
...
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()->status->name }}
+        </x-filament::badge>
+    @endif
     <x-filament::badge>
-        {{ $getRecord()?->status?->name ?? '' }}
+        {{ $getRecord()['public'] ?? false ? 'Public' : 'Internal' }}
     </x-filament::badge>
+    @if (!empty($getRecord()['public']))
```
Orrison, line 44: I believe selecting a category is currently optional. And an article needs to be in a category and be public to show up in the portal.
So let's also check here that `$getRecord()->category_id` is also not empty or else the link would not work.

Orrison (changes requested):

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: yes, status is optional field. so i have added like you suggest. but we cannot remove this @if ($getRecord()->status) condition because if we remove that condition then it's print blank badge on screen.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: condition added. done.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
...
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()?->status?->name ?? '' }}
+        </x-filament::badge>
+    @endif
     <x-filament::badge>
-        {{ $getRecord()?->status?->name ?? '' }}
+        {{ !empty($getRecord()['public']) ? 'Public' : 'Internal' }}
     </x-filament::badge>
+    @if (!empty($getRecord()['public']) && !empty($getRecord()['category_id']))
```
danharrin, line 44: Is it possible to keep the way we access attributes on a record consistent? So as properties instead of as array keys? `$getRecord()->public`

danharrin (changes requested):

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: No, we cannot do like that
Code Checks / fix-code-style(PR check)  giving error for that because i think public is reserve keyword so.

ketan-canyon: I think now it will not give because we are using not empty i think. I write wrong syntax i think that's why it giving. below is error i got.

Error:  app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php: SyntaxError: Parse Error : syntax error, unexpected 'public' (T_PUBLIC) on line 1
Error:  > 1 | <?php echo $getRecord()?->public ?? false ? 'Public' : 'Internal' ?>

I will add now like you suggest because we are now using !empty so.

danharrin: To be honest I think `public` should not have existed as an attribute in the first place, regardless of its keyword status, it should be `is_public` to match the name of other booleans. Thoughts on the best way to correct this @orrison?

Orrison: @danharrin specifically to address whether or not we want to rename the column. I think I agree, but it would beyond the scope of this ticket to do so.
I will add a ticket to our backlog to discuss this and eventually make the change.

@ketan-canyon are you sure the lining issue is because of a reserved keyword issue? I may be wrong but I have a suspicion that this is not the case.
It is likely that linting is throwing that error because static analysis does not know what `$getRecord()` actually returns in this case.
You can fix this by setting `$getRecord()` to a variable and adding a `@var` docblock for that variable to type hint it to the correct Model.

ketan-canyon: @Orrison 
Not lint, its Code Checks / fix-code-style checking giving me syntax error. but now we are using !empty so, i will check again this and update here

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,32 @@
...
+                timeout: 2000,
+            })
+        }).catch(err => {
+            console.error('Failed to copy text: ', err);
+        })
+    "
+            >
+                Copy
+            </button>
+        </x-filament::badge>
```
Orrison, line 62: Let's have the copy icon and the click event happen actually on the public badge itself rather than a being a completely separate badge.

You can have the icon be on the right side in the badge as well by using `icon-position="after"`: https://filamentphp.com/docs/3.x/support/blade-components/badge#adding-an-icon-to-a-badge

Orrison (changes requested):

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
 --}}
 
 <div class="flex gap-2">
-    <x-filament::badge>
-        {{ $getRecord()?->status?->name ?? '' }}
-    </x-filament::badge>
+    @if ($getRecord()->status)
+        <x-filament::badge>
+            {{ $getRecord()?->status?->name ?? '' }}
```
Orrison, line 38: Because you are checking whether or not `$getRecord()->status` is truthy above you can probably simplify this:

```suggestion
            {{ $getRecord()->status->name }}
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
...
+        </x-filament::badge>
+    @endif
+    @if (!empty($getRecord()->public) && !empty($getRecord()->category_id))
```
Orrison, line 41:
```suggestion
    @if ($getRecord()->public && ! empty($getRecord()->category_id))
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
...
+        <x-filament::badge
+            icon="heroicon-m-clipboard"
+            icon-position="after"
+        >
+            <button
+                type="button"
+                x-data
+                x-on:click="
+        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show') . '/categories/' . $getRecord()->category_id . '/articles/' . $getRecord()->id)).then(() => {
```
Orrison, line 50:
```suggestion
        window.navigator.clipboard.writeText(@js(route('portal.knowledge-management.show') . '/categories/' . $getRecord()->category_id . '/articles/' . $getRecord()->getKey())).then(() => {
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
...
+            $tooltip('Copied!', {
+                theme: $store.theme,
+                timeout: 2000,
+            })
+        }).catch(err => {
+            console.error('Failed to copy text: ', err);
+        })
+    "
+            >
+                {{ !empty($getRecord()->public) ? 'Public' : 'Internal' }}
```
Orrison, line 60: You have already confirmed that this is public above, so this can be simplified:

```suggestion
                Public
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
...
+            </button>
+        </x-filament::badge>
+    @else
+        <x-filament::badge>
+            {{ !empty($getRecord()->public) ? 'Public' : 'Internal' }}
```
Orrison, line 65: You have already confirmed that this is not public above, so this can be simplified:

```suggestion
            Internal
```

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -33,7 +33,36 @@
...
+        <x-filament::badge>
+            {{ $getRecord()?->status?->name ?? '' }}
+        </x-filament::badge>
+    @endif
+    @if (!empty($getRecord()->public) && !empty($getRecord()->category_id))
+        <x-filament::badge
+            icon="heroicon-m-clipboard"
+            icon-position="after"
+        >
+            <button
```
Orrison, line 46: Would it work to move this `button` tag to surround the badge? So the entire badge is clickable?

Orrison (changes requested):
Looks good, just a few notes to cleanup.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

Orrison (approved):

PR #204, opened 2024-08-06
[AIDAPP-211]: Refactor the primary url of the application portal

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-211

### Technical Description

Refactor the primary url of the application portal

yes, run below commands,
npm run build
if any error will come because of tip top editor then first run composer ci  and then npm run build again.

No

### DIFF
--- a/app-modules/portal/routes/web.php
+++ b/app-modules/portal/routes/web.php
@@ -41,7 +41,7 @@ use AidingApp\Portal\Http\Middleware\EnsureKnowledgeManagementPortalIsEnabled;
 use AidingApp\Portal\Http\Middleware\EnsureKnowledgeManagementPortalIsEmbeddableAndAuthorized;
 use AidingApp\Portal\Http\Controllers\KnowledgeManagementPortal\KnowledgeManagementPortalAuthenticateController;
 
-Route::prefix('portals')
+Route::prefix('portal')
     ->name('portal.')
     ->middleware([
         'web',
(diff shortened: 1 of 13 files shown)


joelicatajr (changes requested):
@ketan-canyon @Orrison AIDAPP-211 calls for refactoring the baseUrl to `/portal` rather than `/portals`.

ketan-canyon:
> @ketan-canyon @Orrison AIDAPP-211 calls for refactoring the baseUrl to `/portal` rather than `/portals`.

Done.. @joelicatajr

app-modules/portal/routes/web.php
```diff
@@ -41,30 +41,35 @@
...
     ->middleware([
         'web',
         EnsureFrontendRequestsAreStateful::class,
     ])
     ->group(function () {
+        Route::get('/knowledge-management/{any?}', function ($any = null) {
+            $url = $any ? '/portal/' . $any : '/portal';
+
+            return redirect($url, 301);
+        })->where('any', '.*');
```
Orrison, line 55: I believe this is going to need to be moved out of this prefix group and target `/portals/knowledge-management/{any?}` now that we have changed the prefix above.

Without doing so, redirects from the old URL structure do not work.

Orrison (changes requested):

app-modules/portal/routes/web.php
Orrison: Took care of this in [bf6a574](https://github.com/canyongbs/aidingapp/pull/204/commits/bf6a574a7a04a891890c3587ed8c2705c0ef9316)

Orrison (approved):

PR #220, opened 2024-08-23
[AIDAPP-222] Attempting to search by entering in a phrase followed by the return key while in the client portal does not work as expected.

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-222

### Technical Description

> Fixed the issue when Attempting to search by entering in a phrase followed by the return key while in the client portal does not work as expected.

> No

> No

### DIFF
--- a/portals/knowledge-management/src/Pages/Home.vue
+++ b/portals/knowledge-management/src/Pages/Home.vue
@@ -138,41 +138,39 @@
                     <p class="text-primary-100">Search our knowledge base for advice and answers</p>
                 </div>
 
-                <form action="#" method="GET">
-                    <label for="search" class="sr-only">Search</label>
-                    <div class="relative rounded">
-                        <div>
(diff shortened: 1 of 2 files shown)


Orrison (changes requested):
Looks good. But this change needs to be done everywhere in the Portal where there is a search header.

I believe another place is on the category page. Please fix it there and then confirm there is no where else that it is present that needs to be updated.

ankit-canyon:
> Looks good. But this change needs to be done everywhere in the Portal where there is a search header.
> 
> I believe another place is on the category page. Please fix it there and then confirm there is no where else that it is present that needs to be updated.

Yep ! you're right ! apart from these two pages there is no search bar exist on any other page of portal.

Orrison (approved):
Great work!

PR #234, opened 2024-09-06
[AIDAPP-237]: Introduce automatic auditing on newly created data models

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-237

### Technical Description

> Auto auditing has been implemented.

> No.

>No.


app-modules/audit/src/Settings/AuditSettings.php
Orrison, line 1: I believe the `HasDefaultSettings` interface and `defaults` method are not needed here and can be removed and deleted as we changed to a better default handling that works with the defaults of the parameter.

Try that and see if it solves the issue with the test. If not let me know and I can take a look further.

Orrison (changes requested):

app-modules/audit/src/Settings/AuditSettings.php
kandarp-canyon: Hi @Orrison , I've pushed the changes after removing the HasDefaultSettings interface and defaults method, but the test error is still there. can you please check?

kandarp-canyon: @Orrison Please check and provide your valuable feedback.

app-modules/audit/tests/AuditListenerTest.php
```diff
@@ -39,24 +39,26 @@
...
+    $auditSettings->audited_models_exclude = [$serviceRequest->getMorphClass()];
 
     $auditSettings->save();
 
-    $serviceRequest = ServiceRequest::factory()->create();
-
     expect($serviceRequest->audits)->toHaveCount(0);
 
-    $auditSettings->audited_models[] = $serviceRequest->getMorphClass();
+    $auditSettings->audited_models_exclude[] = [];
```
Orrison, line 55: This should be resetting it to a blank array, not setting a new value as a blank array

```suggestion
    $auditSettings->audited_models_exclude = [];
```

Orrison (changes requested):

Orrison (approved):

PR #237, opened 2024-09-11
[AIDAPP-255]: Introduce the ability to insert hyperlinks in the GDPR compliant notice on the portal

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-255

### Technical Description

> Added Tiptap Editor for GDPR Banner Text.

> No

> Yes. GDPRNewBannerText


danharrin (changes requested):
Can you explain the new DTO and the migration to move to that DTO please?

Orrison (changes requested):
I agree with @danharrin that I am not sure there is a need for this DTO. Especially since there is only one value.

This needs to be converted and stored as the TipTap JSON format.

@danharrin I am thinking we can make exceptions for the data migration here to allow for the usage of TipTap helpers/classes to do this conversion. What do you think?

danharrin:
Maybe, but I don't think the TipTap helper is required since it can be wrapped in a single content object which is a very predictable structure `['type' => 'doc', 'content' => [['type' => 'text' => 'text' => 'Content here']]]` if I remember correctly?

payal-canyon:
> Can you explain the new DTO and the migration to move to that DTO please?

Hi @danharrin
Yes, you're right. Earlier, I was adding the cookie banner text and assumed it wouldn't work with the key. However, I've now managed to get it working without the DTO.

payal-canyon:
> Maybe, but I don't think the TipTap helper is required since it can be wrapped in a single content object which is a very predictable structure `['type' => 'doc', 'content' => [['type' => 'text' => 'text' => 'Content here']]]` if I remember correctly?

The response is something different. 
`[
    "type" => "doc",
    "content" => [
        [
            "type" => "paragraph",
            "attrs" => [
                "textAlign" => "start"
            ],
            "content" => [
                [
                    "type" => "text",
                    "text" => "content here"
                ]
            ]
        ]
    ]
]`

app-modules/portal/database/migrations/2024_09_11_120345_data_migrate_gdpr_banner_text.php
```diff
@@ -0,0 +1,108 @@
...
+
+            DB::commit();
+        } catch (Exception $error) {
+            DB::rollBack();
+
+            throw $error;
+        }
+    }
+
+    public function down(): void
```
Orrison, line 83: This wouldn't actually work in Production as it wouldn't set the value back to what it was if it had been modified from the default. But because we do not use the `downs` in Production and I want to get rid of them soon this should be fine for now

app-modules/portal/src/Filament/Pages/ManagePortalSettings.php
```diff
@@ -219,4 +228,30 @@ public function form(Form $form): Form
...
+            $settings->cookie_gdpr_banner_text = $data['cookie_gdpr_banner_text'];
+        }
+
+        $settings->fill($data);
+        $settings->save();
+
+        $this->callHook('afterSave');
+
+        $this->getSavedNotification()?->send();
+    }
```
Orrison, line 256: Is this needed? The `cookie_gdpr_banner_text` field would not be present if the Feature Flag was not active so it wouldn't be present in the data.

resources/views/vendor/cookie-consent/dialog.blade.php
```diff
@@ -42,10 +43,16 @@
             <div class="flex items-center justify-between flex-wrap">
                 <div class="w-0 flex-1 items-center hidden md:inline">
                     <p class="ml-3 text-primary-950 text-sm font-medium cookie-consent__message">
-                        @if(! empty(app(PortalSettings::class)->gdpr_banner_text))
-                            {!! str(app(PortalSettings::class)->gdpr_banner_text)->sanitizeHtml() !!}
+                        @if(FeatureFlag::GDPRNewBannerText->active())
+                            {!! tiptap_converter()->asHtml(app(PortalSettings::class)->cookie_gdpr_banner_text) !!}
```
Orrison, line 47: Do we need to consider sanitizing this HTML? I forget if the `tiptap_converter` `asHtml` method does that for you.

app-modules/portal/src/Settings/PortalSettings.php
```diff
@@ -65,6 +65,8 @@ class PortalSettings extends SettingsWithMedia
 
     public string $gdpr_banner_text = "We use cookies to personalize content, to provide social media features, and to analyze our traffic. We also share information about your use of our site with our partners who may combine it with other information that you've provided to them or that they've collected from your use of their services.";
 
+    public array $cookie_gdpr_banner_text = [];
```
Orrison, line 68: Can you explain the reasoning behind a new property rather than modifying the existing one? Is it possible to use the same one? Perhaps by temporarily changing the type hint to be `string|array`.

Also is it possible to make the default for this one here include the content?

Orrison (changes requested):

Orrison (changes requested):
This looks great, but on the frontend the links in the cookie banner are the same color as the text so it is hard for me to tell what is a link.

Can we make the link the primary color of the portal?

payal-canyon:
> This looks great, but on the frontend the links in the cookie banner are the same color as the text so it is hard for me to tell what is a link.
> 
> Can we make the link the primary color of the portal?

Hi @Orrison 
Yes. I set the primary color with the underline effect on hover.
Thank You!

Orrison (approved):

PR #238, opened 2024-09-12
[AIDAPP-250]: Track portal views of articles and display them as article metadata

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-250

### Technical Description

> Create new feature where every time the article is clicked for view it count increases and end user can see total number of view on that article in the view article page.

> No.

> `portal_view_count`

### DIFF
--- a/app-modules/knowledge-base/src/Models/KnowledgeBaseItem.php
+++ b/app-modules/knowledge-base/src/Models/KnowledgeBaseItem.php
@@ -55,61 +55,62 @@ use AidingApp\Audit\Models\Concerns\Auditable as AuditableTrait;
  */
 class KnowledgeBaseItem extends BaseModel implements Auditable, HasMedia, HasTags
 {
-    use AuditableTrait;
-    use HasUuids;
-    use InteractsWithMedia;
-    use SoftDeletes;
-    use InteractsWithTags;
+  use AuditableTrait;
+  use HasUuids;
+  use InteractsWithMedia;
+  use SoftDeletes;
+  use InteractsWithTags;
 
-    protected $table = 'knowledge_base_articles';
+  protected $table = 'knowledge_base_articles';
 
-    protected $casts = [
-        'public' => 'boolean',
-        'article_details' => 'array',
-    ];
+  protected $casts = [
+    'public' => 'boolean',
+    'article_details' => 'array',
+  ];
 
-    protected $fillable = [
-        'article_details',
-        'category_id',
-        'notes',
-        'public',
-        'quality_id',
-        'status_id',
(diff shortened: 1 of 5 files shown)


app-modules/knowledge-base/src/Models/KnowledgeBaseItem.php
```diff
@@ -71,6 +71,7 @@ class KnowledgeBaseItem extends BaseModel implements Auditable, HasMedia, HasTag
     protected $fillable = [
         'article_details',
         'category_id',
+        'portal_view_count',
```
Orrison, line 74: Let's actually make the choice not to have this fillable. Since we should only ever be incrementing this by 1.

That will mean you will need to increment the value and save it. But that technically is safer

portals/knowledge-management/src/Pages/ViewArticle.vue
```diff
@@ -61,6 +61,8 @@
     const loading = ref(true);
     const category = ref(null);
     const article = ref(null);
+    const portalViewCount = ref(0);
+    const portalViewCountFlag = ref(0);
```
Orrison, line 65: This probably makes more sense to be `ref(false)`

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalArticleController.php
```diff
@@ -47,6 +48,12 @@ class KnowledgeManagementPortalArticleController extends Controller
 {
     public function show(KnowledgeBaseCategory $category, KnowledgeBaseItem $article): JsonResponse
     {
+        $portalViewCountFlag = Feature::active('portal_view_count');
+
+        if ($portalViewCountFlag) {
+            $article->update(['portal_view_count' => $article->portal_view_count + 1]);
```
Orrison, line 54: We should use `increment` instead here: https://laravel.com/docs/11.x/queries#increment-and-decrement

`increment` is more performant and is an atomic operation that will prevent issues if multiple people are loading the page simultaneously.

jayushi-canyon: Done

app-modules/knowledge-base/src/Models/KnowledgeBaseItem.php
jayushi-canyon: Done

Orrison (approved):

PR #248, opened 2024-09-19
[AIDAPP-183]: Improve the underlying method used for rendering Service Request lock icon

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-183

### Technical Description

> Locked service request icon will now show using filament render hook.

> No

> No

### DIFF
--- a/app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
+++ b/app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2024, Canyon GBS LLC. All rights reserved.
+
+    Aiding App™ is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS LLC respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS™ and Aiding App™ are registered trademarks of
+      Canyon GBS LLC, and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS LLC.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 namespace AidingApp\ServiceManagement\Filament\Concerns;
 
 use Filament\View\PanelsRenderHook;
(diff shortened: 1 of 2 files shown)


app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
```diff
@@ -0,0 +1,52 @@
...
+
+namespace AidingApp\ServiceManagement\Filament\Concerns;
+
+use Filament\View\PanelsRenderHook;
+use Illuminate\Contracts\View\View;
+use Filament\Support\Facades\FilamentView;
+
+trait ServiceRequestLocked
+{
+    public function boot()
```
Orrison, line 45: Rather than completely overwritting `boot` here, let's use the Livewire trait hook syntax: https://livewire.laravel.com/docs/lifecycle-hooks#using-hooks-inside-a-trait

So this method can be `bootServiceRequestLocked`.

app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
```diff
@@ -0,0 +1,52 @@
...
+    {
+        FilamentView::registerRenderHook(
+            PanelsRenderHook::PAGE_HEADER_ACTIONS_AFTER,
+            fn (): View => view('filament.pages.service-request-lock-icon'),
+        );
```
Orrison, line 50: Just to be extra safe, let's scope this render hook to the specific page using `static::class`

https://filamentphp.com/docs/3.x/support/render-hooks#scoping-render-hooks

Orrison (changes requested):

app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
```diff
@@ -0,0 +1,60 @@
...
+    public function bootServiceRequestLocked()
+    {
+        FilamentView::registerRenderHook(
+            PanelsRenderHook::PAGE_HEADER_ACTIONS_AFTER,
+            fn (): View => view('filament.pages.service-request-lock-icon'),
+            scopes: [
+                ViewServiceRequest::class,
+                ManageAssignments::class,
+                ManageServiceRequestUpdate::class,
+            ],
```
Orrison, line 57: Does applying `static::class` not work here? That would be better so that we don't have to keep track of every page we use this Trait on.

resources/views/filament/pages/service-request-lock-icon.blade.php
```diff
@@ -0,0 +1,47 @@
...
+    use AidingApp\ServiceManagement\Enums\SystemServiceRequestClassification;
+@endphp
+
+@if ($this->getRecord()?->status?->classification === SystemServiceRequestClassification::Closed)
+    <x-filament::icon-button
+        data-identifier="service_request_closed"
+        icon="heroicon-m-lock-closed"
+        color="gray"
+        size="lg"
+        tooltip="This service request is locked as status is closed."
```
Orrison, line 45:
```suggestion
        tooltip="This service request is locked as the status is {$this->getRecord()->status}."
```

Orrison (changes requested):

app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
```diff
@@ -0,0 +1,53 @@
...
+use Filament\Support\Facades\FilamentView;
+
+trait ServiceRequestLocked
+{
+    public function bootServiceRequestLocked()
+    {
+        FilamentView::registerRenderHook(
+            PanelsRenderHook::PAGE_HEADER_ACTIONS_AFTER,
+            fn (): View => view('filament.pages.service-request-lock-icon'),
+            scopes: $this->getRenderHookScopes(),
```
Orrison, line 50: @danharrin, what are your thoughts on using this instead of just `static::class`?

I can see that this method seems to provide an array with `static::class` for pages, but it also contains a reference to the Resource.

Orrison (changes requested):

app-modules/service-management/src/Filament/Concerns/ServiceRequestLocked.php
payal-canyon: Yep. Applied the static::class in the recent changes.

PR #249, opened 2024-09-20
[AIDAPP-263]: Enhance the portal header by including the ability to search no matter what page of the portal the user is on

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-263

### Technical Description

> Create Global search bar for portal view that when clicked enter will redirect to home page and trigger the search bar on home page.

> No.

> No.


portals/knowledge-management/src/Components/Header.vue
```diff
@@ -32,11 +32,12 @@
 </COPYRIGHT>
 -->
 <script setup>
-    import { defineProps } from 'vue';
+    import { defineProps, ref, onMounted } from 'vue';
```
Orrison, line 35: Do you need these imports for `ref` and `onMounted` in this component?

portals/knowledge-management/src/Stores/globalState.js
Orrison, line 1: Can you explain the reasoning behind making this search value a global state store and binding that as the Model in the `GlobalSearchBar` component, as compared to just having a local `ref` value in the component?

portals/knowledge-management/src/Pages/Home.vue
```diff
@@ -87,11 +90,25 @@
...
+            searchQuery.value = route.query.search;
+        }
+    });
+
+    watch(
+        () => route.query.search,
+        (newVal, oldVal) => {
+            searchQuery.value = newVal || '';
+        },
+    );
```
Orrison, line 110: Do we need the `onMounted` call here? I feel that this may cause the same issue we had with the page view count where this is triggered twice.

If the reason is so that you can clear the search input, I think you should be fine doing that locally within the component after the route push

portals/knowledge-management/src/Components/GlobalSearchBar.vue
```diff
@@ -0,0 +1,56 @@
...
+    };
+</script>
+
+<template>
+    <div class="float-right mr-3">
+        <form @submit.prevent="onSearch">
+            <input
+                v-model="globalSearchQuery"
+                type="text"
+                placeholder="search knowledge portal"
```
Orrison, line 51:
```suggestion
                placeholder="Search knowledge portal"
```

Orrison (changes requested):
Had a few notes. Also, it may be helpful UX-wise if within the search param watch if it changed to put the focus on the main page search input. If that proves more complicated than say 30 minutes of RnD then nevermind. But worth looking into

portals/knowledge-management/src/Stores/globalState.js
jayushi-canyon: @Orrison Once the search is done on the home page and if you redirect to any other page the global search bar still remains filled therefore to empty the value in global search bar I need to access global variable since local does not work in other components. That's why i created global state.

portals/knowledge-management/src/Pages/Home.vue
jayushi-canyon: @Orrison No It would never trigger twice as onMounted only run once when the component is rendered while watch runs when the value in this case 'route.query.search' changes.

jayushi-canyon:
> Had a few notes. Also, it may be helpful UX-wise if within the search param watch if it changed to put the focus on the main page search input. If that proves more complicated than say 30 minutes of RnD then nevermind. But worth looking into

@Orrison Sorry didn't exactly got your point, Could you please enhance it a little.

portals/knowledge-management/src/Pages/Home.vue
Orrison: @jayushi-canyon got it. Though it may not be triggered twice. I am pretty sure the `onMounted` is redundant here, so it would be best to remove it.

Orrison:
> > Had a few notes. Also, it may be helpful UX-wise if within the search param watch if it changed to put the focus on the main page search input. If that proves more complicated than say 30 minutes of RnD then nevermind. But worth looking into
> 
> @Orrison Sorry didn't exactly got your point, Could you please enhance it a little.

Sure, I mean that it would be nice that once you did a global search your browser focus (i.e. the input you are targeted into) is on the main search bar of the page, instead of still in the global search bar.

portals/knowledge-management/src/Components/GlobalSearchBar.vue
jayushi-canyon: Done

jayushi-canyon:
> > > Had a few notes. Also, it may be helpful UX-wise if within the search param watch if it changed to put the focus on the main page search input. If that proves more complicated than say 30 minutes of RnD then nevermind. But worth looking into
> > 
> > 
> > @Orrison Sorry didn't exactly got your point, Could you please enhance it a little.
> 
> Sure, I mean that it would be nice that once you did a global search your browser focus (i.e. the input you are targeted into) is on the main search bar of the page, instead of still in the global search bar.

Made the respective changes

portals/knowledge-management/src/Pages/Home.vue
```diff
@@ -68,6 +70,8 @@
     const loadingResults = ref(false);
     const searchResults = ref(null);
     const selectedTags = ref([]);
+    const route = useRoute();
+    const globalSearchInput = ref(null);
```
Orrison, line 74: I am really confused as to how this is working lol. But this isn't the correct type of `ref` I was talking about for this.

We should change to make sure we are using the right thing: https://vuejs.org/guide/essentials/template-refs.html

Orrison (changes requested):

portals/knowledge-management/src/Pages/Home.vue
```diff
@@ -145,6 +168,7 @@
                             <MagnifyingGlassIcon class="h-5 w-5 text-gray-400" aria-hidden="true" />
                         </div>
                         <input
+                            ref="globalSearchInput"
```
jayushi-canyon, line 171: @Orrison The example that you shared I am using that ref over here.

jayushi-canyon:
@Orrison Product has made changes in the requirement of this ticket. Therefore I have made the changes in the files based on new requirements

portals/knowledge-management/src/Pages/Home.vue
Orrison: Ah I see, this is a different type of way to do this. So many ways to do the same thing in Vue lol. All good

Orrison (approved):

portals/knowledge-management/src/Pages/Home.vue
Orrison: @jayushi-canyon yeah I see now, I see that this is just a different way to do what I was thinking. There are so many different ways to do things in Vue lol

PR #250, opened 2024-09-20
[AIDAPP-260]: Product Ping Health check is failing for all environments

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-260

### Technical Description

> Update the App URL in the PingCheck class instead of checking the IP address (1.1.1.1).

> Might need to restart the docker container as docker-compose.dev.yml file is change

> No

### DIFF
--- a/app/Providers/HealthServiceProvider.php
+++ b/app/Providers/HealthServiceProvider.php
@@ -61,7 +61,7 @@ class HealthServiceProvider extends ServiceProvider
     public function boot(): void
     {
         $local = app()->isLocal() && str(config('app.url'))->contains(['localhost', '.local']);
-        $secure = config('app.force_https') ? 'https://' : 'http://';
+        $parsed_url = parse_url(config('app.url'));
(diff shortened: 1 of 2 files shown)


app/Providers/HealthServiceProvider.php
```diff
@@ -76,7 +77,7 @@ public function boot(): void
                 ->unless($local || app()->environment('staging')),
             // cloudflare dns
             PingCheck::new()
-                ->url('1.1.1.1')
+                ->url(str_replace($secure, '', config('app.url')))
```
Orrison, line 80: Rather than checking `app.force_https` and doing this string replace you could just get the URL without the schema by grabbing it's parts out of https://www.php.net/manual/en/function.parse-url.php

Orrison (changes requested):

Orrison (approved):

PR #253, opened 2024-09-24
[AIDAPP-231]: Introduce the ability to detect an attempted login from an approved organization domain, and collect the requisite information to create a contact and log the user in

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-231

### Technical Description

Creates a process to detect if there is an Organization with a domain that matches the email address of a individual logging into the Portal that does not yet have a Contact record, collects details, and creates a Contact record for them.

No.

No.

### DIFF
--- a/app-modules/portal/src/Notifications/AuthenticatePortalNotification.php
+++ b/app-modules/portal/src/Notifications/AuthenticatePortalNotification.php
@@ -63,9 +63,11 @@ class AuthenticatePortalNotification extends BaseNotification implements EmailNo
 
     public function identifyRecipient(): array
     {
+        $contact = $this->authentication->educatable;
+
         return [
-            $this->authentication->educatable->getKey(),
-            $this->authentication->educatable->getMorphClass(),
+            ! empty($contact) ? $contact->getKey() : null,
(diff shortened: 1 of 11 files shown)


app-modules/portal/src/Notifications/AuthenticatePortalNotification.php
```diff
@@ -63,9 +63,12 @@ public function toEmail(object $notifiable): MailMessage
...
     {
+        /** @var ?Contact $contact */
+        $contact = $this->authentication->educatable;
+
         return [
-            $this->authentication->educatable->getKey(),
-            $this->authentication->educatable->getMorphClass(),
+            ! empty($contact) ? $contact->getKey() : null,
+            ! empty($contact) ? $this->authentication->educatable->getMorphClass() : 'anonymous',
         ];
```
danharrin, line 72: How about refactoring this into an early return for `if (! $contact)` and return a different array instead of using two ternaries

danharrin (changes requested):

danharrin (approved):

PR #256, opened 2024-09-25
[AIDAPP-217]: Separate Public or Internal and Copy badges on Knowledge Base Article details

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-217

### Technical Description

Separate Public or Internal and Copy badges on Knowledge Base Article details

No

No


Orrison (changes requested):
Let's have this just be the stand alone copy icon rather than a tag/button

ketan-canyon:
> Let's have this just be the stand alone copy icon rather than a tag/button

Done. copy text removed. but as mention in ticket description we need separate badge for copy so we need button.

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
```diff
@@ -39,7 +39,11 @@
         </x-filament::badge>
     @endif
     @if ($getRecord()->public && !empty($getRecord()->category_id))
+        <x-filament::badge>
+            Public
+        </x-filament::badge>
         <button
+            class="flex items-center justify-center"
```
Orrison, line 46: This doesn't need to be a button and shouldn't so that we can get rid of the button style and border around the icon. The copy on click functionality does not come from it being a button. It comes from the `x-on:click`. You can just get rid of the button HTML and put the onclick on the x-filament::badge directly

Orrison (changes requested):

app-modules/knowledge-base/resources/views/filament/pages/badges.blade.php
ketan-canyon: Done

Orrison (approved):
@ketan-canyon I made a couple of adjustment to just have this be the icon rather than have the badge border which is what I meant. And also adjusted the AlpineJS to better align in the file.

Other than that this looks good.

Please re-take a screenshot and add it to the ticket.

PR #260, opened 2024-09-30
[AIDAPP-283]: Update the display of the portal when in boxed layout mode to use more of the screen.

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-283

### Technical Description

> Removed the space before and after the content for the all screen resolutions.

> No.

> No.


Orrison (changes requested):
Hmm, based on the code here and the screenshot I don't think this is quite right. This should be increasing the size of the content area when it is set to Fixed-Width (Boxed Layout)

Orrison (approved):

PR #290, opened 2024-10-18
[AIDAPP-306]: Introduce search result pagination when using the client portal

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-306

### Technical Description

> Introduce search result pagination when using the client portal.

> No.

> No.


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalSearchController.php
```diff
@@ -63,31 +62,30 @@ public function get(Request $request): KnowledgeManagementSearchData
                 fn (Stringable $string) => $string->explode(',')
             );
 
-        $itemData = KnowledgeBaseArticleData::collection(
+        $itemData =
```
Orrison, line 65: I don't believe you need to return the DTO here, I think you can pass a paginator into a collection for a DTO: https://spatie.be/docs/laravel-data/v4/as-a-resource/from-data-to-resource#content-paginators

Orrison (changes requested):

Orrison (approved):

PR #305, opened 2024-10-29
[AIDAPP-304]: Update the database configuration for KB article categories to ensure all categories contain a URL slug

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-304

### Technical Description

Update the database configuration for KB article categories to ensure all categories contain a URL slug

No

Yes
Data Migration : 
aidingapp/app-modules/knowledge-base/database/migrations/2024_10_29_183334_data_add_remainig_slug_in_knowledge_base_categories_table.php


app-modules/knowledge-base/database/migrations/2024_10_29_183334_data_add_remainig_slug_in_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,81 @@
...
+                    $category->slug = $slug;
+                    $category->save();
+
+                    DB::commit();
+
+                    break;
+                } catch (QueryException $e) {
+                    DB::rollBack();
+
+                    if ($e->errorInfo[1] === 7) {
```
Orrison, line 67: What is the purpose of this check? Is it because you are catching `QueryException` here and this is some error info code you are targeting?

I would recommend instead of doing this, to specifically catch a `Illuminate\Database\UniqueConstraintViolationException`

app-modules/knowledge-base/database/migrations/2024_10_29_183334_data_add_remainig_slug_in_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,81 @@
...
+                        $attempts++;
+                        $slug = $baseSlug . '-' . $attempts;
+                    } else {
+                        throw $e;
+                    }
+                }
+            }
+
+            if ($attempts == 15) {
+                Log::info("Failed to generate a unique slug after 15 attempts for category ID {$category->name}\n");
```
Orrison, line 77: Let's throw an exception here, this will need to totally fail if there is a problem.

Orrison (changes requested):
A couple of comments.

Also this is missing the second migration where we make the column non-nullable.

ketan-canyon:
> A couple of comments.
> 
> Also this is missing the second migration where we make the column non-nullable.

Migration added for column non-nullable.

app-modules/knowledge-base/database/migrations/2024_10_29_183334_data_add_remainig_slug_in_knowledge_base_categories_table.php
ketan-canyon: Done

app-modules/knowledge-base/database/migrations/2024_10_30_184250_update_slug_column_to_not_nullable_in_knowledge_base_categories_table.php
Orrison, line 1: This cannot execute before your data migration. That will result in a database error being thrown as there may be items where the column is `null`

Move this to go after the data migration.

Orrison (changes requested):

app-modules/knowledge-base/database/migrations/2024_10_30_184250_update_slug_column_to_not_nullable_in_knowledge_base_categories_table.php
ketan-canyon: Done

Orrison (approved):

PR #327, opened 2024-11-19
[AIDAPP-353]: Update routes to use category slugs for articles and categories with fall back

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-353

### Technical Description

Set up routing for Article categories based on slugs rather than ID, but it will still redirect to the correct slug URL if an ID URL is accessed for backward compatibility.

No

No

### DIFF
--- a/app-modules/portal/routes/api.php
+++ b/app-modules/portal/routes/api.php
@@ -97,7 +97,7 @@ Route::prefix('api')
                 Route::get('/categories', [KnowledgeManagementPortalCategoryController::class, 'index'])
                     ->name('category.index');
 
-                Route::get('/categories/{category}', [KnowledgeManagementPortalCategoryController::class, 'show'])
+                Route::get('/categories/{category:slug}', [KnowledgeManagementPortalCategoryController::class, 'show'])
                     ->name('category.show');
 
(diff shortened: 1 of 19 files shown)


app-modules/portal/routes/api.php
```diff
@@ -97,11 +101,40 @@
...
+                    ->name('category.show')
+                    ->missing(function (Request $request) {
+                        throw_if(! str()->isUuid($request->category), ModelNotFoundException::class);
 
-                Route::get('/categories/{category}/articles/{article}', [KnowledgeManagementPortalArticleController::class, 'show'])
-                    ->name('article.show');
+                        $category = KnowledgeBaseCategory::findOrFail($request->category);
+
+                        return redirect()->route('api.portal.category.show', $category->slug);
+                    });
```
danharrin, line 112: Never seen this method used before, nice!

Wondering if we could pass an instance of an invokable class instead of a closure (or convert it to a closure, either `app(Class::class)(...)` or `app(Class::class)->execute(...)`), to keep the logic out of the routes file?

Also applies below

danharrin (changes requested):

app-modules/portal/routes/api.php
Orrison: Done

danharrin (approved):

PR #333, opened 2024-11-22
[AIDAPP-354]: Introduce the Assignment Type Round Robin so that all new service requests are assigned in round robin fashion to applicable managers

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-354

### Technical Description

> Introduce the Assignment Type Round Robin so that all new service requests are assigned in round robin fashion to applicable managers

> No.

> Feature Flag: RounRobinId


app-modules/service-management/src/DataTransferObjects/ServiceRequestDataObject.php
```diff
@@ -0,0 +1,75 @@
...
+            'status_id' => $this->status_id,
+            'type_id' => $this->type_id,
+            'priority_id' => $this->priority_id,
+            'title' => $this->title,
+            'close_details' => $this->close_details,
+            'res_details' => $this->res_details,
+            'respondent_type' => $this->respondent_type,
+            'respondent_id' => $this->respondent_id,
+        ];
+    }
```
danharrin, line 74: Maybe I am wrong, but since we're not doing transformation here do we need this method? Because I think that the base `Data` object has it

And what is the difference between `?string` and `string|Nullable` in this case?

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
danharrin, line 1: Could you please add a new test for the round robin functionality, starting with 3 managers and auto-assigning the first, then the second, then the third, then back to the first?

danharrin (changes requested):

jayushi-canyon:
@danharrin I am deleting this PR because when i took pull from main all the new changes have came into my branch

PR #336, opened 2024-12-01
[AIDAPP-354]: Introduce the Assignment Type Round Robin so that all new service requests are assigned in round robin fashion to applicable managers

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-354

### Technical Description

> Introduce the Assignment Type Round Robin so that all new service requests are assigned in round robin fashion to applicable managers

> No.

> Feature Flag: RoundRobinId


app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,74 @@
...
+use App\Features\RoundRobinId;
+use Illuminate\Database\Eloquent\Builder;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+
+class RoundRobinAssigner implements ServiceRequestTypeAssigner
+{
+    public function execute(ServiceRequest $serviceRequest): void
+    {
+        if ($serviceRequest->priority?->type && RoundRobinId::active()) {
+            $serviceRequestType = $serviceRequest->priority->type;
```
Orrison, line 49: You could make this much more terse by assigning `$serviceRequestType` outside of the if statement first and then checking that that is `! is_null()`

app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,74 @@
...
+            if ($lastAsignee) {
+                $user = User::query()->whereRelation('teams.managableServiceRequestTypes', 'service_request_types.id', $serviceRequestType->getKey())
+                    ->where('name', '>=', $lastAsignee->name)
+                    ->where(fn (Builder $query) => $query
+                        ->where('name', '!=', $lastAsignee->name)
+                        ->orWhere('users.id', '>', $lastAsignee->id))
+                    ->orderBy('name')->orderBy('id')->first();
+            }
+
+            if ($user == null) {
```
Orrison, line 63: @jayushi-canyon here and going forward please do not use loose comparison unless there is actually a need for it. Please use strict comparison by default. i.e. `===`

app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,74 @@
...
+                $user = User::query()->whereRelation('teams.managableServiceRequestTypes', 'service_request_types.id', $serviceRequestType->getKey())
+                    ->orderBy('name')->orderBy('id')->first();
+            }
```
Orrison, line 66: Rather than making use of a `$user` variable you set to `null` and carry down the chain. You could just have a single User::query() and use `when()` to modify the query based on `$lastAsignee` being present.

app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,74 @@
...
+
+            if ($user != null) {
```
Orrison, line 68: Please do not use loose comparisons. This should be `$user !== null`. Though actually in most placed in the application we have done `! is_null()` but there is some debate that technically the comparison is fractions of microseconds faster so I am fine with it.

Thoughts @danharrin?

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+            'respondent_id' => Contact::factory()->create()->getKey(),
+        ])
+        ->call('create');
+
+    $serviceRequest = ServiceRequest::first();
 
     expect($serviceRequest->assignments()->first())->user->id->toBe($user->getKey());
 });
+
+test('check round robin assignment', function () {
```
Orrison, line 488:
```suggestion
test('assignment type round robin will auto-assign to new service requests', function () {
```

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+    User::factory()->licensed(LicenseType::cases())->create();
+    $team = Team::factory()->create();
+
+    $serviceRequestTypesWithManager = ServiceRequestType::factory()
```
Orrison, line 492:
```suggestion
    $serviceRequestTypeWithManager = ServiceRequestType::factory()
```

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+            'assignment_type' => ServiceRequestTypeAssignmentTypes::RoundRobin,
+        ])
+        ->create();
+
+    $request = collect(CreateServiceRequestRequestFactory::new()->create([
+        'priority_id' => ServiceRequestPriority::factory()->create([
+            'type_id' => $serviceRequestTypesWithManager->getKey(),
+        ])->getKey(),
+    ]));
+    $users = User::orderBy('name')->orderBy('id')->get();
```
Orrison, line 507: You really should create a group of Users assigned to the manager team. I am confused as to where they are being generated now other than one at the top of the test and the others maybe from some of the factories above? You want to be more explicit than that though.

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+
+    foreach ($users as $user) {
+        $user->grantLicense(LicenseType::RecruitmentCrm);
+        $user->teams()->attach($team);
+        $user->givePermissionTo('service_request.view-any');
+        $user->givePermissionTo('service_request.create');
+        $user->refresh();
+        actingAs($user);
```
Orrison, line 515: You don't need to setup the Users with access or permissions or anything. The creation can and should be done probably from like a single super admin User. Since the assignment has nothing to do with whether or not the User is the one logged in.

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+            type_id: $request['type_id'],
+            priority_id: $request['priority_id'],
+            title: $request['title'],
+            close_details: $request['close_details'],
+            res_details: $request['res_details'],
+            respondent_type: $request['respondent_type'],
+            respondent_id: $request['respondent_id'],
+        );
+
+        app(CreateServiceRequestAction::class)->execute($serviceRequestDataObject);
```
Orrison, line 534: I am not sure why you are calling the component `create` bother here and then also executing the Action. The Action should be trigger in the component.

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+        $getServiceRequestType = ServiceRequestType::where('assignment_type', 'round-robin')->first();
+        expect($getServiceRequestType->round_robin_last_assigned_id)->ToBe($user->getKey());
```
Orrison, line 536: You should also be testing the assignement of the new Service Request.

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +467,93 @@
...
+        title: $request['title'],
+        close_details: $request['close_details'],
+        res_details: $request['res_details'],
+        respondent_type: $request['respondent_type'],
+        respondent_id: $request['respondent_id'],
+    );
+
+    app(CreateServiceRequestAction::class)->execute($serviceRequestDataObject);
+    $getServiceRequestType = ServiceRequestType::where('assignment_type', 'round-robin')->first();
+    expect($getServiceRequestType->round_robin_last_assigned_id)->ToBe(User::orderBy('name')->orderBy('id')->first()->getKey());
```
Orrison, line 558: You could do like a while loop for the total amount of users plus one instead of this, hard coded.

Orrison (changes requested):
@danharrin please review as well when you have a moment.

app-modules/service-management/src/Services/ServiceRequestType/IndividualAssigner.php
```diff
@@ -0,0 +1,66 @@
...
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+use AidingApp\Notification\Events\TriggeredAutoSubscription;
+use AidingApp\ServiceManagement\Enums\ServiceRequestAssignmentStatus;
+
+class IndividualAssigner implements ServiceRequestTypeAssigner
+{
+    public function execute(ServiceRequest $serviceRequest): void
+    {
+        $user = auth()->user();
+        $serviceRequest->save();
```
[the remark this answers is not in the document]
jayushi-canyon: @Orrison Because the assigner gets call before service request is created unlike observer 'created' method. So to access relationship of service request i need it to be saved. Let me know what you think

app-modules/service-management/src/Services/ServiceRequestType/IndividualAssigner.php
```diff
@@ -0,0 +1,66 @@
...
+
+            if ($manager) {
+                $serviceRequest->assignments()->create([
+                    'user_id' => $manager->getKey(),
+                    'assigned_by_id' => $user->getKey(),
+                    'assigned_at' => now(),
+                    'status' => ServiceRequestAssignmentStatus::Active,
+                ]);
+            }
+        }
```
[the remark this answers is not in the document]
jayushi-canyon: @Orrison You mean like this: 'assigned_by_id' => auth()->user() ? auth()->user()->getKey() : null,

app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
jayushi-canyon: @Orrison If i do so then the condition where we reach the last user and now we need to fetch the first user will never be matched.

`if ($user === null) {
        $user = User::query()->whereRelation('teams.managableServiceRequestTypes', 'service_request_types.id', $serviceRequestType->getKey())
          ->orderBy('name')->orderBy('id')->first();
      }`

app-modules/service-management/src/Services/ServiceRequestType/IndividualAssigner.php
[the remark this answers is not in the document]
Orrison: Got it, it looks like we save this in the main Action now. So I don't think this is needed anymore right?

app-modules/service-management/src/Services/ServiceRequestType/RoundRobinAssigner.php
```diff
@@ -0,0 +1,81 @@
...
+use AidingApp\ServiceManagement\Enums\ServiceRequestAssignmentStatus;
+
+class RoundRobinAssigner implements ServiceRequestTypeAssigner
+{
+    public function execute(ServiceRequest $serviceRequest): void
+    {
+        $serviceRequestType = $serviceRequest->priority->type;
+
+        if (! is_null($serviceRequestType) && RoundRobinId::active()) {
+            $lastAsignee = $serviceRequestType->lastAssignedUser;
```
Orrison, line 52:
```suggestion
            $lastAssignee = $serviceRequestType->lastAssignedUser;
```

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +465,74 @@
...
+        livewire(CreateServiceRequest::class)
+            ->fillForm($request->toArray())
+            ->fillForm([
+                'respondent_id' => Contact::factory()->create()->getKey(),
+            ])
+            ->call('create')
+            ->assertHasNoFormErrors();
+
+        $latestServiceRequest = ServiceRequest::latest()->first();
+        $getServiceRequestType = ServiceRequestType::where('assignment_type', 'round-robin')->first();
```
Orrison, line 520:
```suggestion
        $getServiceRequestType = ServiceRequestType::where('assignment_type', ServiceRequestTypeAssignmentTypes::RoundRobin->value)->first();
```

app-modules/service-management/tests/ServiceRequest/CreateServiceRequestTest.php
```diff
@@ -464,11 +465,74 @@
...
+        expect($getServiceRequestType->assignment_type)->toBe(ServiceRequestTypeAssignmentTypes::RoundRobin);
+        expect($getServiceRequestType->fresh()->last_assigned_id)->ToBe($user->getKey());
```
Orrison, line 522: Does this fail of the `fresh()` is not used here? You just retrieved this from the database on line `520` and made no changes since. So that shouldn't be needed.

Orrison (changes requested):
Looks good! Just a couple of small tweaks

Orrison (approved):

PR #353, opened 2024-12-11
[AIDAPP-355]: Develop internal APIs to support interoperability with partner management system in order to support utilization metrics

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-355

### Technical Description

Develop internal APIs to support interoperability with partner management system in order to support utilization metrics

No

No


app/Http/Controllers/UtilizationMetricsApisController.php
Orrison, line 1: We should drop the plural here for the API part on the file and class name. So `UtilizationMetricsApiController` instead of `UtilizationMetricsApisController`

app/Http/Controllers/UtilizationMetricsApisController.php
```diff
@@ -0,0 +1,71 @@
...
+                'changes' => ChangeRequest::count(),
+                'knowledge_base_articles' => KnowledgeBaseItem::count(),
+                'tasks' => Task::count(),
+            ];
+
+            return response()->json([
+                'data' => $metrics,
+            ], 200);
+        } catch (Exception $e) {
+            return response()->json([
```
Orrison, line 67: Before returning we should `report()` the Exception here so that is sent to our error reporting system.

app/Http/Controllers/UtilizationMetricsApisController.php
```diff
@@ -0,0 +1,71 @@
...
+                'users' => User::count(),
+                'service_requests' => ServiceRequest::count(),
+                'assets' => Asset::count(),
+                'changes' => ChangeRequest::count(),
+                'knowledge_base_articles' => KnowledgeBaseItem::count(),
+                'tasks' => Task::count(),
+            ];
+
+            return response()->json([
+                'data' => $metrics,
```
Orrison, line 64: We can just inline the data here rather than define this unneeded `$metrics` variable.

tests/Unit/UtilizationMetricsApisTest.php
```diff
@@ -0,0 +1,145 @@
...
+use function Pest\Laravel\withoutMiddleware;
+
+use AidingApp\InventoryManagement\Models\Asset;
+use AidingApp\KnowledgeBase\Models\KnowledgeBaseItem;
+use AidingApp\ServiceManagement\Models\ChangeRequest;
+use App\Multitenancy\Http\Middleware\CheckOlympusKey;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+
+it('checks the API returns users', function () {
+    withoutMiddleware(CheckOlympusKey::class);
```
Orrison, line 52: Because every single test here calls `withoutMiddleware` here you could pull this out and put it in a `beforeEach`: https://pestphp.com/docs/hooks#content-beforeeach

Orrison (changes requested):

tests/Unit/UtilizationMetricsApisTest.php
Orrison, line 1: For each test where it makes sense, if the Model has soft deletes it think it may also be a good idea to create an extra record that is soft deleted.

So we can make sure that the API does not return soft deleted data when appropriate.

Orrison (changes requested):

app/Http/Controllers/UtilizationMetricsApisController.php
ketan-canyon: Done

app/Http/Controllers/UtilizationMetricsApisController.php
ketan-canyon: Done

app/Http/Controllers/UtilizationMetricsApisController.php
ketan-canyon: Done

tests/Unit/UtilizationMetricsApisTest.php
ketan-canyon: Done

tests/Unit/UtilizationMetricsApisTest.php
ketan-canyon: Done

Orrison (approved):

PR #361, opened 2024-12-13
[AIDAPP-386] Fix issue when attempting to associate a contact to an organization, an exception is generated

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-386

### Technical Description

> Fix issue when attempting to associate a contact to an organization, an exception is generated.

> No.

> No.


Orrison (changes requested):
Though this does fix the issue. Let's actually rename the relationship as well in the `Contact` Model.

Right now it is `organizations` plural. But it is a `BelongsTo` so there can only be one.

So let's change it to `organization`. Making sure that where ever that is used/referenced for `Contact` that we update it.

Orrison (commented):
Please re-request review once you have tested the other item with Portal registration

Orrison (approved):

PR #365, opened 2024-12-17
[AIDAPP-373]: Display sub categories in the portal after navigating to the main category

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-373

### Technical Description

> Display sub categories in the portal after navigating to the main category

> No

> No

### DIFF
--- a/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
+++ b/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
@@ -46,7 +46,6 @@ class KnowledgeBaseCategoryData extends Data
         public string $name,
         public ?string $description,
         public ?string $icon,
-        /** @var null|KnowledgeBaseCategoryData */
         public ?KnowledgeBaseCategoryData $parentCategory,
         /** @var null|Collection<int, KnowledgeBaseCategoryData> */
         public ?Collection $subCategories,
--- a/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
+++ b/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
@@ -45,6 +45,5 @@ class KnowledgeBaseCategoryData extends Data
         public string $name,
         public ?string $description,
         public ?string $icon,
-        public ?string $parent_id,
     ) {}
 }
--- a/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
+++ b/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
@@ -36,9 +36,8 @@
 
 namespace AidingApp\Portal\DataTransferObjects;
 
-use Spatie\LaravelData\Data;
 use Illuminate\Support\Collection;
-use Illuminate\Database\Eloquent\Model;
+use Spatie\LaravelData\Data;
 
 class KnowledgeBaseCategoryData extends Data
 {
@@ -47,7 +46,9 @@ class KnowledgeBaseCategoryData extends Data
         public string $name,
         public ?string $description,
         public ?string $icon,
-        public ?Model $parentCategory,
+        /** @var null|KnowledgeBaseCategoryData */
+        public ?KnowledgeBaseCategoryData $parentCategory,
+        /** @var null|Collection<int, KnowledgeBaseCategoryData> */
         public ?Collection $subCategories,
     ) {}
 }
--- a/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
+++ b/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
@@ -37,6 +37,8 @@
 namespace AidingApp\Portal\DataTransferObjects;
 
 use Spatie\LaravelData\Data;
+use Illuminate\Support\Collection;
+use Illuminate\Database\Eloquent\Model;
 
 class KnowledgeBaseCategoryData extends Data
 {
@@ -45,5 +47,7 @@ class KnowledgeBaseCategoryData extends Data
         public string $name,
         public ?string $description,
         public ?string $icon,
+        public ?Model $parentCategory,
+        public ?Collection $subCategories,
     ) {}
 }
--- a/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
+++ b/app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
@@ -45,5 +45,6 @@ class KnowledgeBaseCategoryData extends Data
         public string $name,
         public ?string $description,
         public ?string $icon,
+        public ?string $parent_id,
     ) {}
 }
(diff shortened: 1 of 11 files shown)


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalArticleController.php
```diff
@@ -76,6 +76,7 @@ public function show(KnowledgeBaseCategory $category, KnowledgeBaseItem $article
                 'name' => $category->name,
                 'description' => $category->description,
             ]),
+            'parentCategory' => $category->parentCategory,
```
Orrison, line 79: It may be best if this here and in other places was a part of the `KnowledgeBaseCategoryData` used in the above `category` item rather than stand alone.

portals/knowledge-management/src/portal.js
```diff
@@ -70,6 +70,11 @@ customElements.define(
                         name: 'view-category',
                         component: ViewCategory,
                     },
+                    {
+                        path: baseUrl + '/subcategories/:subCategorySlug',
```
Orrison, line 74: What do you think about this being related to the parent category. So `baseUrl + '/categories/:parentCategorySlug/:categorySlug'`.

This way may make more sense for the pathing

portals/knowledge-management/src/Pages/ViewCategory.vue
```diff
@@ -193,22 +210,31 @@
...
 
-                category.value = response.data.category;
-                articles.value = response.data.articles.data;
-                setPagination(response.data.articles);
-                loadingResults.value = false;
-            },
-        );
+            category.value = response.data.category;
+            subCategories.value = response.data.subCategories;
+            parentCategory.value = response.data.parentCategory;
```
Orrison, line 233: Rather than having all these seperate values for `subCategories` and `parentCategory` I think it may be easier to parse if that data is just stored as optional/nullable values in the `category`. And then just use it / modify the experience if they are present.

portals/knowledge-management/src/Pages/ViewCategory.vue
```diff
@@ -193,22 +210,31 @@
             debounceSearch(searchQuery.value, page);
             return;
         }
+
         loadingResults.value = true;
 
         const { get } = consumer();
+        const slug = route.params.categorySlug ?? route.params.subCategorySlug;
```
Orrison, line 217: If you make the route change I suggest I think you can get rid of this.

portals/knowledge-management/src/Components/SubCategories.vue
```diff
@@ -0,0 +1,96 @@
...
+        subCategories: {
+            type: Object,
+            required: true,
+        },
+    });
+</script>
+
+<template>
+    <div>
+        <h3 class="text-xl">Article Subcategories</h3>
```
Orrison, line 47:
```suggestion
        <h3 class="text-xl">Subcategories</h3>
```

Orrison (changes requested):

app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
```diff
@@ -45,5 +47,7 @@ public function __construct(
         public string $name,
         public ?string $description,
         public ?string $icon,
+        public ?Model $parentCategory,
```
Orrison, line 50: You should be able to type hint the actual Model of Knowledge Category rather than the base Model. Though you may want to consider making this another `KnowledgeBaseCategoryData` so that you only pass it the data needed rather than the entire Model.

app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
```diff
@@ -45,5 +47,7 @@ public function __construct(
...
+        public ?Collection $subCategories,
```
Orrison, line 51: You should add an annotation here. Something like:

```php
/** @var null|Collection<int, KnowledgeBaseCategoryData> */
```

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalArticleController.php
```diff
@@ -75,6 +75,7 @@ public function show(KnowledgeBaseCategory $category, KnowledgeBaseItem $article
                 'slug' => $category->slug,
                 'name' => $category->name,
                 'description' => $category->description,
+                'parentCategory' => $category->parentCategory,
```
Orrison, line 78: You may want to consider making this a `KnowledgeBaseCategoryData` item as well so that only the needed data is passed.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalCategoryController.php
```diff
@@ -72,6 +73,20 @@ public function show(KnowledgeBaseCategory $category): JsonResponse
...
+                    ->with(['parentCategory:id,name,slug'])
+                    ->get()
+                    ->map(function (KnowledgeBaseCategory $subCategory) {
+                        return [
+                            'slug' => $subCategory->slug,
+                            'name' => $subCategory->name,
+                            'description' => $subCategory->description,
+                            'icon' => $subCategory->icon ? svg($subCategory->icon, 'h-6 w-6')->toHtml() : null,
+                            'parentCategory' => $subCategory->parentCategory,
+                        ];
```
Orrison, line 88: You could map this into another `KnowledgeBaseCategoryData` item

Orrison (changes requested):

app-modules/portal/src/DataTransferObjects/KnowledgeBaseCategoryData.php
```diff
@@ -45,5 +46,9 @@ public function __construct(
         public string $name,
         public ?string $description,
         public ?string $icon,
+        /** @var null|KnowledgeBaseCategoryData */
```
Orrison, line 49: This annotation shouldn't be needed. Just the one below for the collection so that you can define what it is a collection of. But this definition is automatically inferred from the type hinting.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalArticleController.php
```diff
@@ -75,6 +75,11 @@ public function show(KnowledgeBaseCategory $category, KnowledgeBaseItem $article
                 'slug' => $category->slug,
                 'name' => $category->name,
                 'description' => $category->description,
+                'parentCategory' => $category?->parentCategory ? KnowledgeBaseCategoryData::from([
+                    'slug' => $category?->parentCategory->slug,
+                    'name' => $category?->parentCategory->name,
+                    'description' => $category?->parentCategory->description,
```
Orrison, line 81: I don't think you should have to null safe `$category` here as there is not a scenario in this Controller that it would be `null`.

portals/knowledge-management/src/portal.js
```diff
@@ -70,6 +70,11 @@ customElements.define(
                         name: 'view-category',
                         component: ViewCategory,
                     },
+                    {
+                        path: baseUrl + '/categories/:parentCategorySlug/:categorySlug',
```
Orrison, line 74: When I navigate to a SubCategory page I am currently not seeing this URL structure, even in incognito mode. Are you experiencing the same?

Orrison (changes requested):

Orrison (approved):

PR #379, opened 2025-01-03
[AIDAPP-398]: Introduce the ability for managers and auditors of service requests to control notifications and emails related emails by service request type

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-398

### Technical Description

Introduces matrix for managing notification preferences. Sends notifications based on service request events.

Fixes misc issue with chat notification icon autofocusing.

No

`ServiceRequestClosedNotificationColumns`


app-modules/service-management/src/Filament/Resources/ServiceRequestTypeResource/Pages/EditServiceRequestTypeNotifications.php
```diff
@@ -62,128 +61,72 @@ public function form(Form $form): Form
...
-                        Toggle::make('is_customers_service_request_closed_email_enabled')
-                            ->label('Email'),
-                        Toggle::make('is_customers_service_request_closed_notification_enabled')
-                            ->label('Notification'),
-                    ]),
+                        ViewField::make('settings')
+                            ->rules(['array'])
+                            ->view('service-management::filament.resources.service-request-type-resource.pages.edit-service-request-type-notifications.matrix'),
+                    ])
+                    ->visible(ServiceRequestClosedNotificationColumns::active())
```
Orrison, line 71: I think this is fine since no one is currently actively using this. But in the future we would have to make sure this was more backwards compatible.

app-modules/service-management/src/Observers/ServiceRequestObserver.php
```diff
@@ -68,11 +74,20 @@ public function created(ServiceRequest $serviceRequest): void
...
+        app(NotifyServiceRequestUsers::class)->execute(
+            $serviceRequest,
+            app(ServiceRequestCreated::class, ['serviceRequest' => $serviceRequest, 'channel' => EmailChannel::class]),
+            $serviceRequest->priority?->type->is_managers_service_request_created_email_enabled ?? false,
+            $serviceRequest->priority?->type->is_auditors_service_request_created_email_enabled ?? false,
+        );
+
+        app(NotifyServiceRequestUsers::class)->execute(
+            $serviceRequest,
+            app(ServiceRequestCreated::class, ['serviceRequest' => $serviceRequest, 'channel' => DatabaseChannel::class]),
```
Orrison, line 87: Why resolve the notification from the container? Unless we feel this will benefit us enough I would rather we new up so that we have proper type hinting and static analysis of the constructor arguments.

app-modules/service-management/database/migrations/2025_01_03_123821_rename_service_request_resolved_columns_to_closed_on_service_request_types_table.php
```diff
@@ -0,0 +1,61 @@
...
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('service_request_types', function (Blueprint $table) {
+            $table->renameColumn('is_managers_service_request_resolved_email_enabled', 'is_managers_service_request_closed_email_enabled');
+            $table->renameColumn('is_managers_service_request_resolved_notification_enabled', 'is_managers_service_request_closed_notification_enabled');
+            $table->renameColumn('is_auditors_service_request_resolved_email_enabled', 'is_auditors_service_request_closed_email_enabled');
+            $table->renameColumn('is_auditors_service_request_resolved_notification_enabled', 'is_auditors_service_request_closed_notification_enabled');
```
Orrison, line 48: In the future renames like this will need to be done more carefully with a transaction and feature flag update inline here and making sure the Model can handle both values.

But given that this is an unused feature. We are fine.

app-modules/service-management/src/Notifications/ServiceRequestAssigned.php
```diff
@@ -0,0 +1,97 @@
...
+
+    /**
+     * @return array<int, string>
+     */
+    public function via(object $notifiable): array
+    {
+        return [match ($this->channel) {
+            DatabaseChannel::class => 'database',
+            EmailChannel::class => 'mail',
+        }];
```
Orrison, line 73: Hmm, I wonder if perhaps we want to consider `channel` becoming `channels` which accepts an array of channels?

That way you only have to dispatch a single notification that creates jobs for the correct channels for you.

Orrison (changes requested):

app-modules/service-management/src/Notifications/ServiceRequestAssigned.php
danharrin: A user could be a manager, or an auditor, or both, you would then have to do logic to determine which groups they are part of and dispatch a different notification object based on that. If you check out the code that queries the managers or auditors I think this is the most efficient way to avoid that, so you don't have to combine the query results. What do you think?

app-modules/service-management/src/Observers/ServiceRequestObserver.php
danharrin: I try and use the container everywhere so objects (even notifications) can be mocked or replaced in tests in the future if we need them, but I will change this now.

Orrison: Yeah, I think that approach makes sense. But I think for now it works for things like Notifications and Jobs to be instantiated.

Most of what we do is Unit testing or short scope Feature testing so most tests include ensuring a Job/notification WAS dispatched then a job/notification has it's own test.

app-modules/service-management/src/Notifications/ServiceRequestAssigned.php
Orrison: Hmm, I see your point but let's chat about it all a little more. Perhaps at office hours.

Orrison (approved):

PR #384, opened 2025-01-13
[AIDAPP-405]: Remove all legacy engagement capabilities from the product

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-405

### Technical Description

> Remove all legacy engagement capabilities from the product

> No

> No


Orrison (changes requested):
As discussed in office hours. EngagementDeliveryMethod needs to be modified to remove SMS and data migrations need to be written.

Orrison (changes requested):
It looks like you did not test your migration with any sort of test data at all because it fails and immediately causes a foreign key constraint.

In the very new future it will become a requirement for all data migrations to have tests. But in the meantime please do at least try and test with some dummy data locally.

Orrison (approved):

amit-canyon:
> It looks like you did not test your migration with any sort of test data at all because it fails and immediately causes a foreign key constraint.
> 
> In the very near future it will become a requirement for all data migrations to have tests. But in the meantime please do at least try and test with some dummy data locally.

Got it, and Sorry for this i will take care of this in future.

PR #401, opened 2025-01-29
[AIDAPP-161]: Upgrade the core application to run on the latest PHP framework Laravel 11

danharrin:
Signed-off-by: Dan Harrin <dan.harrin@canyongbs.com>

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-161

### Technical Description

Upgrades to Laravel 11. Also has to upgrade Spatie Permissions (which was already upgraded in Advising App to resolve performance issues) to remain compatible.

No

No


app-modules/authorization/src/Models/Role.php
```diff
@@ -81,7 +80,7 @@ public function scopeWeb(Builder $query): void
     public function scopeSuperAdmin(Builder $query): void
     {
         $query
-            ->where('name', Authenticatable::SUPER_ADMIN_ROLE)
+            ->where('name', 'SaaS Global Admin')
```
Orrison, line 83: I think this should still be and use the const set by `Authenticatable`. I believe that is what we do in Advising App as well.

Orrison (changes requested):

app-modules/authorization/src/Models/Role.php
danharrin: Updated

Orrison (approved):

PR #403, opened 2025-01-31
[AIDAPP-436]: Intermittent error occurs when clicking the "Assign to Me" button in service request assignments

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-436

### Technical Description

Intermittent error occurs when clicking the "Assign to Me" button in service request assignments

No

No

### DIFF
--- a/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
+++ b/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
@@ -80,7 +80,7 @@ class AssignedToRelationManager extends RelationManager
             ->paginated(false)
             ->headerActions([
                 Action::make('assign-to-me')
-                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()))
+                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()) && is_null($this->getOwnerRecord()->assignedTo))
                     ->label('Assign To Me')
                     ->color('gray')
                     ->requiresConfirmation()
--- a/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
+++ b/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
@@ -80,13 +80,16 @@ class AssignedToRelationManager extends RelationManager
             ->paginated(false)
             ->headerActions([
                 Action::make('assign-to-me')
-                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()) && is_null($this->getOwnerRecord()->assignedTo))
+                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()) && is_null($this->getOwnerRecord()->assignedTo) && in_array(auth()->user()?->getKey(), $this->getOwnerRecord()->priority->type?->managers
+                        ->flatMap(fn ($managers) => $managers->users)
+                        ->pluck('id')
+                        ->toArray()))
                     ->label('Assign To Me')
                     ->color('gray')
                     ->requiresConfirmation()
                     ->action(fn (array $data) => $this->getOwnerRecord()->assignments()->create([
-                        'user_id' => auth()->user()?->id,
-                        'assigned_by_id' => auth()->user()->id ?? null,
+                        'user_id' => auth()->user()?->getKey(),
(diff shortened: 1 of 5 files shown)


Orrison (changes requested):
@ketan-canyon as you mentioned there is still work to do on this one. Please re-request review when ready.

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
```diff
@@ -52,7 +52,7 @@ class ServiceRequestAssignmentObserver
 {
     public function creating(ServiceRequestAssignment $serviceRequestAssignment): void
     {
-        throw_if(! auth()->user()->can('update', $serviceRequestAssignment->serviceRequest), new AttemptedToAssignNonManagerToServiceRequest());
+        throw_if(! auth()->user()->can('update', $serviceRequestAssignment->serviceRequest) && ! $serviceRequestAssignment->user()->can('update', $serviceRequestAssignment->serviceRequest), new AttemptedToAssignNonManagerToServiceRequest());
```
Orrison: I don't think I am following the logic of this change here. Can you please explain how this fixes the issue and what the issue is?

Orrison (changes requested):

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
ketan-canyon: Hi @Orrison 
Actually issue is in ServiceRequestAssignmentObserver. we are not skipping admin(SaaS Global Admin) from condition in observer means we are not allowing admin (SaaS Global Admin) to do assignment of service request but for visibility we are allowing admin(SaaS Global Admin) so observer throwing error so I have made same condition which we are using in visibility of this action.

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
```diff
@@ -52,10 +52,7 @@ class ServiceRequestAssignmentObserver
 {
     public function creating(ServiceRequestAssignment $serviceRequestAssignment): void
     {
-        throw_if(! in_array($serviceRequestAssignment->user_id, $serviceRequestAssignment->serviceRequest->priority->type?->managers
-            ->flatMap(fn ($managers) => $managers->users)
-            ->pluck('id')
-            ->toArray()), new AttemptedToAssignNonManagerToServiceRequest());
+        throw_if(! auth()->user()->can('update', $serviceRequestAssignment->serviceRequest) && ! $serviceRequestAssignment->user->can('update', $serviceRequestAssignment->serviceRequest), new AttemptedToAssignNonManagerToServiceRequest());
```
Orrison, line 55: After thinking about this more. I think that this needs to be reverted back to the way it was.

We should avoid `user` checks in an observer as having this in her effects the programmatic creation of this Model.

And I will explain in another comment how we should fix the problem

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
```diff
@@ -80,7 +80,7 @@ public function table(Table $table): Table
             ->paginated(false)
             ->headerActions([
                 Action::make('assign-to-me')
-                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()))
+                    ->visible(fn () => auth()->user()->can('update', $this->getOwnerRecord()) && is_null($this->getOwnerRecord()->assignedTo))
```
Orrison, line 83: I think this issue should be fixed here. The change you have already made is good. But we need to add on a third condition, totaling the three following:

1. The User needs to be able to update the record (already here)
2. `assignedTo` needs to be `null` (already here)
3. The User needs to be a Manager of this Service Requests type

That third option not being here is what is breaking this. Because even though a Super Admin CAN edit this record, they shouldn't be able to be assigned.
So the error is actually correct and valid.
We just need to add this condition so that people like Super Admins and others do not see this action.

Orrison (changes requested):

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
ketan-canyon: Done

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/RelationManagers/AssignedToRelationManager.php
ketan-canyon: Done

Orrison (approved):

PR #409, opened 2025-02-07
[AIDAPP-448]: Optimize display when viewing a service request

kandarp-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-448

### Technical Description

> Optimize display when viewing a service request.

> No.

> No.

### DIFF
--- a/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
+++ b/app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
@@ -92,10 +92,10 @@ class ViewServiceRequest extends ViewRecord
                             ->label('Type'),
                         TextEntry::make('close_details')
                             ->label('Description')
-                            ->columnSpan(1),
+                            ->columnSpan(2),
                         TextEntry::make('res_details')
                             ->label('Internal Details')
-                            ->columnSpan(1),
+                            ->columnSpan(2),
                         TextEntry::make('respondent')
                             ->label('Related To')
                             ->color('primary')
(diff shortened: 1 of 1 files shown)


app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
```diff
@@ -92,10 +92,10 @@ public function infolist(Infolist $infolist): Infolist
                             ->label('Type'),
                         TextEntry::make('close_details')
                             ->label('Description')
-                            ->columnSpan(1),
+                            ->columnSpan(2),
```
Orrison, line 95: Let's do `columnSpanFull` instead here so we don't have to remember to update it if we ever change the columns here.

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
```diff
@@ -92,10 +92,10 @@ public function infolist(Infolist $infolist): Infolist
...
                         TextEntry::make('res_details')
                             ->label('Internal Details')
-                            ->columnSpan(1),
+                            ->columnSpan(2),
```
Orrison, line 98: Let's do `columnSpanFull` instead here so we don't have to remember to update it if we ever change the columns here.

Orrison (changes requested):

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
kandarp-canyon: done.

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ViewServiceRequest.php
kandarp-canyon: done.

Orrison (approved):

PR #424, opened 2025-02-18
[AIDAPP-466]: Indicate contact's name when that task is related to a contact in the resulting notification.

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-466

### Technical Description

> Indicate contact's name when that task is related to a contact in the resulting notification.

> No

> No

### DIFF
--- a/app-modules/task/src/Notifications/TaskAssignedToUserNotification.php
+++ b/app-modules/task/src/Notifications/TaskAssignedToUserNotification.php
@@ -36,6 +36,8 @@
 
 namespace AidingApp\Task\Notifications;
 
+use AidingApp\Contact\Filament\Resources\ContactResource;
+use AidingApp\Contact\Filament\Resources\OrganizationResource;
 use AidingApp\Notification\Notifications\BaseNotification;
 use AidingApp\Notification\Notifications\Concerns\DatabaseChannelTrait;
 use AidingApp\Notification\Notifications\Concerns\EmailChannelTrait;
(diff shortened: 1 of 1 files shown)


amit-canyon:
@Orrison  This is branch from [AIDAPP-470].
Thanks!

app-modules/task/src/Notifications/TaskAssignedToUserNotification.php
```diff
@@ -77,11 +78,17 @@ public function toDatabase(object $notifiable): array
 
         $title = str($this->task->title)->limit();
 
-        $link = new HtmlString("<a href='{$url}' target='_blank' class='underline'>{$title}</a>");
+        $message = match (true) {
+            is_null($this->task->concern->organization) => "You have been assigned a new Task: <a href='{$url}' target='_blank' class='underline'>{$title}</a> related to Contact <a href='" . ContactResource::getUrl('view', ['record' => $this->task->concern]) . "' target='_blank' class='underline'>{$this->task->concern?->full_name}</a>",
+
+            ! is_null($this->task->concern->organization) => "You have been assigned a new Task: <a href='{$url}' target='_blank' class='underline'>{$title}</a> related to Contact <a href='" . ContactResource::getUrl('view', ['record' => $this->task->concern]) . "' target='_blank' class='underline'>{$this->task->concern?->full_name}</a><a href='" . OrganizationResource::getUrl('view', ['record' => $this->task->concern->organization]) . "' target='_blank' class='underline'>({$this->task->concern->organization?->name})</a>",
+
+            default => "You have been assigned a new Task: <a href='{$url}' target='_blank' class='underline'>{$title}</a>",
```
Orrison, line 86: A problem here is that the `task->concern` can be null which I think would cause trouble here both from a logic perspective and also perhaps throw an error is some of your checks here.

You should check
1. If there is not a concern
2. If there is a concern but no organization then display the new message with the name but no org in parenthesis
3. if there is a concern with an org then display the new message with the name and org in parenthesis

Orrison (changes requested):

PR #428, opened 2025-02-19
[AIDAPP-455]: Improve the user relationship between teams and divisions

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-455

### Technical Description

>Add 'is_default' feature to Division model and related components

> No.

> Yes.
> DivisionIsDefault

### DIFF
--- a/app-modules/division/database/factories/DivisionFactory.php
+++ b/app-modules/division/database/factories/DivisionFactory.php
@@ -53,8 +53,6 @@ class DivisionFactory extends Factory
             'name' => fake()->unique()->company(),
             'code' => fake()->unique()->word(),
             'description' => fake()->optional()->sentences(asText: true),
-            'header' => fake()->words(asText: true),
-            'footer' => fake()->words(asText: true),
             'is_default' => false,
         ];
     }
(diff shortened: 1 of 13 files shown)


Orrison (changes requested):
There are some merge conflicts that need to be resolved here.

app-modules/division/database/factories/DivisionFactory.php
```diff
@@ -53,6 +53,16 @@ public function definition(): array
             'name' => fake()->unique()->company(),
             'code' => fake()->unique()->word(),
             'description' => fake()->optional()->sentences(asText: true),
+            'is_default' => false,
         ];
     }
+
+    public function default(): Factory
```
Orrison, line 60: Nice!

app-modules/team/src/Filament/Resources/TeamResource/Pages/CreateTeam.php
Orrison, line 1: Let's add `protected ?bool $hasDatabaseTransactions = true;` to this page to make sure this update is done in a transaction. Just in case something goes wrong we don't want the the other default to be wiped if this fails after `saving`.

app-modules/team/src/Filament/Resources/TeamResource/Pages/EditTeam.php
Orrison, line 1: Let's add `protected ?bool $hasDatabaseTransactions = true;` to this page to make sure this update is done in a transaction. Just in case something goes wrong we don't want the the other default to be wiped if this fails after `saving`.

Orrison (approved):

PR #429, opened 2025-02-19
[AIDAPP-456]: Remove obsolete header and footer configuration options for Division configuration

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-456

### Technical Description

Remove header and footer fields from Division pages, related files, and database.

No

No

### DIFF
--- a/app-modules/division/database/migrations/2025_02_19_171906_remove_header_footer_coloumns_from_divisions_table.php
+++ b/app-modules/division/database/migrations/2025_02_19_171906_remove_header_footer_coloumns_from_divisions_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2025, Canyon GBS LLC. All rights reserved.
+
+    Aiding App™ is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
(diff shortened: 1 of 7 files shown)


app-modules/division/database/migrations/2025_02_19_171906_remove_header_footer_coloumns_from_divisions_table.php
```diff
@@ -0,0 +1,58 @@
...
+*/
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        if (Schema::hasTable('divisions')) {
```
Orrison, line 44: Hmm, we don't typically do checks like this as we assume the table should be here. We should probably remove this.

Orrison (changes requested):

app-modules/division/database/migrations/2025_02_19_171906_remove_header_footer_coloumns_from_divisions_table.php
Orrison: I made this change so we could include this in today's release.

Orrison (approved):

app-modules/division/database/migrations/2025_02_19_171906_remove_header_footer_coloumns_from_divisions_table.php
rebecca-canyon: Thank you! I'll keep this in mind for the future.

PR #462, opened 2025-03-11
[AIDAPP-494]: Add a filter for service requests for the party a ticket is assigned to, including unassigned as an option

ketan-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-494

### Technical Description

Add a filter for service requests for the party a ticket is assigned to, including unassigned as an option

No

No


app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ListServiceRequests.php
```diff
@@ -164,11 +165,33 @@ public function table(Table $table): Table
...
+                                fn (Builder $baseQuery) => $baseQuery->whereDoesntHave('assignedTo', function (Builder $query) {
+                                    $query->where('status', ServiceRequestAssignmentStatus::Active);
+                                })
+                            )
+                            ->when(
+                                ! empty($data['value']) && $data['value'] !== 'unassigned',
+                                fn (Builder $baseQuery) => $baseQuery->whereHas(
+                                    'assignedTo.user',
+                                    fn (Builder $userQuery) => $userQuery->where('id', $data['value'])
+                                )
```
Orrison, line 193: I think we similarly want to check that the assignment is active. Similiar to what you did above with `->where('status', ServiceRequestAssignmentStatus::Active)`

Orrison (changes requested):

Orrison (changes requested):
@ketan-canyon, something is very wrong with this PR. It looks like a bad rebase or something. Please resolve.

ketan-canyon:
> @ketan-canyon, something is very wrong with this PR. It looks like a bad rebase or something. Please resolve.

Hi @Orrison 
I think its because by mistake i run --rebase=true when take pull from main. nothing else i think.

Orrison (changes requested):
@ketan-canyon this was not fixed. There were still 200+ file changes that were related to things that already merged into `main` due to a bad rebase. Please pay attention to that and make sure to review your own work in GitHub and ensure that the changes are looking correct before requesting review from others.

I have re-done the rebase for you. Please review and re-request my review if what I accepted looked correct to what you intended to change.

app-modules/service-management/src/Filament/Resources/ServiceRequestResource/Pages/ListServiceRequests.php
```diff
@@ -166,11 +167,38 @@ public function table(Table $table): Table
...
+                                ! empty($data['value']) && $data['value'] !== 'unassigned',
+                                fn (Builder $baseQuery) => $baseQuery->whereHas(
+                                    'assignedTo.user',
+                                    fn (Builder $userQuery) => $userQuery
+                                        ->where('id', $data['value'])
+                                )->whereHas(
+                                    'assignedTo',
+                                    fn (Builder $assignedToQuery) => $assignedToQuery
+                                        ->where('status', ServiceRequestAssignmentStatus::Active)
+                                )
```
Orrison, line 200: I could be wrong. But I don't think this check should be separate from the first `whereHas` here.

Does this result in a query that filters by users with that ID where the `ServiceRequestAssignmentStatus` is active?

We should probably adjust the test/s to ensure all of this is covered. So have Users where the assignment is active and ones where it is not, and then users where there is no assignment records at all.

Orrison (changes requested):

Orrison (approved):

PR #487, opened 2025-03-28
[AIDAPP-526]: Restrict the visibility of a license record to the creator and assigned party only

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-526

### Technical Description

> Restrict the visibility of a license record to the creator and assigned party only

> No

> No

### DIFF
--- a/app-modules/license-management/src/Models/Scopes/AuthorizeLicensesScope.php
+++ b/app-modules/license-management/src/Models/Scopes/AuthorizeLicensesScope.php
@@ -36,6 +36,7 @@
 
 namespace AidingApp\LicenseManagement\Models\Scopes;
 
+use App\Models\User;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Scope;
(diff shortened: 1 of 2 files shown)


app-modules/license-management/src/Models/Scopes/AuthorizeLicensesScope.php
```diff
@@ -0,0 +1,63 @@
...
+
+        if (! $user instanceof User) {
+            return;
+        }
+
+        if ($user?->isSuperAdmin()) {
+            return;
+        }
+
+        $builder->where('created_by_id', $user?->getKey());
```
Orrison, line 61: The nullsafe operators should not be needed here, as on line `53` you have now ensured that the `$user` variable IS an instance of `User`. So it could not be `null`.

```suggestion
        if ($user->isSuperAdmin()) {
            return;
        }

        $builder->where('created_by_id', $user->getKey());
```

Orrison (changes requested):

Orrison (approved):

PR #493, opened 2025-03-31
[AIDAPP-530]: Engagements sent to contacts are not using the correct communication settings configuration attached to the division.

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-530

### Technical Description

> Fixed: Engagements sent to contacts are not using the correct communication settings configuration attached to the division.

> No.

> No.

### DIFF
--- a/app/Models/User.php
+++ b/app/Models/User.php
@@ -280,7 +280,6 @@ class User extends Authenticatable implements HasLocalePreference, FilamentUser,
         return $this
             ->belongsToMany(Team::class, 'team_user', 'user_id', 'team_id')
             ->using(TeamUser::class)
-            ->limit(1)
             ->withTimestamps();
     }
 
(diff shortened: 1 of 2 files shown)


app/Models/User.php
```diff
@@ -280,7 +280,6 @@ public function teams(): BelongsToMany
         return $this
             ->belongsToMany(Team::class, 'team_user', 'user_id', 'team_id')
             ->using(TeamUser::class)
-            ->limit(1)
```
Orrison, line 283: Why was this removed? This was intentionally placed here for now as we only want Users to be related to a single team.

Orrison (changes requested):

app/Models/User.php
Orrison: Explained in office hours. Good to go.

Orrison (approved):

PR #500, opened 2025-04-02
[AIDAPP-537]: Change PHPStan to Level 6 and create a baseline

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-537

### Technical Description

> Changed PHPStan to Level 6 and create a baseline.

> No.

> No.


Orrison (changes requested):

phpstan.neon.dist
```diff
@@ -18,9 +19,9 @@ parameters:
...
-    ignoreErrors:
-        -
-            identifier: larastan.noEnvCallsOutsideOfConfig
-            path: app-modules/*/config/*
+    level: 6
+    
+    #ignoreErrors:
+        #-
+            #identifier: larastan.noEnvCallsOutsideOfConfig
+            #path: app-modules/*/config/*
```
rebecca-canyon, line 27: I don't believe that we want to comment out these lines.

rebecca-canyon (changes requested):

phpstan.neon.dist
Orrison: @rebecca-canyon I had asked Ankit in a teams channel to comment these out instead of remove them. But I am going to opt to remove it now. We will probably need to add this back later. But we can deal with it then.

PR #516, opened 2025-04-11
[AIDAPP-553]: Resolve PHPStan linting errors for a time block

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-553

### Technical Description

> Resolve PHPStan linting errors for a time block

> No

> No


app-modules/alert/src/Listeners/NotifySubscribersOfAlertCreated.php
```diff
@@ -49,7 +49,10 @@ public function handle(AlertCreated $event): void
         /** @var Contact $concern */
         $concern = $event->alert->concern;
 
-        $concern->subscriptions?->each(function (Subscription $subscription) use ($event) {
+        /** @var Subscription|null $subscriptions */
```
danharrin, line 52: Is this not a collection, multiple subscriptions?

danharrin (changes requested):

app-modules/alert/src/Listeners/NotifySubscribersOfAlertCreated.php
amit-canyon: I created a new variable to address a phpstan linting error caused by removing the nullsafe operator. @danharrin , what's the best approach here?

danharrin: Should it not be `@var Collection<Subscription> $subscriptions`?

danharrin (changes requested):

app-modules/alert/src/Listeners/NotifySubscribersOfAlertCreated.php
```diff
@@ -49,7 +50,10 @@ public function handle(AlertCreated $event): void
         /** @var Contact $concern */
         $concern = $event->alert->concern;
 
-        $concern->subscriptions?->each(function (Subscription $subscription) use ($event) {
+        /** @var ?Collection<int, Subscription> $subscriptions */
+        $subscriptions = $concern->subscriptions;
```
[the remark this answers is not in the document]
danharrin: Can you please run the code and dump the values of the variables to see what if the types that are written down are correct? And see if `concern` is a nullable relationship on an alert?

danharrin (approved):

PR #518, opened 2025-04-15
[AIDAPP-512]: Enhance Service Monitoring by adding the ability to schedule checks and notify users when performance issues are detected via an monitored endpoint

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-512

### Technical Description

Introduce jobs to sent out notifications for domains that are down.

No

No

### DIFF
--- a/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
+++ b/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
@@ -72,11 +72,14 @@ class ServiceMonitoringCheckJob implements ShouldQueue
             ]);
 
             if ($response->status() !== 200) {
-                /** @var Collection<int,User> $recipients */
+                /** @var Collection<int, User> $recipients */
                 $recipients = $this->serviceMonitoringTarget->users()->get();
 
                 $this->serviceMonitoringTarget->teams()->each(function ($team) use ($recipients) {
-                    $recipients->concat($team->users()->get())->unique();
+                  /** @var Collection<int, User> $users */
+                  $users = $team->users()->get();  
+                  
+                  $recipients->concat($users)->unique();
                 });
 
                 $recipients->each(fn ($user) => $user->notify(new ServiceMonitoringNotification()));
--- a/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
+++ b/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
@@ -80,30 +80,26 @@ class ServiceMonitoringCheckJob implements ShouldQueue, ShouldBeUnique
 
     public function handle(): void
     {
-        try {
-            $response = Http::get($this->serviceMonitoringTarget->domain);
+      $response = Http::get($this->serviceMonitoringTarget->domain);
 
-            $this->serviceMonitoringTarget->history()->create([
-                'response' => $response->status(),
-                'response_time' => $response->transferStats->getTransferTime() ?? 0,
-                'succeeded' => $response->status() === 200,
-            ]);
+      $this->serviceMonitoringTarget->history()->create([
+          'response' => $response->status(),
+          'response_time' => $response->transferStats->getTransferTime() ?? 0,
+          'succeeded' => $response->status() === 200,
+      ]);
 
-            if ($response->status() !== 200) {
-                /** @var Collection<int, User> $recipients */
-                $recipients = $this->serviceMonitoringTarget->users()->get();
+      if ($response->status() !== 200) {
+          /** @var Collection<int, User> $recipients */
+          $recipients = $this->serviceMonitoringTarget->users()->get();
 
-                $this->serviceMonitoringTarget->teams()->each(function ($team) use ($recipients) {
-                    /** @var Collection<int, User> $users */
-                    $users = $team->users()->get();
+          $this->serviceMonitoringTarget->teams()->each(function ($team) use ($recipients) {
+              /** @var Collection<int, User> $users */
+              $users = $team->users()->get();
 
-                    $recipients->concat($users)->unique();
-                });
+              $recipients->concat($users)->unique();
+          });
 
-                Notification::send($recipients, new ServiceMonitoringNotification($this->serviceMonitoringTarget->history));
-            }
-        } catch (Exception $e) {
-            report($e);
-        }
+          Notification::send($recipients, new ServiceMonitoringNotification($this->serviceMonitoringTarget->history));
+      }
     }
 }
--- a/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
+++ b/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
@@ -81,7 +81,7 @@ class ServiceMonitoringCheckJob implements ShouldQueue, ShouldBeUnique
     {
         $response = Http::get($this->serviceMonitoringTarget->domain);
 
-        $this->serviceMonitoringTarget->history()->create([
+        $history = $this->serviceMonitoringTarget->histories()->create([
             'response' => $response->status(),
             'response_time' => $response->transferStats->getTransferTime() ?? 0,
             'succeeded' => $response->status() === 200,
@@ -98,7 +98,7 @@ class ServiceMonitoringCheckJob implements ShouldQueue, ShouldBeUnique
                 $recipients->concat($users)->unique();
             });
 
-            Notification::send($recipients, new ServiceMonitoringNotification($this->serviceMonitoringTarget->history));
+            Notification::send($recipients, new ServiceMonitoringNotification($history));
         }
     }
 }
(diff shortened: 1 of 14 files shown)


app-modules/service-management/tests/Filament/ServiceMonitoring/Pages/EditServiceMonitoringTest.php
```diff
@@ -81,10 +82,19 @@
...
         ->assertHasNoFormErrors();
 
+    /** @var string $description */
+    $description = $serviceMonitoringTarget->fresh()->description;
+
+    /** @var string $domain */
+    $domain = $serviceMonitoringTarget->fresh()->domain;
+
+    /** @var ServiceMonitoringFrequency $frequency */
+    $frequency = $serviceMonitoringTarget->fresh()->frequency;
```
Orrison, line 92: Ah, we should not have done this before. Rather than calling `fresh()` every time here, before all these variable definitions lets just do one `$serviceMonitoringTarget->refresh()`.

Which will bring these 3 queries, 2 of which are not needed, down to 1.

app-modules/service-management/tests/Jobs/ServiceMonitoringCheckJobTest.php
```diff
@@ -0,0 +1,69 @@
...
+    Notification::fake();
+
+    $user = User::factory()->create();
+
+    $serviceMonitorTarget = ServiceMonitoringTarget::factory()
+        ->hasAttached($user)
+        ->create(['frequency' => $frequency]);
+
+    (new ServiceMonitoringCheckJob($serviceMonitorTarget))->handle();
+
```
Orrison, line 58: Let's also asser that the proper history record was created.

app-modules/service-management/tests/Jobs/ServiceMonitoringCheckJobTest.php
```diff
@@ -0,0 +1,69 @@
...
+*/
+
+use AidingApp\ServiceManagement\Enums\ServiceMonitoringFrequency;
+use AidingApp\ServiceManagement\Jobs\ServiceMonitoringCheckJob;
+use AidingApp\ServiceManagement\Models\ServiceMonitoringTarget;
+use AidingApp\ServiceManagement\Notifications\ServiceMonitoringNotification;
+use App\Models\User;
+use Illuminate\Support\Facades\Http;
+use Illuminate\Support\Facades\Notification;
+
```
Orrison, line 44: Let's add an additional test here for if the response WAS a 200. And ensure that the proper history record was created and that there was no notification sent out at all.

app/Console/Kernel.php
```diff
@@ -76,6 +79,20 @@ protected function schedule(Schedule $schedule): void
                         ->onOneServer()
                         ->withoutOverlapping(15);
 
+                    $schedule->call(function () use ($tenant) {
+                        $tenant->execute(function () {
+                            Bus::batch([new ServiceMonitoringJob(ServiceMonitoringFrequency::OneHour)])->dispatch();
```
Orrison, line 84: I don't think we need the batch functionality at the moment for these. So let's just do a normal `dispatch`

app/Console/Kernel.php
```diff
@@ -76,6 +79,20 @@ protected function schedule(Schedule $schedule): void
...
+                        });
+                    })
+                        ->hourly();
+
+                    $schedule->call(function () use ($tenant) {
+                        $tenant->execute(function () {
+                            Bus::batch([new ServiceMonitoringJob(ServiceMonitoringFrequency::TwentyFourHours)])->dispatch();
```
Orrison, line 91: I don't think we need the batch functionality at the moment for these. So let's just do a normal `dispatch`

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -0,0 +1,93 @@
...
+                    $users = $team->users()->get();
+
+                    $recipients->concat($users)->unique();
+                });
+
+                Notification::send($recipients, new ServiceMonitoringNotification($historicalServiceMonitoring, $this->serviceMonitoringTarget));
+            }
+        } catch (Exception $e) {
+            report($e);
+        }
```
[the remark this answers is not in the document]
rebecca-canyon: It was mentioned in the tech planning notes to put the HTTP call in a try/catch block, though I don't recall any discussion on doing anything in particular to handle an exception. I can certainly remove the try/catch if you don't think it's necessary, though.

app-modules/service-management/src/Models/HistoricalServiceMonitoring.php
```diff
@@ -0,0 +1,64 @@
...
+class HistoricalServiceMonitoring extends BaseModel implements Auditable
+{
+    use AuditableTrait;
+    use SoftDeletes;
+
+    protected $fillable = [
+        'response',
+        'response_time',
+        'succeeded',
+        'service_monitoring_target_id',
```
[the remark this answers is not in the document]
rebecca-canyon: Even with the rest of the changes you've suggested implemented, removing the fillable property causes the notification to be unable to send with an `Add [X] to fillable property to allow mass assignment on [AidingApp\ServiceManagement\Models\HistoricalServiceMonitoring].` error. Maybe I'm missing something, though?

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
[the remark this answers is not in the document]
Orrison: I do recall us discussing that but do not remember how I was suggesting it setup. Reading this now though I don't think it is needed, specifically so surrounding the whole job.

Right now this would just cause the job not to fail, which I think we would want to happen.

So, yeah, let's remove it for now. We can iterate later if we find a need for it to not cause the job to fail or to only report certain errors, etc.

app-modules/service-management/src/Jobs/ServiceMonitoringJob.php
```diff
@@ -0,0 +1,85 @@
...
+    use Dispatchable;
+    use InteractsWithQueue;
+    use Queueable;
+    use SerializesModels;
+
+    public function __construct(public ServiceMonitoringFrequency $interval) {}
+
+    public function uniqueId(): string
+    {
+        return $this->interval->getLabel();
```
Orrison, line 60: I think you may want `value` instead of `getLabel()` here. As you would want the no spaces, storable, representation of the ENUM.

app-modules/service-management/src/Models/HistoricalServiceMonitoring.php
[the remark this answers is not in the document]
Orrison: Hmm, can you post the actual stack trace or show this error in office hours? That seems strange, if you are using the relationship it shouldn't require the fillable properties.

app-modules/service-management/src/Notifications/ServiceMonitoringNotification.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return ['mail'];
+    }
+
+    public function toMail(User $notifiable): MailMessage
+    {
+        $this->historicalServiceMonitoring->loadMissing('serviceMonitoringTarget');
+
+        return MailMessage::make()
+            ->settings(null)
```
Orrison, line 65: I could be wrong, but I think you should be able to just not include `settings` here.

Orrison (changes requested):

app-modules/service-management/src/Models/ServiceMonitoringTarget.php
```diff
@@ -66,6 +70,14 @@ class ServiceMonitoringTarget extends BaseModel implements Auditable
         'frequency' => ServiceMonitoringFrequency::class,
     ];
 
+    /**
+     * @return HasOne<HistoricalServiceMonitoring, $this>
+     */
+    public function history(): HasOne
+    {
+        return $this->hasOne(HistoricalServiceMonitoring::class);
```
Orrison, line 78: This should probably be a `hasMany` as this will be running every hour/day and storing a history record for each invocation.

And then at that point it may make more sense for this to be plural, `histories`.

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -0,0 +1,104 @@
...
+            $recipients = $this->serviceMonitoringTarget->users()->get();
+
+            $this->serviceMonitoringTarget->teams()->each(function ($team) use ($recipients) {
+                /** @var Collection<int, User> $users */
+                $users = $team->users()->get();
+
+                $recipients->concat($users)->unique();
+            });
+
+            Notification::send($recipients, new ServiceMonitoringNotification($this->serviceMonitoringTarget->history));
```
Orrison, line 101: Like I said in the other comment, we want to change the history relationship to be a one to many.

So `$this->serviceMonitoringTarget->history` would not work here.

You can save the history record to a variable on like `84` here and dispatch using that record specifically.

Orrison (changes requested):

Orrison (approved):

PR #527, opened 2025-04-28
[AIDAPP-559]: Populate the email templates for service requests with the current messaging that is used

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-559

### Technical Description

> feat: implement service request type email templates and button block for email management

> No

> Yes.


payal-canyon:
Hi @Orrison and @danharrin 

Whenever you're reviewing this PR, there's one test case I'd like your input on.

In the NotifyServiceRequestUsersTest.php file, I've added a skip message to the test:
does not notify a user twice if they belong to a team managing and auditing a service request type.

The skip reason currently says:
skip('we can attach user to one team only. also, NotifyServiceRequestUsers logic has been updated. so do we need this test case?')

Once you've completed the overall PR review, could you please check this test as well and let me know whether it should be kept or removed?
If you think it's still relevant, a brief suggestion or hint on how it could be written based on the updated logic would be really helpful.

Thanks!

app-modules/service-management/src/Providers/ServiceManagementServiceProvider.php
```diff
@@ -75,6 +77,13 @@ public function register(): void
     {
         Panel::configureUsing(fn (Panel $panel) => ($panel->getId() !== 'admin') || $panel->plugin(new ServiceManagementPlugin()));
 
+        TiptapEditor::configureUsing(function (TiptapEditor $component) {
+            $component
+                ->blocks([
+                    ServiceRequestTypeEmailTemplateButtonBlock::class,
+                ]);
+        });
```
danharrin, line 85: Instead of registering these blocks globally, they should probably be registered only on the TipTap instances that they should be used

danharrin (changes requested):

danharrin (dismissed):
Updates to requests marked as "internal" currently send emails to customers. Should this be the case?

Orrison (approved):

Orrison:
> Where did we land with the "survey response" templates? These forms exist in the UI but are not populated with data currently

I am creating a separate ticket to follow up here to add these

PR #547, opened 2025-05-13
[AIDAPP-586]: Introduce the ability to set a custom label on the button for viewing service requests in the email notification blocks for customers, managers, and auditors

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-586

### Technical Description

> Introduce the ability to set a custom label on the button for viewing service requests in the email notification blocks for customers, managers, and auditors.

> No

> No

### DIFF
--- a/app-modules/service-management/src/Filament/Blocks/ServiceRequestTypeEmailTemplateButtonBlock.php
+++ b/app-modules/service-management/src/Filament/Blocks/ServiceRequestTypeEmailTemplateButtonBlock.php
@@ -59,7 +59,7 @@ class ServiceRequestTypeEmailTemplateButtonBlock extends TiptapBlock
     public function getFormSchema(): array
     {
         return [
-            TextInput::make('view_service_request')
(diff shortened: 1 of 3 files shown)


app-modules/service-management/src/Filament/Blocks/ServiceRequestTypeEmailTemplateButtonBlock.php
```diff
@@ -58,6 +59,12 @@ class ServiceRequestTypeEmailTemplateButtonBlock extends TiptapBlock
     public function getFormSchema(): array
     {
         return [
+            TextInput::make('view_service_request')
```
Orrison, line 62: I would maybe call this `label`. It will then need to be changed in the blade files.

Orrison (changes requested):

Orrison (approved):

PR #550, opened 2025-05-14
[AIDAPP-582]: ServiceMonitoringCheckJob fails and throws an error if it's target requires more than 5 redirects

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-582

### Technical Description

> Fixed: ServiceMonitoringCheckJob fails and throws an error if it's target requires more than 5 redirects.

> No.

> No.

### DIFF
--- a/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
+++ b/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
@@ -80,7 +80,7 @@ class ServiceMonitoringCheckJob implements ShouldQueue, ShouldBeUnique
     public function handle(): void
     {
         $response = Http::maxRedirects(15)
-                          ->get($this->serviceMonitoringTarget->domain);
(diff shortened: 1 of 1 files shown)


app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -80,11 +80,11 @@ public function uniqueFor(): int
...
     {
         $response = Http::withOptions([
-                            'allow_redirects' => [
-                              'max' => 15,
-                            ],
-                          ])
-                          ->get($this->serviceMonitoringTarget->domain);
+            'allow_redirects' => [
+                'max' => 15,
+            ],
```
Orrison, line 85: I believe there exists a fluent method for this `maxRedirects`

Orrison (changes requested):

Orrison (approved):

PR #551, opened 2025-05-14
[AIDAPP-585]: Purge subscriptions from the product

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-585

### Technical Description

> Purge subscriptions from the product

> No

> No


database/seeders/DatabaseSeeder.php
```diff
@@ -87,7 +87,6 @@ public function run(): void
             TaskSeeder::class,
             AlertSeeder::class,
             TeamSeeder::class,
-            SuperAdminSeeder::class,
```
Orrison, line 90: Why delete this? I think it is still needed locally to give us the Super Admin User.

database/seeders/SuperAdminSeeder.php
Orrison, line 1: Why delete this? I think it is still needed locally to give us the Super Admin User.

Orrison (changes requested):

database/seeders/SuperAdminSeeder.php
payal-canyon: Hi @Orrison, 
I removed this because the SuperAdmin wasn't being created in this file. Instead, it's creating contact subscriptions, so I went ahead and removed the SuperAdminSeeder.php file along with its references.

database/seeders/DatabaseSeeder.php
payal-canyon: same as below comment.

Orrison (approved):

PR #553, opened 2025-05-19
[AIDAPP-591]: Scan and improve underlying codebase

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-591

### Technical Description

> Scan and improve underlying codebase.

> No.

> No.

### DIFF
--- a/app/Models/Scopes/EducatableSort.php
+++ b/app/Models/Scopes/EducatableSort.php
@@ -37,7 +37,6 @@
 namespace App\Models\Scopes;
 
 use AidingApp\Contact\Models\Contact;
-use AidingApp\ServiceManagement\Models\ServiceRequest;
 use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Facades\DB;
 
--- a/app/Models/Scopes/EducatableSort.php
+++ b/app/Models/Scopes/EducatableSort.php
@@ -37,6 +37,7 @@
 namespace App\Models\Scopes;
 
 use AidingApp\Contact\Models\Contact;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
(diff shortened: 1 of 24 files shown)


tests/Helpers.php
```diff
@@ -40,6 +40,9 @@
 use AidingApp\Authorization\Models\Role;
 use App\Actions\Paths\ModulePath;
 use App\Models\User;
+use Illuminate\Foundation\Testing\TestCase;
```
Orrison, line 43: I just want to make sure we are using the correct `TestCase` reference here since there is one in this directory. Did this have to be fixed because `actingAs` returns the `Illuminate\Foundation\Testing\TestCase` one?

app/Models/Scopes/EducatableSort.php
```diff
@@ -46,6 +47,9 @@ public function __construct(
         protected string $direction
     ) {}
 
+    /**
+     * @param Builder<ServiceRequest> $query
```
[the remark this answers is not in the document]
ankit-canyon: Contact*

tests/Helpers.php
ankit-canyon: > I just want to make sure we are using the correct `TestCase` reference here since there is one in this directory. Did this have to be fixed because `actingAs` returns the `Illuminate\Foundation\Testing\TestCase` one?

Yep so previous error is 

Call to an undefined method Pest\PendingCalls\TestCall|Pest\Support\HigherOrderTapProxy::actingAs()

to solve I removed the `test()` and returns only `actingsAs()` which should return  `Illuminate\Foundation\Testing\TestCase`

Orrison (approved):

PR #558, opened 2025-05-26
[AIDAPP-592]: ServiceMonitoringCheckJob fails if the provided URL is unresolvable

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-592

### Technical Description

> ServiceMonitoringCheckJob fails if the provided URL is unresolvable

> No

> No

### DIFF
--- a/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
+++ b/app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
@@ -45,6 +45,7 @@ use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Database\Eloquent\Collection;
 use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Foundation\Queue\Queueable;
+use Illuminate\Http\Client\ConnectionException;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Support\Facades\Http;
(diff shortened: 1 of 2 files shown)


app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -79,24 +80,41 @@ public function uniqueFor(): int
...
-            $this->serviceMonitoringTarget->teams()->each(function ($team) use ($recipients) {
-                /** @var Collection<int, User> $users */
+            $this->serviceMonitoringTarget->teams()->each(function ($team) use (&$recipients) {
                 $users = $team->users()->get();
-
-                $recipients->concat($users)->unique();
+                $recipients = $recipients->merge($users)->unique('id');
             });
 
             Notification::send($recipients, new ServiceMonitoringNotification($history));
```
Orrison, line 120: Because this code is pretty much the exact same as within the status check. Do you think we can turn this into a protected method in this class that can be re-used in each place?

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -79,24 +80,41 @@ public function uniqueFor(): int
...
+                    /** @var Collection<int, User> $users */
+                    $users = $team->users()->get();
+
+                    $recipients->concat($users)->unique();
+                });
+
+                Notification::send($recipients, new ServiceMonitoringNotification($history));
+            }
+        } catch (ConnectionException $e) {
+            $history = $this->serviceMonitoringTarget->histories()->create([
```
Orrison, line 107: Just before this, let's also `report()` the Exception if the message doesn't contain `Could not resolve host`. That way we can be notified if this fails for some other strange reason we have not yet accounted for.

app-modules/service-management/tests/Jobs/ServiceMonitoringCheckJobTest.php
```diff
@@ -95,3 +96,31 @@
             fn () => ServiceMonitoringFrequency::TwentyFourHours,
         ]
     );
+
+it('Handle unresolvable host errors gracefully', function ($frequency) {
+    Http::fake(function () {
+        throw new ConnectionException('Connection failed');
```
Orrison, line 102: That is not what the execption message looks like. Look at the error in the ticket.

app-modules/service-management/tests/Jobs/ServiceMonitoringCheckJobTest.php
```diff
@@ -95,3 +96,31 @@
             fn () => ServiceMonitoringFrequency::TwentyFourHours,
         ]
     );
+
+it('Handle unresolvable host errors gracefully', function ($frequency) {
```
Orrison, line 100:
```suggestion
it('handles unresolvable host errors gracefully', function ($frequency) {
```

Orrison (changes requested):

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -79,26 +78,32 @@ public function uniqueFor(): int
...
-            'response_time' => $response->transferStats->getTransferTime() ?? 0,
-            'succeeded' => $response->status() === 200,
+            'response' => $status,
+            'response_time' => $responseTime,
+            'succeeded' => $success,
         ]);
 
-        if ($response->status() !== 200) {
-            /** @var Collection<int, User> $recipients */
+        if ($status !== 200) {
```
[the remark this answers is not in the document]
Orrison: Yes

PR #560, opened 2025-05-26
[AIDAPP-605]: Introduce full text searching from customer portal

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-605

### Technical Description

> Introduce full text searching from customer portal.

> No.

> Yes
> ArticleFullTextSearch
> Data migration: app-modules/knowledge-base/database/migrations/2025_05_26_103749_data_migrate_article_details_to_article_details_fulltext_in_knowledge_base_articles.php


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalSearchController.php
```diff
@@ -67,7 +68,19 @@ public function get(Request $request): KnowledgeManagementSearchData
...
+                    fn (Builder $query) => $query
+                        ->whereRaw(
+                            "search_vector @@ websearch_to_tsquery('english', ?)",
+                            [$search]
+                        )
+                        ->orderByRaw(
+                            "ts_rank_cd(search_vector, websearch_to_tsquery('english', ?)) DESC",
+                            [$search]
+                        ),
+                    fn (Builder $query) => $query->tap(new SearchBy('title', $search))
```
Orrison, line 82: Setting the old search query to the "default" here is not going to work. As this will now result in the old query being triggered if `$search` is empty.

You need to leave it as two different whens but check the `ArticleFullTextSearch` Feature flag in each.

PR #566, opened 2025-06-06
[AIDAPP-598]: Implement on page content for status page

ankit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-598

### Technical Description

> Implement on page content for status page.

> No.

> No.


app-modules/portal/tests/Unit/ServiceMonitorStatusApiTest.php
```diff
@@ -0,0 +1,74 @@
...
+            'response_time' => 0.123,
+            'succeeded' => true,
+            'response' => 200,
+        ])->forceFill(['id' => '9f18ac66-7369-4282-b834-9f672f49b2bc'])->save();
+    }
+
+    $url = URL::route(name: 'api.portal.status', absolute: false);
+    $response = get($url);
+
+    expect($response)->toMatchSnapshot();
```
Orrison, line 73: Because of the need to hardcode things like the `id` here I think it would be better to test his normally. Not using a snapshot.

tests/.pest/snapshots/var/www/html/app-modules/portal/tests/Unit/ServiceMonitorStatusApiTest/returns_all_service_monitoring_targets_with_latest_history_when_portal_is_enabled.snap
Orrison, line 1: This should be deleted when we change the test to not use snapshots.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/ServiceMonitorStatusController.php
```diff
@@ -0,0 +1,102 @@
...
+            ->orderBy('created_at', 'ASC')
+            ->paginate($perPage);
+
+        $paginated->through(function (ServiceMonitoringTarget $target) {
+            $latestHistory = $target->latestHistory;
+
+            if ($latestHistory) {
+                $statusMessage = match ($latestHistory->succeeded) {
+                    true => 'No known issues at this time',
+                    false => '1 unknown issue reported',
```
Orrison, line 67: Let's change this to say something like "Unable to reach service, status code: [STATUS_CODE_FROM_HISTORY]"

portals/knowledge-management/src/Pages/ServiceMonitorStatus.vue
```diff
@@ -0,0 +1,211 @@
...
+    const okMessage =
+        'All systems are functioning seamlessly, with no disruptions or downtime reported. Every component, from critical infrastructure to auxiliary services is running at full capacity, ensuring optimal performance and reliability.';
+
+    const issueTitle = 'Some systems are experiencing issues';
+    const issueMessage = 'One or more services are currently experiencing disruptions or downtime.';
+
+    const hasIssues = computed(() =>
+        result.value.some(
+            (serviceMonitor) =>
+                serviceMonitor.latest_history !== null && serviceMonitor.latest_history.succeeded !== true,
```
Orrison, line 74: I don't think we should consider a service monitor not having any history as an issue.

It should probably count as there not being any problems. Since we have 24 hour checks this would mean that when someone added a new 24 hour check it would say there are problems until the check is first run later.

Orrison (changes requested):

PR #604, opened 2025-07-02
[Bugfix]: ServiceRequestAssignmentObserver missing relation issue

Orrison:
### Ticket(s) or GitHub Issue

- N/A

### Technical Description

Work to fix an issue where the serviceRequest relation may not yet be loaded in within `ServiceRequestAssignmentObserver`.

https://canyongbs.sentry.io/issues/6713712757/events/fae876f5ba624b6a9d4b0df7cef5dd75/

No

No

### DIFF
--- a/app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
+++ b/app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
@@ -56,6 +56,8 @@ class ServiceRequestAssignmentObserver
 
     public function creating(ServiceRequestAssignment $serviceRequestAssignment): void
     {
+        $serviceRequestAssignment->loadMissing('serviceRequest');
+
         throw_if(! in_array($serviceRequestAssignment->user_id, $serviceRequestAssignment->serviceRequest->priority->type?->managers
             ->flatMap(fn ($managers) => $managers->users)
             ->pluck('id')


app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
```diff
@@ -56,6 +56,8 @@ class ServiceRequestAssignmentObserver
 
     public function creating(ServiceRequestAssignment $serviceRequestAssignment): void
     {
+        $serviceRequestAssignment->loadMissing('serviceRequest');
+
         throw_if(! in_array($serviceRequestAssignment->user_id, $serviceRequestAssignment->serviceRequest->priority->type?->managers
```
danharrin, line 61:
```suggestion
        throw_if(! in_array($serviceRequestAssignment->user_id, $serviceRequestAssignment->serviceRequest?->priority?->type?->managers
```

For safety, it's probably worth also doing this, in case the service request is truly missing and the load doesn't fix the issue? Also on line 64, adding a `?? []` after the `toArray()` so that the `in_array()` doesn't fix the issue.

However, I guess that `AttemptedToAssignNonManagerToServiceRequest` would actually get thrown in that case, which isn't caught anywhere. It would be really useful to inspect that `$serviceRequestAssignment` object, can it be converted into an array and attached to the exception message?

danharrin (changes requested):

PR #667, opened 2025-09-16
[AIDAPP-695]: Introduce cards at the top of the page and calculate uptime for a variety of intervals to support service monitoring

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-695

### Technical Description

Create and display service uptime widget to display various intervals of service uptime.

No

No


Orrison (changes requested):

Orrison (approved):
@rebecca-canyon we can add tests later, if you say this works after manually testing then this is fine.

PR #669, opened 2025-09-16
[AIDAPP-696]: Introduce a button at the top of the page for service monitoring to reset monitoring

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-696

### Technical Description
- Create a custom action for Reseting Monitoring (soft deleting the history)
- Add the button to header actions on View / Edit pages.

- No

- No

### DIFF
--- a/app-modules/service-management/src/Filament/Actions/ResetAction.php
+++ b/app-modules/service-management/src/Filament/Actions/ResetAction.php
@@ -48,7 +48,7 @@ class ResetAction extends Action
 
         $this->form([])
             ->label('Reset Monitoring')
-            ->action(fn (ServiceMonitoringTarget $record) => ResetMonitoringAction::execute($record))
+            ->action(fn (ServiceMonitoringTarget $record, ResetMonitoringAction $resetMonitoringAction) => $resetMonitoringAction::execute($record))
             ->requiresConfirmation()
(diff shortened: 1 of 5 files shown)


app-modules/service-management/src/Filament/Actions/ResetAction.php
```diff
@@ -0,0 +1,61 @@
...
+{
+    protected function setup(): void
+    {
+        parent::setup();
+
+        $this->form([])
+            ->label('Reset Monitoring')
+            ->action(function (array $data, ServiceMonitoringTarget $record): void {
+                $record->histories()->delete();
+            })
```
Orrison, line 52: Though this code is small and simple, the tech plan does specifically call for a separate action class for the execution that should be called here.

I agree with that in case we need to reset elsewhere and if the logic needed for it grows over time.

So I would create an additional action class, which can be loaded in with dep injection into this function, and call it instead of doing the `$record->histories()->delete();` here.
You can also probably turn this into a short closure.

app-modules/service-management/src/Filament/Actions/ResetAction.php
Orrison, line 1: Let's add a test for this ensuring what we expect happens.

Orrison (changes requested):

Orrison (approved):

PR #672, opened 2025-09-19
[AIDAPP-700]: Introduce a placeholder for sequences in service request settings

krlmrr:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-700

### Technical Description
- Create a Sequence model
- Create a Sequence Resource
- Comment out Create and Edit Pages
- Set the Index to the coming soon page

- No

- No

### DIFF
--- a/app-modules/service-management/src/Models/Sequence.php
+++ b/app-modules/service-management/src/Models/Sequence.php
@@ -1,0 +1,10 @@
+<?php
+
+namespace AidingApp\ServiceManagement\Models;
+
+use Illuminate\Database\Eloquent\Model;
+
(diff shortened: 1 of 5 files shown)


Orrison (changes requested):
Let's go ahead and revert the Model / Filament resource additions here.

When we add coming soon pages we typically just add a simple single Filament Page with our coming soon blade.

Since we have no other details at this time for this feature we don't even fully know if there is going to be an entity called Sequence or how it should work, so we don't want to jump the gun at all here.

Just a simple coming soon page checking the settings view-any permission is all that is needed here.

PR #682, opened 2025-09-29
[AIDAPP-705]: Implement display settings for default time zone display in the application and via emails

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-705

### Technical Description

> Implement display settings for default time zone display in the application and via emails

> No

> DisplaySettingsFeature

### DIFF
--- a/app-modules/service-management/src/Notifications/Concerns/HandlesServiceRequestTemplateContent.php
+++ b/app-modules/service-management/src/Notifications/Concerns/HandlesServiceRequestTemplateContent.php
@@ -87,8 +87,8 @@ trait HandlesServiceRequestTemplateContent
         return [
             'contact name' => $this->serviceRequest->respondent->{$this->serviceRequest->respondent::displayNameKey()},
             'service request number' => $this->serviceRequest->service_request_number,
-            'created date' => !is_null($timezone) ? $this->serviceRequest->created_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $this->serviceRequest->created_at->format('M j, Y \a\t h:i A (T)'),
-            'updated date' => !is_null($timezone) ? $this->serviceRequest->updated_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $this->serviceRequest->updated_at->format('M j, Y \a\t h:i A (T)'),
+            'created date' => ! is_null($timezone) ? $this->serviceRequest->created_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $this->serviceRequest->created_at->format('M j, Y \a\t h:i A (T)'),
+            'updated date' => ! is_null($timezone) ? $this->serviceRequest->updated_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $this->serviceRequest->updated_at->format('M j, Y \a\t h:i A (T)'),
             'assigned staff name' => $this->serviceRequest->assignedTo->user->name ?? 'Unassigned',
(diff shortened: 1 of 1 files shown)


app/Filament/Pages/ManageDisplaySettings.php
```diff
@@ -0,0 +1,97 @@
...
+            ->schema([
+                TimezoneSelect::make('timezone')
+                    ->helperText('Default: ' . config('app.timezone')),
+            ])
+            ->disabled(! auth()->user()->can('product_admin.*.update'));
+    }
+
+    public function save(): void
+    {
+        if (! auth()->user()->can('product_admin.*.update')) {
```
rebecca-canyon, line 79: Same as above, I think this needs to be a `settings` check.

rebecca-canyon (changes requested):

database/migrations/2025_09_29_154502_seed_permission_add_manage_display_settings.php.php
Orrison, line 1: Are we using these permissions anywhere?

And in Advising App were they removed later?

I feel like I am not seeing these permissions being used anywhere in this PR and also the permission name structure does not match the current pattern we have now of only CRUD type permissions.

database/migrations/2025_09_29_151515_data_activate_display_settings_feature_flag.php
Orrison, line 1: This activation has an earlier date then the other two migrations, it would need to come AFTER for the feature flag protection to work properly.

database/settings/2025_09_29_110507_create_display_settings.php
```diff
@@ -0,0 +1,49 @@
...
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        $this->migrator->add('display.timezone');
+    }
+
+    public function down(): void
+    {
+        $this->migrator->delete('display.timezone');
+    }
```
Orrison, line 48: If you can look at other settings adding migrations, there is a try/catch wrapper we need to put around the add to protect against issues of folks loading pages too early.

Orrison (changes requested):

database/migrations/2025_09_29_151515_data_activate_display_settings_feature_flag.php
jayushi-canyon: This seems to be little odd! but will make the required changes.

database/migrations/2025_09_29_154502_seed_permission_add_manage_display_settings.php.php
jayushi-canyon: Yes this was added in advising app and was removed later.

Orrison (changes requested):
The final piece of this ticket as listed at the end of it.

"Finally, make sure the created date and update date merge tags for Service Request Type emails use this to output the values in the timezone provided."

Please make sure you do so. Any of the Service Request Template emails that display a date or datetime need to have this timezone applied to it.

If the email is being sent to a User then it needs to be the timezone they have selected, else fall back to the tenant timezone, else the default.
Because the emails will be sent often not in a context where the User is logged in (like in the queue) you should probably add a new method called like `getTimezoneForUser` that allows you to pass in a User or null and get the timezone similar to the `getTimezone` method, but not relying on the USer being logged in during the request.

Please do work along with @payal-canyon on this one as she has a dependent ticket that needs to make use of this.

danharrin:
Yeah I agree with @Orrison's suggestion

PR #971, opened 2026-04-02
[AIDAPP-924]: Combine the Engagement Actions into a single action where possible

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-924

### Technical Description

> Combine the Engagement Actions into a single action where possible

> No

> No

---


payal-canyon (dismissed):

danharrin (dismissed):

app-modules/engagement/src/Filament/Actions/SendEngagementAction.php
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @return Collection<int, Model&CanBeNotified>
+     */
+    protected function resolveRecipients(array $data): Collection
+    {
+        if ($this->resolveRecipientsUsing) {
+            return $this->evaluate($this->resolveRecipientsUsing);
+        }
+
+        $contact = Contact::find($data['concern_id']);
```
payal-canyon: The fallback path here is copied from Olympus, but Olympus also includes `EngagementContactSelect` in the form schema, which is what populates `$data['concern_id']`. This PR copied the fallback without that select, so the fallback can never actually execute - the only caller (`EngagementsRelationManager`) always provides `resolveRecipientsUsing()`. So, Delete the entire `resolveRecipients()` method.
Inline the call in createEngagement() - replace:
`$recipient = $this->resolveRecipients($data)->first();`
with:
`$recipient = $this->evaluate($this->resolveRecipientsUsing)?->first();`

app-modules/engagement/src/Filament/Schemas/Components/EngagementSendLaterToggle.php
```diff
@@ -0,0 +1,48 @@
...
+namespace AidingApp\Engagement\Filament\Schemas\Components;
+
+use Filament\Forms\Components\Toggle;
+
+class EngagementSendLaterToggle
+{
+    public static function make(): Toggle
+    {
+        return Toggle::make('send_later')
+            ->reactive();
```
payal-canyon, line 47: The extracted `EngagementSendLaterToggle::make()` doesn't carry a `helperText`. `SendEngagementAction` chains its own ("By default, this email will send as soon as..."), but `BulkEngagementAction` doesn't - so the bulk step lost the "email or text" helper copy it used to show. Please add `->helperText(...)` at the bulk call site.

app-modules/engagement/src/Filament/Schemas/Components/EngagementChannelSelect.php
```diff
@@ -0,0 +1,54 @@
...
+
+class EngagementChannelSelect
+{
+    public static function make(): Select
+    {
+        return Select::make('channel')
+            ->label('How would you like to send this engagement?')
+            ->options(NotificationChannel::getEngagementOptions())
+            ->default(NotificationChannel::Email->value)
+            // ->disableOptionWhen(fn (string $value): bool => NotificationChannel::tryFrom($value)?->getCaseDisabled())
```
payal-canyon: `getCaseDisabled()` isn’t defined anywhere in the codebase, so this line appears to be unused. Let’s remove it.

app-modules/engagement/src/Filament/Actions/BulkEngagementAction.php
payal-canyon: I understand you didn’t add this, but let’s maintain consistency. The merge tags are currently hardcoded as `['contact full name', 'contact email']`. Since `SendEngagementAction::getDefaultMergeTags()` already delegates to `Engagement::getMergeTags()`, the bulk path should use the same source of truth so both actions stay in sync.

payal-canyon (dismissed):

danharrin (approved):

PR #994, opened 2026-04-10
[AIDAPP-1021]: Update the interface for the knowledge base to improve usability

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1021

### Technical Description

Add view page for knowledge base items, utilizing tabs; change edit experience to match, also using tabs

No

No

---


app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItem.php
```diff
@@ -65,61 +74,136 @@ public function form(Schema $schema): Schema
...
+                                    )
+                                    ->searchable()
+                                    ->preload()
+                                    ->multiple()
+                                    ->columnSpanFull(),
+                                Toggle::make('has_table_of_contents')
+                                    ->label('Table of Contents')
+                                    ->default(false)
+                                    ->onColor('success')
+                                    ->offColor('gray'),
```
Orrison: Let's put this under the title like it is in the current version

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItem.php
```diff
@@ -65,61 +74,136 @@ public function form(Schema $schema): Schema
...
+                                Select::make('division')
+                                    ->label('Division')
+                                    ->multiple()
+                                    ->relationship('division', 'name')
+                                    ->searchable(['name', 'code'])
+                                    ->preload()
+                                    ->default(fn () => Division::count() === 1 ? Division::query()->first()?->getKey() : null)
+                                    ->visible(fn (): bool => Division::count() > 1)
+                                    ->dehydratedWhenHidden()
+                                    ->exists((new Division())->getTable(), (new Division())->getKeyName()),
```
Orrison, line 208: Can we do a quick confirmation that this is working (and in the create process as well) Or if there is maybe a problem with displaying it on the View end? As I am not seeing the a Division display on the view page. But it should always have had a division?

Orrison (changes requested):
The list page needs to be updated to remove the Edit action from the row actions and make it so clicking on a record goes to the view page. Right now the view page is not accessible from the list page.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItem.php
```diff
@@ -65,61 +74,136 @@ public function form(Schema $schema): Schema
...
+                                    ->exists((new KnowledgeBaseCategory())->getTable(), (new KnowledgeBaseCategory())->getKeyName()),
+                                Select::make('division')
+                                    ->label('Division')
+                                    ->multiple()
+                                    ->relationship('division', 'name')
+                                    ->searchable(['name', 'code'])
+                                    ->preload()
+                                    ->default(fn () => Division::count() === 1 ? Division::query()->first()?->getKey() : null)
+                                    ->visible(fn (): bool => Division::count() > 1)
+                                    ->dehydratedWhenHidden()
```
Orrison: Do we need to make the same change that you did in the create, here?

Orrison (changes requested):

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItem.php
rebecca-canyon: Not exactly the same, but it did need to be different; thank you for catching this!

Orrison (approved):

PR #1011, opened 2026-04-14
[AIDAPP-1010]: Some users have reported that images fail to copy over to knowledge base articles from Microsoft Word

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1010
- https://canyongbs.atlassian.net/browse/AIDAPP-1056

### Technical Description

Updates Filament to fix the issue, also makes a few other places more consistent with Advising App and Olympus engagements.

No

`app-modules/engagement/database/migrations/2026_04_14_091532_tmp_data_process_rich_content_in_engagement_tables.php`

---

### DIFF
--- a/app-modules/engagement/database/migrations/2026_04_14_091532_tmp_data_process_rich_content_in_engagement_tables.php
+++ b/app-modules/engagement/database/migrations/2026_04_14_091532_tmp_data_process_rich_content_in_engagement_tables.php
@@ -17,7 +17,7 @@
       in the software, and you may not remove or obscure any functionality in the
       software that is protected by the license key.
     - You may not alter, remove, or obscure any licensing, copyright, or other notices
-      of the licensor in the software. Any use of the licensor's trademarks is subject
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
       to applicable law.
     - Canyon GBS LLC respects the intellectual property rights of others and expects the
       same in return. Canyon GBS™ and Aiding App™ are registered trademarks of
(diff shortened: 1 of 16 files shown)


app-modules/engagement/database/migrations/2026_04_14_091532_tmp_data_process_rich_content_in_engagement_tables.php
```diff
@@ -0,0 +1,123 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Support\Facades\DB;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        $tables = [
+            ['table' => 'email_templates', 'column' => 'content'],
+            ['table' => 'engagements', 'column' => 'subject'],
```
payal-canyon: Both the engagements and engagement_batches tables store the subject as a plain string, so we can remove it.

payal-canyon (commented):
When viewing an engagement, the styling does not appear, but it works fine in the email. Are you experiencing the same issue?

Image

Image

app-modules/engagement/database/migrations/2026_04_14_091532_tmp_data_process_rich_content_in_engagement_tables.php
danharrin: Good catch, fixed

payal-canyon (approved):

PR #1016, opened 2026-04-15
[AIDAPP-1041]:Enhance the service request report with service request categorization

mittal-canyon:
…description change

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1041

### Technical Description

> Enhance the service request report with service request categorization

> NO

> NO

---


Orrison (dismissed):
There are a lot of changes in here that look like they are from other branches / PRs.

Please properly rebase this branch on `main` and make sure you only include your commits / changes.

You should mostly use rebase here, this kind of stuff is often caused by trying to merge main in normally.
If you need help with rebasing let the team know.

When this is ready again please request review from @payal-canyon or @danharrin as I will be OOO

app-modules/report/src/Filament/Widgets/ServiceRequestsTable.php
```diff
@@ -84,6 +86,7 @@ public function table(Table $table): Table
     {
         $startDate = $this->getStartDate();
         $endDate = $this->getEndDate();
+        $category = $this->pageFilters['category'] ?? null;
```
payal-canyon: The page `app-modules/report/src/Filament/Pages/ServiceRequests.php` only defines `startDate` and `endDate` as page filters. There is no category page filter, so `$this->pageFilters['category']` is always null and this `when()` never runs. The actual filtering is already handled by the `SelectFilter::make(...)` in `ServiceRequestsTable.php`

delete the `$category = ...` line and the `->when(filled($category), ...)` clause. Filtering via `SelectFilter` alone is what was requested.

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestTypesTableTest.php
```diff
@@ -140,3 +141,64 @@
...
+    livewire(ServiceRequestTypesTable::class, [
+        'cacheTag' => 'test-service-request-types-incident-request-counts',
+        'pageFilters' => [],
+    ])
+        ->assertTableColumnFormattedStateSet('service_requests_count', '8', record: $type)
+        ->assertTableColumnFormattedStateSet('incident_count', '3', record: $type)
+        ->assertTableColumnFormattedStateSet('request_count', '5', record: $type);
+});
+
+it('incident and request counts are not affected by the category page filter', function () {
```
payal-canyon: The `pageFilters = ['category' => ...]` key is never present in the page filters, so this test is asserting a scenario that cannot occur. It may give a false sense of confidence. Could you please remove it?

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestCategoryDistributionDonutChartTest.php
```diff
@@ -0,0 +1,197 @@
...
+        'priority_id' => $priority->id,
+        'status_id' => $status->id,
+        (ServiceRequestCategoryRenameFeature::active() ? 'category' : 'issue_category') => ServiceRequestCategory::Request,
+        'created_at' => now()->subDays(2),
+    ])->create();
+
+    $widget = new ServiceRequestCategoryDistributionDonutChart();
+    $widget->cacheTag = 'test-category-distribution-category-filter';
+    // Even with a stale category in pageFilters, the chart ignores it
+    $widget->pageFilters = ['category' => ServiceRequestCategory::Incident->value];
```
payal-canyon: The `pageFilters = ['category' => ...]` key is never present in the page filters, so this test is asserting a scenario that cannot occur. It may give a false sense of confidence. Could you please remove it?

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
```diff
@@ -0,0 +1,156 @@
...
+            'danger' => 'rgb(239, 68, 68)',     // red-500
+            'primary' => 'rgb(59, 130, 246)',   // blue-500
+            'success' => 'rgb(34, 197, 94)',    // green-500
+            'warning' => 'rgb(245, 158, 11)',   // amber-500
+            'info' => 'rgb(14, 165, 233)',      // sky-500
+            'gray' => 'rgb(107, 114, 128)',     // gray-500
+        ];
+
+        return $colorMap[$color] ?? 'rgb(59, 130, 246)';
+    }
```
payal-canyon: drop `colorToRgb()` from the chart, and add this to `ServiceRequestCategory.php`

```javascript
public function getRgb(): string
{
    return Color::convertToRgb(FilamentColor::getColors()[$this->getColor()][500]);
}
```

Then we can simply call `'bg_color' => $case->getRgb()`

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestTypesTableTest.php
```diff
@@ -140,3 +141,64 @@
...
+
+    $status = ServiceRequestStatus::factory()->state([
+        'classification' => SystemServiceRequestClassification::Open,
+    ])->create();
+
+    ServiceRequest::factory()->count(3)->state([
+        'priority_id' => $priority->id,
+        'status_id' => $status->id,
+        (ServiceRequestCategoryRenameFeature::active() ? 'category' : 'issue_category') => ServiceRequestCategory::Incident,
+    ])->create();
```
payal-canyon, line 151: This test proves the invariant (3 + 5 = 8) only incidentally. Kevin's concern was specifically that Incidents + Requests must always equal Count under a date filter. Please add a second test that creates incidents and requests both inside and outside a date window, applies startDate/endDate via pageFilters, and asserts all three columns reflect only the in-window records (so the invariant incident_count + request_count == service_requests_count holds). That locks the guarantee in the same scenario where it originally broke - if someone later edits one of the three withCount closures and forgets to apply the date filter, this test will catch it.

app-modules/report/composer.json
```diff
@@ -5,6 +5,7 @@
     "version": "1.0",
     "license": "proprietary",
     "require": {
+        "canyon-gbs/aiding-app-service-management": "*",
```
payal-canyon: Could you please clarify why this was added? If it was included by mistake, please revert the changes to composer.json.

payal-canyon (changes requested):

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestTypesTableTest.php
mittal-canyon: Resolved

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestCategoryDistributionDonutChartTest.php
mittal-canyon: Resolved

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
mittal-canyon: Resolved

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestTypesTableTest.php
mittal-canyon: Resolved

app-modules/report/composer.json
mittal-canyon: Resolved

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
```diff
@@ -0,0 +1,139 @@
...
+
+namespace AidingApp\Report\Filament\Widgets;
+
+use AidingApp\Report\Filament\Widgets\Concerns\InteractsWithPageFilters;
+use AidingApp\ServiceManagement\Enums\ServiceRequestCategory;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+use App\Features\ServiceRequestCategoryRenameFeature;
+use Carbon\Carbon;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Collection as BaseCollection;
```
payal-canyon: Both refer to the same class, so we can remove the `BaseCollection` alias and use `Collection` consistently.

payal-canyon (changes requested):

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
mittal-canyon: Resolved

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * @return Collection<int, array{label: string, count: int, bg_color: string}>
+     */
+    private function getServiceRequestCategoryData(?Carbon $startDate = null, ?Carbon $endDate = null): Collection
+    {
+        $counts = [];
+
+        $categoryColumn = ServiceRequestCategoryRenameFeature::active() ? 'category' : 'issue_category';
```
payal-canyon: As I recall, the `ServiceRequestCategoryRenameFeature` is already enabled in both staging and production environments. Given that, it may no longer be necessary to wrap this logic in a condition. You can use `category` directly instead.

Could you please update this here, as well as in other places within this PR where `ServiceRequestCategoryRenameFeature` is being used, and remove those checks? Since the feature is fully enabled, keeping these conditions might be redundant.

payal-canyon (changes requested):

app-modules/service-management/src/Models/ServiceRequestType.php
```diff
@@ -59,6 +59,9 @@
 use OwenIt\Auditing\Contracts\Auditable;
 
 /**
+ * @property ServiceRequestCategory|null $default_issue_category
+ * @property ServiceRequestCategory|null $default_category
+ *
```
payal-canyon: Once the changes related to the feature flag are completed, could you please remove these lines and regenerate the `ide_helper_file`? After that, kindly run the linter again-only re-add these lines if the linting requires them; otherwise, they can be left out.

The same approach applies to the `ServiceRequest` model as well.

mittal-canyon: Resolved

app-modules/report/src/Filament/Widgets/ServiceRequestCategoryDistributionDonutChart.php
mittal-canyon: Ahh! it wasn't when i had updated the changes. Resolved now.
Thanks!

payal-canyon (approved):

payal-canyon:
@danharrin , One last check from your side would be great - could you please take a final look? Thank You!

danharrin (approved):

PR #1020, opened 2026-04-16
[AIDAPP-926]: Enhance the text editor for Service Management Email Templates

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-926

### Technical Description

> Update service request type email templates to support rich content by adding a migration to make the body field nullable, refactoring content and subject generation to use RichContentRenderer, updating button blocks to extend RichContentCustomBlock, enhancing the page to use RichEditor with custom blocks and plugins, implementing merge tags for dynamic content, and adding tests to validate migration and content integrity.

> No

> Yes. Temporary Data Migration: `app-modules/service-management/database/migrations/2026_04_16_095944_tmp_data_migrate_service_request_type_email_templates_to_rich_editor_format.php`
In `tests/TenantMigrationTests.php`, we need to remove the test case related to the temporary migration (please do not remove the entire file).

---


app-modules/service-management/src/Actions/GenerateServiceRequestTypeEmailTemplateSubject.php
```diff
@@ -46,14 +46,12 @@ class GenerateServiceRequestTypeEmailTemplateSubject
      * @param string|array<int, string|array<string, mixed>> $content
      * @param array<string, mixed> $mergeData
      */
-    public function __invoke(string|array $content, array $mergeData, Model $record, string $recordAttribute): HtmlString
+    public function __invoke(string|array $content, array $mergeData): HtmlString
```
danharrin: Maybe we should move these subject/content action classes into getter methods on the model like we do for engagements?

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ServiceRequestTypeEmailTemplatePage.php
danharrin, line 1: Maybe use `dehydrateStateUsing()` to prevent the content from being saved if it is empty, return `null` if it is empty or the content if it isn't? Then you won't need to get the state & the raw state at the same time when saving, I think?

app-modules/service-management/src/Notifications/Concerns/HandlesServiceRequestTemplateContent.php
danharrin, line 1: I might be wrong, and I know that this code existed before the upgrade, but are we sure that we need `injectButtonUrlIntoContent()`, when you can pass data to custom blocks when they are rendered like this? https://filamentphp.com/docs/5.x/forms/rich-editor#rendering-content-with-custom-blocks

danharrin (changes requested):

app-modules/service-management/src/Notifications/Concerns/HandlesServiceRequestTemplateContent.php
payal-canyon: Good catch! You're right, I've refactored to use Filament's native customBlocks() data-passing pattern. Now the URLs are passed as runtime data via `$data['serviceRequestUrl']` and `$data['feedbackUrl']` in the custom blocks' `toHtml()` methods, and the model's `getBody()` accepts a `$customBlockData` array that's forwarded to `->customBlocks([Class::class => $data])`. The `injectButtonUrlIntoContent()` method has been removed entirely - no more recursive JSON mutation. Thanks for the suggestion!

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ServiceRequestTypeEmailTemplatePage.php
danharrin, line 1: I think we need to make the body content toolbar consistent with the engagement one (no video, details, grid etc) because those elements aren't supported in emails, right?

danharrin (changes requested):

app-modules/service-management/database/migrations/2026_04_16_095944_tmp_data_migrate_service_request_type_email_templates_to_rich_editor_format.php
danharrin, line 1: I think this migration can now be simplified because of the removed tools we no longer need to handle. Those nodes should be removed if encountered.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ServiceRequestTypeEmailTemplatePage.php
```diff
@@ -98,79 +111,138 @@ public function form(Schema $schema): Schema
...
+        $hasContent = false;
+
+        $walk = function (array $node) use (&$walk, &$hasContent): void {
+            if (! empty($node['text'])) {
+                $hasContent = true;
+
+                return;
+            }
+
+            if (in_array($node['type'] ?? null, ['image', 'customBlock', 'videoEmbed', 'horizontalRule', 'hardBreak'], true)) {
```
danharrin: I think we can remove video from this, and potentially hard break?

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ServiceRequestTypeEmailTemplatePage.php
```diff
@@ -98,79 +111,138 @@ public function form(Schema $schema): Schema
...
 
-    /** @return array<int, TiptapEditor> */
+    /**
+     * Returns null when the given rich content is an empty document (no text, images, or custom blocks).
+     *
+     * @param mixed $content
+     *
+     * @return mixed
+     */
+    protected function normalizeRichContent($content)
```
danharrin: I think we can use the `mixed` param and return type on this method instead of needing the docblock

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ServiceRequestTypeEmailTemplatePage.php
```diff
@@ -98,79 +111,138 @@ public function form(Schema $schema): Schema
...
+                    ['bold', 'italic', 'link'],
+                    [ToolbarButtonGroup::make('Heading', ['h1', 'h2', 'h3', 'h4', 'h5', 'h6'])->textualButtons(), 'bulletList', 'orderedList', 'horizontalRule'],
+                    ['textColor', 'small'],
+                    ['attachFiles', 'mergeTags', 'customBlocks'],
+                    ['clearFormatting'],
+                    ['undo', 'redo'],
+                ])
                 ->mergeTags($mergeTags)
-                ->blocks([
+                ->activePanel('mergeTags')
```
danharrin: I don't think the merge tags panel was active by default on the old editor so we shouldn't use this, the user might want a custom block instead

app-modules/service-management/src/Models/ServiceRequestTypeEmailTemplate.php
```diff
@@ -79,4 +90,101 @@ public function serviceRequestType(): BelongsTo
...
+        // Convert CSS variable-based styles to inline styles for email client compatibility.
+        // Text colors: style="--color: #hex; --dark-color: #hex" → style="color: #hex"
+        $html = preg_replace(
+            '/style="--color:\s*([^;]+);\s*--dark-color:\s*[^"]*"/',
+            'style="color: $1"',
+            $html,
+        );
+
+        // Grid layout: style="--cols: repeat(X, ...)" → style="display: table; width: 100%"
+        $html = preg_replace(
```
danharrin: I think we can remove this handling since grids aren't supported

app-modules/service-management/src/Notifications/Concerns/HandlesServiceRequestTemplateContent.php
```diff
@@ -53,16 +52,12 @@ trait HandlesServiceRequestTemplateContent
...
-        }
-
-        return app(GenerateServiceRequestTypeEmailTemplateContent::class)(
-            $body,
-            $this->getMergeData($timezone),
-            $this->serviceRequest,
-            'body',
-        );
+        return $this->emailTemplate->getBody($body, $this->getMergeData($timezone), [
+            'serviceRequestUrl' => $urlType === ServiceRequestTypeEmailTemplateRole::Customer
```
danharrin: I don't mind this much, but is there a way we don't need to pass both data params to both blocks, and each block only receives the data it needs? Maybe just have dedicated service request URL and feedback URL params on `getBody()` and then pass the correct one into each block. My worry is that in the future we might have more blocks and they might need the same data param names but different data values, in which case we would have to refactor this, which is avoidable.

danharrin (changes requested):

danharrin:
There is also a merge conflict for the migration tests, sorry.

danharrin (approved):

PR #1029, opened 2026-04-20
[AIDAPP-1040]: Introduce the ability to support file attachments in service request updates & [AIDAPP-1104]: Some users have reported that buttons are not displaying in emails

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1040
- https://canyongbs.atlassian.net/browse/AIDAPP-1104

### Technical Description

Add ability to upload media attachments to service request updates in admin panel and portal; account for media in service request updates in related merge tag.

No

No

---


app-modules/service-management/src/Models/ServiceRequestUpdate.php
```diff
@@ -82,6 +87,35 @@ class ServiceRequestUpdate extends BaseModel implements Auditable, ProvidesATime
         'update_type' => ServiceRequestUpdateType::class,
     ];
 
+    public function registerMediaCollections(): void
+    {
+        $this->mediaCollections[] = UploadsMediaCollection::create()
```
Orrison: Is there a particular reason we are adding the media collection in this array assignment way rather than using the fluent `$this->addMediaCollection` collection pattern from the docs?

https://spatie.be/docs/laravel-medialibrary/v11/working-with-media-collections/defining-media-collections

app-modules/service-management/src/Models/ServiceRequestUpdate.php
```diff
@@ -82,6 +87,35 @@ class ServiceRequestUpdate extends BaseModel implements Auditable, ProvidesATime
...
+                'text/markdown' => ['md', 'markdown', 'mkd'],
+                'text/plain' => ['txt', 'text'],
+                'application/octet-stream' => ['log'],
+                '.log' => ['log'],
+                'video/mp4' => ['mp4'],
+                'video/webm' => ['webm'],
+                'video/ogg' => ['ogg'],
+                'video/quicktime' => ['quicktime'],
+                'video/x-msvideo' => ['x-msvideo'],
+            ]);
```
Orrison, line 115: We should also define the disk here: https://spatie.be/docs/laravel-medialibrary/v11/working-with-media-collections/defining-media-collections#content-using-a-specific-disk

That way we don't have to define that in other places (like where we are adding the media as I mention on some other comments.)

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
```diff
@@ -55,6 +55,14 @@ public function __invoke(StoreServiceRequestUpdateRequest $request): JsonRespons
 
         $serviceRequestUpdate->save();
 
+        if ($request->hasFile('files')) {
+            foreach ($request->file('files') as $file) {
+                $serviceRequestUpdate
+                    ->addMedia($file)
+                    ->toMediaCollection('uploads', config('media-library.disk_name'));
```
Orrison: If you define the disk in the collection definition within the model itself you won't have to (and better yet, shouldn't) define that here

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
```diff
@@ -68,9 +76,28 @@ public function __invoke(StoreServiceRequestUpdateRequest $request): JsonRespons
...
+     */
+    public function rules(): array
+    {
+        return [
+            'description' => ['required', 'string'],
+            'serviceRequestId' => ['required', 'exists:service_requests,id'],
+            'files' => ['nullable', 'array'],
+            'files.*' => ['file'],
+        ];
+    }
```
Orrison: What is this method doing / being used for? Can you define rules directly in invokable controllers now?

If not, I think you may have intended to add this using a validator in the controller or via a custom form request?
Regardless of the choice (or if this really does work) we should probably add some tests ensuring this validation is working.

app-modules/service-management/src/Filament/Resources/ServiceRequests/RelationManagers/ServiceRequestUpdatesRelationManager.php
```diff
@@ -71,6 +74,8 @@ class ServiceRequestUpdatesRelationManager extends RelationManager
 
     public function form(Schema $schema): Schema
     {
+        $uploadsMediaCollection = app(ResolveUploadsMediaCollectionForServiceRequest::class)->__invoke();
```
Orrison: Idk why we made this class, it's not worth a whole class for what this is doing / returning, I would just do it in an inline variable.

Also this is returning the wrong details. This is for the Service Request media collection, not the one you defined for ServiceRequestUpdate.

app-modules/service-management/src/Filament/Resources/ServiceRequestUpdates/Pages/ViewServiceRequestUpdate.php
```diff
@@ -38,26 +38,33 @@
...
+use Filament\Support\Enums\IconSize;
+use Spatie\MediaLibrary\MediaCollections\Models\Media;
 
 class ViewServiceRequestUpdate extends ViewRecord
 {
     protected static string $resource = ServiceRequestUpdateResource::class;
 
     public function infolist(Schema $schema): Schema
     {
+        $uploadsMediaCollection = app(ResolveUploadsMediaCollectionForServiceRequest::class)->__invoke();
```
Orrison: Same as my other comment

app-modules/service-management/src/Filament/Resources/ServiceRequestUpdates/Pages/ViewServiceRequestUpdate.php
```diff
@@ -87,6 +94,51 @@ public function infolist(Schema $schema): Schema
...
+                                    ->color('primary')
+                                    ->url($media->getTemporaryUrl(now()->addMinute()), true);
+
+                                if ($isImage) {
+                                    return ImageEntry::make($media->getKey())
+                                        ->label($media->name)
+                                        ->visibility('private')
+                                        ->getStateUsing($media->getTemporaryUrl(now()->addMinute()))
+                                        ->hintAction($downloadAction);
+                                }
```
Orrison, line 135: You may want to set what you are doing in `$downloadAction` within a function for `hintAction`. That may work and execute properly on click.

Without doing so, this will not work after a minute of the page loading.

app-modules/service-management/src/Filament/Resources/ServiceRequestUpdates/ServiceRequestUpdateResource.php
```diff
@@ -69,6 +72,8 @@ class ServiceRequestUpdateResource extends Resource
 
     public static function form(Schema $schema): Schema
     {
+        $uploadsMediaCollection = app(ResolveUploadsMediaCollectionForServiceRequest::class)->__invoke();
```
Orrison: Same as other comments on this

Orrison (changes requested):

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
```diff
@@ -46,6 +46,13 @@ class StoreServiceRequestUpdateController extends Controller
 {
     public function __invoke(StoreServiceRequestUpdateRequest $request): JsonResponse
     {
+        $request->validate([
+            'description' => ['required', 'string'],
+            'serviceRequestId' => ['required', 'uuid', 'exists:service_requests,id'],
+            'files' => ['nullable', 'array'],
+            'files.*' => ['file'],
+        ]);
```
Orrison: There is already a custom form request for this controller method `StoreServiceRequestUpdateRequest` that has a `rules` method defined for validation.

Doing this here is duplicating that.

You should just update the `rules` method in `StoreServiceRequestUpdateRequest` for what you need.

app-modules/service-management/src/Filament/Resources/ServiceRequests/RelationManagers/ServiceRequestUpdatesRelationManager.php
```diff
@@ -101,6 +103,16 @@ public function form(Schema $schema): Schema
...
                         return $this->getOwnerRecord()->status->getKey();
                     }),
+                Section::make('Uploads')
+                    ->schema([
+                        SpatieMediaLibraryFileUpload::make('uploads')
+                            ->hiddenLabel()
+                            ->visibility('private')
+                            ->collection('uploads')
+                            ->multiple(true)
+                            ->acceptedFileTypes(app(ServiceRequestUpdate::class)->getMediaCollection('uploads')->acceptsMimeTypes)
```
Orrison: It's not a huge deal. But there really isn't any reason to be resolving the class from the container here (and in other places you do this in this PR) i.e. `app(ServiceRequestUpdate::class)`

You really only do that with interfaces / facades to resolve their correct implementations OR if you need to override the class for mocking in tests.

Since you are just trying to get details from the model here I would just new up the class `new ServiceRequestUpdate()->...`

app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -461,7 +462,15 @@ public function getRecentUpdateFormatted(?string $timezone = null): string
 
         $updateDate = ! is_null($timezone) ? $recentUpdate->created_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $recentUpdate->created_at->format('M j, Y \a\t h:i A (T)');
 
-        return "{$creatorInfo}\n\n{$updateDate} - {$recentUpdate->update}";
+        $uploadsMediaCollection = app(ResolveUploadsMediaCollectionForServiceRequest::class)->__invoke();
```
Orrison: This is still calling the incorrect media collection class thing for the Service Request, not the SRUpdate class.

Just a heads up, I assume this doesn't throw errors ONLY because the collections are both called `uploads` in both places.
But this will still be checking the wrong details in some instances.

You kind of get away with it here because all your are pulling out is the name, which like I said, is the same. But we should remain correct / consistent.

Orrison (changes requested):

Orrison (approved):

PR #1035, opened 2026-04-22
[AIDAPP-1061] : Ensure the user lands on the initial page they attempted to navigate to when visiting a link when first unauthenticated

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1061

### Technical Description

> Ensure the user lands on the initial page they attempted to navigate to when visiting a link when first unauthenticated

> NO

> NO

---

### DIFF
--- a/app-modules/authorization/src/Http/Controllers/SocialiteController.php
+++ b/app-modules/authorization/src/Http/Controllers/SocialiteController.php
@@ -56,18 +56,18 @@ class SocialiteController extends Controller
 {
     public function redirect(SocialiteProvider $provider, Request $request)
     {
+        $intendedUrl = $request->session()->pull('url.intended');
+
         // Regenerate session and logout user to try to fix InvalidStateException
         if ($request->hasSession()) {
-            $intendedUrl = $request->session()->get('url.intended');
-
             $request->session()->regenerate(true);
-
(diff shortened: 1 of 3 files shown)


app-modules/authorization/src/Http/Controllers/SocialiteController.php
```diff
@@ -58,7 +58,13 @@ public function redirect(SocialiteProvider $provider, Request $request)
     {
         // Regenerate session and logout user to try to fix InvalidStateException
         if ($request->hasSession()) {
+            $intendedUrl = $request->session()->get('url.intended');
+
             $request->session()->regenerate(true);
+
+            if (filled($intendedUrl)) {
```
danharrin: I think this needs to happen after the logout on line 70, and outside of the if statement. Please see the PR I linked https://github.com/canyongbs/advisingapp/pull/2479/changes

danharrin (changes requested):

app-modules/authorization/src/Http/Controllers/SocialiteController.php
mittal-canyon: Resolved

danharrin (approved):

PR #1046, opened 2026-04-24
[AIDAPP-990]: Remove Prompt Model

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-990

### Technical Description

> Remove Prompt Model

> NO

> NO

---


danharrin (changes requested):
Please revert all changes to migrations, migrations are permanent. If tables or columns need dropping, please create new migrations

app-modules/ai/database/migrations/2026_04_24_000000_drop_prompt_tables.php
```diff
@@ -34,27 +34,33 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::dropIfExists('prompt_upvotes');
+            Schema::dropIfExists('prompt_uses');
+
+            if (Schema::hasColumn('ai_messages', 'prompt_id')) {
```
danharrin, line 49: Are you sure we need this? Do we have existing migrations to drop these tables and columns? If not then we shouldn't need these checks?

danharrin (changes requested):

app-modules/ai/database/migrations/2026_04_24_000000_drop_prompt_tables.php
danharrin: I see migrations that already exist for dropping prompts and prompt types, and I don't think it would even be possible to have a `prompt_id` FK if those tables were already dropped?

danharrin: It seems locally, the prompts table still exists even though there is a migration to drop it already in the codebase. I wonder what happened there...

danharrin (approved):

app-modules/ai/database/migrations/2026_04_24_000000_drop_prompt_tables.php
mittal-canyon: Yes, that's why I had to add!
Thanks!

PR #1051, opened 2026-04-24
[AIDAPP-1075]: Enhance the timeline to display service request lifecycle events

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1075

### Technical Description

> Implement service request history tracking and update timeline components.

> No

> Yes. Feature Flag: `ServiceRequestHistoryActorFeature`

---

### DIFF
--- a/app-modules/service-management/database/migrations/2026_04_24_115625_add_actor_fields_to_service_request_histories_table.php
+++ b/app-modules/service-management/database/migrations/2026_04_24_115625_add_actor_fields_to_service_request_histories_table.php
@@ -42,7 +42,6 @@ return new class () extends Migration {
     public function up(): void
     {
         Schema::table('service_request_histories', function (Blueprint $table) {
-            $table->string('event_type')->nullable();
             $table->nullableUuidMorphs('actor');
         });
     }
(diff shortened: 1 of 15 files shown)


app-modules/service-management/database/migrations/2026_04_24_115625_add_actor_fields_to_service_request_histories_table.php
danharrin, line 1: Do we need a feature flag to protect against this column not existing?

danharrin (changes requested):
@Orrison do we want to investigate larger refactoring opportunities here, for example introducing an abstraction for history entry types, like having a class per type of history entry or something? Probably worthwhile considering product are planning even more events in this service request timeline soon.

Orrison:
> @Orrison do we want to investigate larger refactoring opportunities here, for example introducing an abstraction for history entry types, like having a class per type of history entry or something? Probably worthwhile considering product are planning even more events in this service request timeline soon.

Let's hold off for now and do so when more changes are asked for the SR timeline.

danharrin (approved):

PR #1054, opened 2026-04-27
[AIDAPP-1058]: Backend rename Incidents to Advisories

amit-canyon:
Co-authored-by: Copilot <copilot@github.com>

### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1058

### Technical Description

> Backend rename Incidents to Advisories

> No

> Yes, `IncidentRenameFeature`

---


app-modules/service-management/src/Enums/ServiceRequestCategory.php
```diff
@@ -43,22 +43,22 @@
 
 enum ServiceRequestCategory: string implements HasColor, HasLabel
 {
-    case Incident = 'incident';
+    case Advisory = 'advisory';
```
Orrison: This, and all references to it, actually should NOT be renamed.

This is a different thing than Advisories (previously called Incidents)

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestCategoryDistributionDonutChartTest.php
Orrison, line 1: This should be reverted. `ServiceRequestCategory` is NOT related to this ticket.

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestsTableTest.php
Orrison, line 1: Revert - Not related

app-modules/report/tests/Tenant/Filament/Widgets/ServiceRequestTypesTableTest.php
Orrison, line 1: Revert - Not related

app-modules/report/src/Filament/Widgets/ServiceRequestTypesTable.php
Orrison, line 1: This should be reverted. This is NOT related to the ticket, service request category is unrelated to the Incident/Advisory model

app-modules/service-management/database/migrations/2026_04_24_000001_rename_incident_tables_and_columns_to_advisory.php
```diff
@@ -0,0 +1,91 @@
...
+*/
+
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        if (Schema::hasTable('incident_severities') && ! Schema::hasTable('advisory_severities')) {
```
Orrison: I would recommend removing this `if` statements throughout this migration.

Migrations run sequentially and are typically the ONLY way the database schema is modified. So, unless for some reason the DB was modified manually, migrations should be written assuming the schema exists as expected, as you essentially WANT it to fail to alert you of something being really wrong.

app-modules/service-management/database/migrations/2026_04_24_000003_data_rename_incident_category_to_advisory_in_service_requests.php
Orrison, line 1: I would suggest getting rid of the `if` statements here like above. But actually this entire migration should be dropped as SR categories has nothing to do with this ticket.

app-modules/service-management/database/migrations/2026_04_27_101702_data_activate_incident_rename_feature.php
Orrison, line 1: After you delete the unneeded migration I mentioned above, I would actually recommend combining ALL the migrations here into one single migration encapsulating all schema changes, data changes, and the FF activation at the end.

This is because with the way this is structured now, there is going to be a gap in time between the table / column renames and the FF activation, which could cause some issues.

app-modules/service-management/src/Models/Advisory.php
Orrison, line 1: The strategy you have done, adding `canAccess` methods and preventing access to pages before the migrations are run and the FF is activated is a little problematic.

Primarily the problem as users will lose access to the feature unnecessarily.

To prevent this, you should instead override the `getTable` method as well as update all column references in relationships in the Models (like this one and the other ones changed) to point to the old table / column name while the FF is inactive, and then point to the new table / column name once it IS active.

That way the data is always accessible.

tests/Tenant/Unit/legacy-v4-uuid-models.php
```diff
@@ -229,10 +225,6 @@
     ChangeRequestStatus::class,
     ChangeRequestType::class,
     HistoricalServiceMonitoring::class,
-    Incident::class,
-    IncidentSeverity::class,
-    IncidentStatus::class,
-    IncidentUpdate::class,
```
Orrison, line 235: These can't be removed, the model references need to be updated to the new reference / names. Without this set people would not be prevented from accidentally changing the UUID type.

app/Http/Requests/Tenants/CreateTenantRequest.php
Orrison, line 1: This and the sync request are fine. But the related ticket in ticket will need to be done and both merged in at the same time or else it will break deployment / sync of Tenants. I will pull it into the Sprint and assign it to you.

Orrison (changes requested):
Left some feedback on a few issues. Primarily that the SR category is unrelated and all changes to that should be reverted and left out of these changes, the feature rename needs to happen at the same time as an Olympus change (which I will assign over to you), and some of the migration / gating could be improved.

Orrison (approved):

PR #1055, opened 2026-04-27
[AIDAPP-996] : The service request feedback page does not gate based on access to it's service request

monali-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-996

### Technical Description

- Implement access control for feedback entry and submission in Feedback page

> No

> No

---


danharrin (changes requested):
Do you also get this error when trying to load the feedback page?

Image

Orrison (changes requested):
I also just want to note @danharrin and @monali-canyon, though the changes here could be fine.

This ticket was actually in reference to the Feedback page of a Service Request, in the admin panel.

Apologies if that was not clear.

monali-canyon:
@Orrison @danharrin Can you provide more clarity on this because I'm not able to understand the actual task then?

`app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/Feedback.php` is the file to fix it?

Orrison:
> @Orrison @danharrin Can you provide more clarity on this because I'm not able to understand the actual task then?
> 
> `app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/Feedback.php` is the file to fix it?

@monali-canyon, correct. you should only have access to that page if you have access to the service request it is related to

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/Feedback.php
Orrison, line 1: These changes look pretty good and seem to be working.

But I am thinking we may want to go a little further here to fix the underlying issues with this access for this page and going forward with `ServiceRequestFeedback`.

One issue here is that this page is a `ViewRecord` page targeting the `ServiceRequestResource`. So this is always acting as a Service Request page.

Is there a way we can apply these proper access rules to a `ServiceRequestFeedbackPolicy` and have this page make sure it is looking at that properly.
That way any other place we look at feedback also requires the proper permissions when scopped through the policy.

WDYT @danharrin?

Orrison (changes requested):

app-modules/service-management/src/Policies/ServiceRequestFeedbackPolicy.php
```diff
@@ -0,0 +1,122 @@
...
+        return $authenticatable->canOrElse(
+            abilities: 'service_request.view-any',
+            denyResponse: 'You do not have permission to view service request feedback.'
+        );
+    }
+
+    public function view(Authenticatable $authenticatable, ServiceRequestFeedback $serviceRequestFeedback): Response
+    {
+        return $authenticatable->canOrElse(
+            abilities: ['service_request.*.view', "service_request.{$serviceRequestFeedback->service_request_id}.view"],
```
Orrison, line 70: Here, and in other parts of this file the `{$serviceRequestFeedback->service_request_id}` injection can be removed.
This is left over in some policies back from when we did permissions per entity, which we do not do anymore. the `*` is sufficient.

```suggestion
            abilities: ['service_request.*.view'],
```

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/Feedback.php
```diff
@@ -56,6 +57,11 @@ class Feedback extends ViewRecord
 
     protected static ?string $breadcrumb = 'Feedback';
 
+    public static function canAccess(array $arguments = []): bool
+    {
+        return auth()->user()->can('viewAny', [ServiceRequestFeedback::class, $arguments['record']]);
```
Orrison, line 62: The `viewAny` check is good here, but once you modify the policy method to require passing in the Service Request you may need to change something here. Though possible not since you are already passing it in.

app-modules/service-management/src/Policies/ServiceRequestFeedbackPolicy.php
Orrison, line 1: This is looking good. But in the individual methods here we should be checking if the User has the equivalent ability on the related (or possibly related) service request.

For viewAny and create you are going to need to require, and pass in where it is used, the Service Request Model you are attempting to view these / create these for. (like in `app-modules/project/src/Policies/ProjectMilestonePolicy.php`)

For the other methods, like `view` and others where you have an existing ServiceRequestFeedback model, you can check directly via the Service Request it is related to

Orrison (changes requested):

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/Feedback.php
danharrin: I'm not really sure what you mean to be honest 🤔

PR #1056, opened 2026-04-27
[AIDAPP-1059]: Many User selection inputs exist that allow for the selection of Admin Users

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1059

### Technical Description

> Created a base User select which by default will not list admin in the user select input, but if needed admins can be added in select by calling respective method in select.

> No

> No

---


config/internal-users.php
```diff
@@ -36,4 +36,5 @@
 
 return [
     'emails' => env('DEMO_INTERNAL_USER_EMAILS') ? explode(',', env('DEMO_INTERNAL_USER_EMAILS')) : null,
+    'filter_admins_from_selection' => env('USER_SELECT_FILTER_ADMINS', true),
```
Orrison: Let's actually move this to the main `config/app.php`. I am pretty sure the other config setting in here is not used anymore (left over from some old stuff) so I am going to make a ticket to delete it later.

app/Filament/Forms/Components/UserSelect.php
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    public function relationship(string|Closure|null $name = null, string|Closure|null $titleAttribute = null, ?Closure $modifyQueryUsing = null, bool $ignoreRecord = false): static
+    {
+        return parent::relationship(
+            $name,
+            $titleAttribute ?? 'name',
+            function (Builder $query) use ($modifyQueryUsing) {
+                if ($this->shouldFilterAdmins()) {
+                    $query->tap(new WithoutAnyAdmin());
```
Orrison: Though generally usage of this scope here is fine, one thing we have to consider is that it is possible that people have already selected admin users in these select inputs in the past. Which we now need to allow for the display and de-selection of, but not the display or selection of them if NOT already selected.

Ergo, this will need to account for "without any admins, BUT if the selection / relationship already exists, they should show up

app/Filament/Forms/Components/UserSelect.php
```diff
@@ -0,0 +1,88 @@
...
+        );
+    }
+
+    protected function shouldFilterAdmins(): bool
+    {
+        if (! $this->filterAdmins) {
+            return false;
+        }
+
+        return config('internal-users.filter_admins_from_selection', true);
```
Orrison: I actually think this should return a combination of checking `$this->filterAdmins` AND the config value.

The `filterAdmins` property dictates whether or not the specific usage of this `UserSelect` wants to have admins filtered or not. While the config value enables or disables this for ALL usages of the field.

So both should need to be true for admins to be filtered

Orrison (changes requested):
Left some feedback. I think we should also add some tests specifically for the component testing that it has the proper records depending on the property, config, and/or pre-selected values/relationships

Orrison:
@danharrin I also requested your review on this item if you can please review the component from a Filament perspective while also verifying that my feedback seems good. (just want a second opinion)

Finally, once all changes are complete and both @danharrin and I approve, I will add one final change to this PR adding a document to the docs folder explaining this field and that it should be used whenever there is UI to select a user.

app/Filament/Forms/Components/UserSelect.php
danharrin: For this pattern of allowing already-selected users, we could copy the pattern we have for the "archiving" feature?

danharrin (commented):
From a Filament perspective, the approach seems sane

app/Filament/Forms/Components/UserSelect.php
```diff
@@ -0,0 +1,109 @@
...
+                        $qualifiedKey = $this->getRelationship()->getRelated()->getQualifiedKeyName();
+
+                        $alreadySelected = $record->{$relationshipName}()
+                            ->getQuery()
+                            ->pluck($qualifiedKey)
+                            ->map(fn ($variable) => (string) $variable)
+                            ->toArray();
+                    }
+
+                    $state = $this->getState();
```
danharrin: You should inject `UserSelect $component` into the closure instead of using $this inside of the closure, otherwise this field won't work in repeaters as `$this` always references the last item in the repeater (the component object gets cloned between repeater items).

danharrin (changes requested):

danharrin (dismissed):

danharrin (approved):

Orrison (approved):

PR #1057, opened 2026-04-27
[AIDAPP-1018]: Handle file attachments from inbound emails for Service Request Reply

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1018

### Technical Description

Allow for processing email attachments for service request updates

No

No

---


app-modules/engagement/tests/Landlord/Jobs/ProcessSesS3InboundEmailTest.php
```diff
@@ -1136,6 +1136,92 @@ function (MadeTenantCurrentEvent $event) use ($fakeStorage) {
...
+            assert($serviceRequestUpdate instanceof ServiceRequestUpdate);
+
+            assertDatabaseHas(ServiceRequestUpdate::class, [
+                'service_request_id' => $serviceRequest->getKey(),
+                'internal' => false,
+                'created_by_id' => $contact->getKey(),
+                'created_by_type' => $contact->getMorphClass(),
+            ]);
+
+            expect($serviceRequestUpdate->getMedia('uploads'))->toHaveCount(2);
```
Orrison: If we know what the items are, it could be good to assert more on them, like what the name of the file is, the mime, etc.

app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
```diff
@@ -324,12 +324,19 @@ protected function handleServiceRequestReply(Tenant $tenant, Parser $parser, str
...
                 'internal' => false,
                 'created_by_id' => $serviceRequest->respondent->getKey(),
                 'created_by_type' => $serviceRequest->respondent->getMorphClass(),
             ]);
+
+            foreach ($parser->getAttachments(false) as $attachment) {
+                $serviceRequestUpdate->addMediaFromStream($attachment->getStream())
+                    ->setName($attachment->getFilename())
+                    ->setFileName($attachment->getFilename())
+                    ->toMediaCollection('uploads');
```
Orrison: Reading this I am realizing one issue we could run into is if one of the attachments is NOT an accepted filetype for the collection, I am pretty sure an exception will be thrown here.

Or the it could be larger than the allowed file-size

So I think we may want to add a try/catch around this, catching any Throwable (because there could be multiple possible errors coming from this) and simply `report` the error so that we get notified in logging, but it won't cause this to fail totally.

Orrison (changes requested):

app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
```diff
@@ -324,12 +324,23 @@ protected function handleServiceRequestReply(Tenant $tenant, Parser $parser, str
...
+
+            foreach ($parser->getAttachments(false) as $attachment) {
+                try {
+                    $serviceRequestUpdate->addMediaFromStream($attachment->getStream())
+                        ->setName($attachment->getFilename())
+                        ->setFileName($attachment->getFilename())
+                        ->toMediaCollection('uploads');
+                } catch (Throwable $throw) {
+                    report($throw);
+                }
```
Orrison, line 346: This looks good, if you can please wrap the media add for Engagements as well

Orrison (changes requested):
Once the stacked branch merged, rebase on main, and resolve the last comment

Orrison (approved):

PR #1058, opened 2026-04-28
[AIDAPP-1088]: Introduce support for service request management in widget

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1088

### Technical Description

Introduces the ability to sign in and submit a service request (type, priority, title, description, and attachments only) using the existing Assistant widget. Inside or outside the portal.

No

No

---

### DIFF
--- a/app-modules/ai/routes/widgets.php
+++ b/app-modules/ai/routes/widgets.php
@@ -35,7 +35,10 @@
 */
 
 use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantBroadcastController;
+use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetAuthController;
 use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetController;
+use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetServiceRequestController;
+use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetServiceRequestSubmitController;
 use AidingApp\Ai\Http\Middleware\AssistantWidgetCors;
 use AidingApp\Ai\Http\Middleware\EnsureAssistantWidgetIsEmbeddableAndAuthorized;
 use AidingApp\Portal\Http\Middleware\EnsureKnowledgeManagementPortalIsEnabled;
@@ -65,8 +68,28 @@ Route::middleware([
                         Route::get('config', [AssistantWidgetController::class, 'config'])
                             ->name('config');
 
+                        Route::post('authenticate/request', [AssistantWidgetAuthController::class, 'request'])
+                            ->middleware(['signed:relative'])
+                            ->name('authenticate.request');
+
+                        Route::post('authenticate/{authentication}', [AssistantWidgetAuthController::class, 'authenticate'])
+                            ->middleware(['signed:relative'])
+                            ->name('authenticate');
+
                         Route::post('messages', [AssistantWidgetController::class, 'sendMessage'])
                             ->name('messages');
+
+                        Route::get('service-request-types', [AssistantWidgetServiceRequestController::class, 'index'])
+                            ->middleware(['auth:sanctum'])
+                            ->name('service-request-types');
+
+                        Route::get('service-request/upload-url', [AssistantWidgetServiceRequestSubmitController::class, 'uploadUrl'])
+                            ->middleware(['auth:sanctum'])
+                            ->name('service-request.upload-url');
+
+                        Route::post('service-request/{type}', [AssistantWidgetServiceRequestSubmitController::class, 'store'])
+                            ->middleware(['auth:sanctum'])
+                            ->name('service-request.store');
 
                         Route::match(
                             ['GET', 'POST', 'HEAD'],
--- a/app-modules/ai/routes/widgets.php
+++ b/app-modules/ai/routes/widgets.php
@@ -34,11 +34,15 @@
 </COPYRIGHT>
 */
 
-use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantBroadcastController;
-use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetAuthController;
-use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetController;
-use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetServiceRequestController;
-use AidingApp\Ai\Http\Controllers\AssistantWidget\AssistantWidgetServiceRequestSubmitController;
+use AidingApp\Ai\Http\Controllers\AssistantWidget\AuthenticateController;
+use AidingApp\Ai\Http\Controllers\AssistantWidget\AuthorizeBroadcastController;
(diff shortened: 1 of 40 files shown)


app-modules/ai/routes/widgets.php
```diff
@@ -65,9 +68,29 @@
                         Route::get('config', [AssistantWidgetController::class, 'config'])
                             ->name('config');
 
+                        Route::post('authenticate/request', [AssistantWidgetAuthController::class, 'request'])
+                            ->middleware(['signed:relative'])
```
Orrison, line 76: Is there a reason that here, and in some other routes added, this needs to be a relative signing and not an absolute?

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestController.php
```diff
@@ -0,0 +1,127 @@
...
+namespace AidingApp\Ai\Http\Controllers\AssistantWidget;
+
+use AidingApp\ServiceManagement\Models\ServiceRequestType;
+use AidingApp\ServiceManagement\Models\ServiceRequestTypeCategory;
+use App\Http\Controllers\Controller;
+use Illuminate\Http\JsonResponse;
+
+class AssistantWidgetServiceRequestController extends Controller
+{
+    public function index(): JsonResponse
```
Orrison: Typically we have created single method controllers as invokable. Should we consider sticking with that pattern here, and in some of the other controllers in this PR?

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestSubmitController.php
```diff
@@ -0,0 +1,152 @@
...
+use Illuminate\Support\Facades\Bus;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Storage;
+use Illuminate\Support\Str;
+use Symfony\Component\HttpFoundation\Response;
+use Throwable;
+
+class AssistantWidgetServiceRequestSubmitController extends Controller
+{
+    public function uploadUrl(Request $request): JsonResponse
```
Orrison: I may just not be seeing it yet, but do we have any sort of frontend / backend validation of the accepted file-types for these uploads?

There is protection on the media collection of the `ServiceRequest` model itself. But with no validation during the flow an uncaught exception would be thrown upon submission.

app-modules/ai/src/Http/Controllers/AssistantWidget/StoreServiceRequestController.php
```diff
@@ -0,0 +1,152 @@
...
+                        $serviceRequest,
+                        $file['path'],
+                        $file['original_file_name'],
+                        $uploadsMediaCollection->getName(),
+                    ),
+                    $data['attachments'] ?? [],
+                ),
+            )
+                ->name("persist-service-request-uploads-{$serviceRequest->getKey()}")
+                ->dispatchAfterResponse();
```
Orrison, line 116: What is the reasoning behind queueing the persistence of the attachments?

IMO if any of them failed to persist I would consider the entire opening of the SR to have failed as it would lack context the user intended to provide. And there would be now user or agent awareness of this.

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestSubmitController.php
```diff
@@ -0,0 +1,152 @@
...
+            'attachments' => ['nullable', 'array'],
+            'attachments.*.path' => [
+                'required_with:attachments',
+                'string',
+                'regex:/^tmp\/[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}\.[a-zA-Z0-9]+$/i',
+            ],
+            'attachments.*.original_file_name' => ['required_with:attachments', 'string', 'max:255'],
+        ]);
+
+        $priority = $type->priorities()->findOrFail($data['priority_id']);
```
Orrison: Do we want to consider moving this down into the `try/catch`? If it fails will the user get an understandable error?

widgets/assistant/src/components/ServiceRequest/DetailsStep.vue
```diff
@@ -0,0 +1,146 @@
...
+        />
+
+        <FormKit
+            v-if="rawData?.upload_url"
+            type="upload"
+            name="attachments"
+            label="Attachments"
+            :upload-url="rawData.upload_url"
+            :multiple="true"
+            :accept="['*/*']"
```
Orrison: Like I started talking about in my other comments, we may want to consider passing down the accepted mimetypes (which can be retrieved from the collection and should be so that remains central, there are helpers to do this)

Without doing so there is no frontend validation for acceptable files.

Orrison (changes requested):

app-modules/ai/src/Http/Controllers/AssistantWidget/StoreServiceRequestController.php
Orrison: Since this is an existing pattern right now we are going to leave this how it is and @danharrin is going to make a ticket to address it here and in the place we copied it from.

app-modules/ai/routes/widgets.php
Orrison: Since this is an existing pattern right now we are going to leave this how it is and @danharrin is going to make a ticket to address it here and in the place we copied it from.

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestSubmitController.php
danharrin: Sure

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestController.php
danharrin: Refactored all controllers to invokable

app-modules/ai/src/Http/Controllers/AssistantWidget/AssistantWidgetServiceRequestSubmitController.php
danharrin: Backend is on the media collection inside the upload job

widgets/assistant/src/components/ServiceRequest/DetailsStep.vue
danharrin: I have passed this from the backend to the frontend and applied it now

Orrison (changes requested):
This is all looking good but I ran into some final issues I think we should resolve.

1. I think we should also pass through the max file size and max files limit down to the frontend if possible
2. I was able to submit the SR form while an upload was still uploading, which resulted in the SR getting created without the upload. Like the other places in the app and portal we should disable the form submit while the upload field is processing.

Orrison (approved):

PR #1059, opened 2026-04-28
[AIDAPP-1097]: Add "dev mode" for compiling widget/portal assets

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1097

### Technical Description

> Added dev watch scripts to package.json so that changes to portal/widget files are automatically rebuilt without needing to manually run a build command each time
> Added `dev:portals` and `dev:widgets` scripts to the root package.json as convenience shortcuts

> No

> No

---

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,8 +7,8 @@
         "build:vite": "vite build",
         "build:widgets": "(cd widgets && npm i && npm run build)",
         "dev": "vite",
-        "dev:portals": "(cd portals && npm run dev)",
-        "dev:widgets": "(cd widgets && npm run dev)",
+        "dev:portals": "(cd portals && npm i && npm run dev)",
+        "dev:widgets": "(cd widgets && npm i && npm run dev)",
         "postinstall": "node npm-postinstall.cjs"
     },
     "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -7,6 +7,8 @@
         "build:vite": "vite build",
         "build:widgets": "(cd widgets && npm i && npm run build)",
         "dev": "vite",
+        "dev:portals": "(cd portals && npm run dev)",
+        "dev:widgets": "(cd widgets && npm run dev)",
         "postinstall": "node npm-postinstall.cjs"
     },
     "devDependencies": {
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -7,6 +7,8 @@
         "build:vite": "vite build",
         "build:widgets": "(cd widgets && npm i && npm run build)",
         "dev": "vite",
+        "dev:portals": "(cd portals && npm run dev)",
```
danharrin: I think we probably want to `npm i` here like we do for the build commands, in case the user hasn't done it before. Should be fast if the current lock file hash is the same as what is installed.

portals/package.json
```diff
@@ -5,6 +5,8 @@
     "scripts": {
         "build": "npm run build:knowledge-management",
         "build:knowledge-management": "(cd knowledge-management && vite build)",
+        "dev": "npm run dev:knowledge-management",
+        "dev:knowledge-management": "(cd knowledge-management && vite build --watch)",
```
danharrin, line 9: I ran this command, but it doesn't look like HMR works. Do you know how to get that working?

danharrin (changes requested):

portals/package.json
nency-canyon: > I ran this command, but it doesn't look like HMR works. Do you know how to get that working?

I ran into the same issue while implementing this `vite build --watch` gives automatic incremental rebuilds on file save, but it doesn't provide true HMR. The reason is that the portal uses `defineCustomElement` which renders inside a Shadow DOM. Hard refresh is still needed to see changes after each rebuild.

To get the portal working, I had to:
1. Run pls down && pls build --no-cache && pls up -d 
2. Remove `node_modules` and `package-lock.json` from root, Portals and Widgets
3. Run npm i from the root, then run the build command

But I'm not sure If this is the reason I could get this working. Can you please try this and let me know if this works for you or not?

package.json
nency-canyon: Agreed, added `npm i` && to both dev:portals and dev:widgets in the root package.json to match the pattern used by the build commands. Can you please verify this and let me know if now this works.

danharrin (approved):

PR #1064, opened 2026-04-29
[AIDAPP-1098] : Some users have reported Advisory severity not displaying properly

monali-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1098

### Technical Description

> Fix color handling in IncidentSeverity model and update RGB color retrieval logic

> No

> No

---


app-modules/service-management/src/Models/IncidentSeverity.php
```diff
@@ -80,7 +80,7 @@ public function incidents(): HasMany
     protected function rgbColor(): Attribute
     {
         return new Attribute(
-            get: fn () => 'rgb(' . Color::all()[$this->color][600] . ')',
+            get: fn () => $this->color ? Color::from($this->color)->getRgb() : null,
```
payal-canyon: Add the cast - `'color' => Color::class` on `IncidentSeverity`. Matches the convention in `ServiceRequestStatus` and `ContactType`. The accessor then collapses to `$this->color?->getRgb()`.

Update the factory - with the cast in place, the factory can pass enum cases directly: `'color' => $this->faker->randomElement(Color::cases())`.

Nit / optional: In the model `protected $appends = ['rgb_color']` likely isn't needed anymore - the Filament pages access `rgb_color` as a direct attribute, not via JSON/array serialization. Safe to remove as long as nothing else serializes the model. Once it’s removed, it would be helpful to run a quick UI test to ensure everything is working as expected.

payal-canyon (changes requested):

app-modules/service-management/src/Models/IncidentSeverity.php
monali-canyon: `$this->color?->getRgb()` is returning PHPStan error because `getRgb() `is a method on the `Color` enum. The issue is that `$this->color` is a string. That's why I used `instanceof`

payal-canyon: @monali-canyon , After updating this line to `$this->color->getRgb()`, could you regenerate the helper files? The IDE helper is currently declaring $color as string because it hasn't been regenerated since the cast was added - that's why PHPStan is flagging the call. Once regenerated it should resolve to Color and the lint error will go away. Let me know how it goes!

payal-canyon (approved):

PR #1069, opened 2026-05-01
[AIDAPP-1116]: Update the knowledge base article list to display helpfulness rating

monali-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1116

### Technical Description

- Add rating column to Knowledge Base Items list with helpful vote percentage

> No

> No

---


app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
```diff
@@ -86,6 +86,20 @@ public function table(Table $table): Table
...
+                        $totalVotes = $record->votes()->count();
+
+                        if ($totalVotes === 0) {
+                            return 'Unrated';
+                        }
+
+                        $helpfulVotes = $record->votes()->where('is_helpful', true)->count();
+
+                        return (int) round(($helpfulVotes / $totalVotes) * 100) . '%';
+                    }),
```
payal-canyon, line 104: The `getStateUsing` callback currently runs once per row and fires two `COUNT(*)` queries each time - one for the total number of votes and another for the helpful ones - so a 25-row page issues up to 50 extra queries on top of the main listing query. We can avoid this by eager-loading both counts on the base table query using `modifyQueryUsing` together with `withCount`, where the second count is scoped with a closure to only include rows where `is_helpful` is `true` and aliased as `helpful_votes_count`. Once those counts are preloaded, the column itself becomes a cheap in-memory accessor that just reads `$record->votes_count` and `$record->helpful_votes_count`, returning `Unrated` when the total is zero and otherwise the rounded percentage. The end result is the same output as today, but the whole page resolves in a single SQL pass instead of scaling linearly with the number of rows.

payal-canyon (changes requested):

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
```diff
@@ -86,6 +90,22 @@ public function table(Table $table): Table
                         );
                     })
                     ->sortable(),
+                TextColumn::make('rating')
+                    ->label('Rating')
+                    ->toggleable()
+                    ->getStateUsing(function (KnowledgeBaseItem $record): string {
+                        /** @var int $totalVotes */
```
payal-canyon: Could you please clarify why this doc block was added? Are you encountering any linting issues because of it? It seems unnecessary at the moment.

If possible, could you try removing this doc block and regenerating the IDE helper file to see if everything still works as expected? If it does, we can consider removing this one and any similar doc blocks in the same area.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
```diff
@@ -86,6 +90,22 @@ public function table(Table $table): Table
...
+                        $totalVotes = $record->votes_count;
+
+                        if ($totalVotes === 0) {
+                            return 'Unrated';
+                        }
+
+                        /** @var int $helpfulVotes */
+                        $helpfulVotes = $record->getAttribute('helpful_votes_count');
```
payal-canyon: Using `getAttribute` works fine, but I’d suggest directly using `$record->helpful_votes_count` in the `return` statement. This would allow us to remove the extra variable, as we can reference it directly instead of using the current `$helpfulVotes`.

payal-canyon (changes requested):

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
monali-canyon: I have modified as per your suggestions

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
monali-canyon: Without `getAttribute` it return PHPStan error:
Access to an undefined property `AidingApp\KnowledgeBase\Models\KnowledgeBaseItem::$helpful_votes_count`.

payal-canyon (approved):

PR #1070, opened 2026-05-01
[AIDAPP-1125]: Some users have reported that menu items are available in the portal despite authorization being needed

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1125

### Technical Description

> Fixed Service, Status, and Advisories menu items displaying in the knowledge management portal navigation for unauthenticated users by replacing static boolean expressions with reactive computed() refs tied to the authenticated user state

> No

> No

---


portals/knowledge-management/src/Components/Header.vue
Orrison, line 124: It's strange that these 3 don't seem to have the same access issues. (I wonder if I didn't have assets enabled when I first loaded a page) Regardless, I think we should do the same  user check on all 3 of these as well.

Orrison (changes requested):

portals/knowledge-management/src/Components/Header.vue
nency-canyon: I agree, I have added explicit checks on all three to make the auth requirement intentional at the UI layer rather than relying on server-side data being absent. Please verify and let me know if you find any issue. Thanks!

Orrison (approved):

PR #1083, opened 2026-05-06
[AIDAPP-1118]: Enhance the knowledge base by introducing article health and deprecating article quality

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1118

### Technical Description

> Remove Knowledge base quality from application.
> Create a new Column "Heath" in Knowledge Base Item page that will check Article Title, Details, Manager, Broken Links and Broken Images. And if all this components are fine that will give green icon showing article is healthy or red for otherwise.
> All this details will be shown in tab inside view of knowledge base article.
> Jobs are created to check  Broken Links and Broken Images added in article_details column.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_05_12_broken_links_feature.md

---


app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/KnowledgeBaseItemResource.php
payal-canyon, line 82: The quality relationship was removed from `KnowledgeBaseItem`. Because of that, typing anything into the global search currently throws a `RelationNotFoundException`.

Please remove the `quality` reference from the global search.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ViewKnowledgeBaseItem.php
```diff
@@ -133,6 +133,51 @@ public function infolist(Schema $schema): Schema
...
+                                        ->where('status', '!=', ConcernStatus::Archived)
+                                        ->isEmpty()),
+                                IconEntry::make('no_broken_links')
+                                    ->label('No Broken Links Detected')
+                                    ->boolean()
+                                    ->getStateUsing(fn (KnowledgeBaseItem $record) => ! $record->are_broken_links_detected),
+                                IconEntry::make('no_broken_images')
+                                    ->label('No Broken Images Detected')
+                                    ->boolean()
+                                    ->getStateUsing(fn (KnowledgeBaseItem $record) => ! $record->are_broken_images_detected),
```
payal-canyon: The broken link/image data is already being stored in the `broken_links` and `broken_images` columns, but it is not currently surfaced in the UI.

Per the requirement  `Store which links are broken and show it in a tooltip` (same for broken images) - we should attach a `->tooltip(...)` to the `no_broken_links` and `no_broken_images` `IconEntry` components in `ViewKnowledgeBaseItem.php` so the failing URLs are visible to users.

app-modules/knowledge-base/database/migrations/2026_05_05_164358_add_broken_links_and_images_column_in_knowledge_base_articles.php
```diff
@@ -34,36 +34,33 @@
...
-                    ->columns(),
-            ]);
+        DB::transaction(function () {
+            Schema::table('knowledge_base_articles', function (Blueprint $table) {
+                $table->boolean('are_broken_links_detected')->default(false);
+                $table->jsonb('broken_links')->nullable();
+                $table->boolean('are_broken_images_detected')->default(false);
+                $table->jsonb('broken_images')->nullable();
+            });
+        });
```
payal-canyon: Since this only performs a single operation, we do not need to wrap it in a DB transaction.
Also, we should add a feature flag so that `Health` is only displayed when the feature flag is active.
Please follow the latest feature flag documentation here: `docs/how-tos/manage-feature-flags.md`

app-modules/knowledge-base/src/Observers/KnowledgeBaseItemObserver.php
```diff
@@ -50,10 +52,13 @@ public function saving(KnowledgeBaseItem $knowledgeBaseItem): void
...
 
-    public function saved(): void
+    public function saved(KnowledgeBaseItem $knowledgeBaseItem): void
     {
         if (app(AiSupportAssistantSettings::class)->is_enabled && app(PortalSettings::class)->ai_support_assistant) {
             PrepareKnowledgeBaseVectorStore::dispatch();
         }
+
+        CheckKnowledgeBaseArticleLinksJob::dispatch($knowledgeBaseItem);
+        CheckKnowledgeBaseArticleImagesJob::dispatch($knowledgeBaseItem);
```
payal-canyon: I think we can wrap these jobs inside a `wasChanged()` check instead:
`if ($knowledgeBaseItem->wasChanged('article_details'))`
This way, the jobs will only run when `article_details` has actually changed.

payal-canyon (commented):
`RecentKnowledgeBaseArticlesList` does not appear to be used anywhere in the codebase anymore, and it is also not visible anywhere in the UI. It looks like completely dead code at this point, so let’s remove the entire file.

payal-canyon (changes requested):
@jayushi-canyon, I uploaded a broken video URL, but the check is still marked as true in the Health tab. Could you please take a look?
The same issue seems to happen for images as well. Is it working correctly on your side?
Also, it looks like the feature flag was missed. Could you please add that as well?
Thanks!

payal-canyon (dismissed):

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
```diff
@@ -106,10 +108,35 @@ public function table(Table $table): Table
...
+                        $hasNoUnresolvedConcerns = $record->concerns
+                            ->where('status', '!=', ConcernStatus::Resolved)
+                            ->where('status', '!=', ConcernStatus::Archived)
+                            ->isEmpty();
+
+                        if (! BrokenLinksFeature::active()) {
+                            return $hasTitle && $hasArticle && $hasManager && $hasNoUnresolvedConcerns;
+                        }
+                        $hasNoBrokenLinks = ! $record->are_broken_links_detected;
+                        $hasNoBrokenImages = ! $record->are_broken_images_detected;
```
Orrison: Since we do these checks in multiple places (here, on the view page, etc.) What do you think about turning them in Attributes on the model so we centralize the logic?

app-modules/knowledge-base/src/Jobs/CheckKnowledgeBaseArticleImagesJob.php
```diff
@@ -0,0 +1,243 @@
...
+
+    public function uniqueId(): string
+    {
+        return $this->knowledgeBaseItem->getKey();
+    }
+
+    public function uniqueFor(): int
+    {
+        return 180;
+    }
```
Orrison, line 68: I would make this a little longer here

```suggestion
    public function uniqueFor(): int
    {
        return 600;
    }
```

app-modules/knowledge-base/src/Jobs/CheckKnowledgeBaseArticleLinksJob.php
```diff
@@ -0,0 +1,206 @@
...
+    public function __construct(public KnowledgeBaseItem $knowledgeBaseItem) {}
+
+    public function uniqueId(): string
+    {
+        return $this->knowledgeBaseItem->getKey();
+    }
+
+    public function uniqueFor(): int
+    {
+        return 180;
```
Orrison:
```suggestion
        return 600;
```

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItemsTest.php
Orrison, line 1: Let's add tests for the health column ensuring the state with all combinations of issues and non issues.

Let's also make tests for the Entries in the view page as well too

Orrison (changes requested):
Some feedback and requested tests.

Let's also make tests for both check jobs as well.

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
jayushi-canyon: Yes, doing that we won't have to take care of logic on different places, but only take care of just on single method.

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItemsTest.php
```diff
@@ -421,3 +378,237 @@
...
+        ->create();
+
+    KnowledgeBaseItemConcern::factory()
+        ->for($knowledgeBaseItem, 'knowledgeBaseItem')
+        ->state(['status' => ConcernStatus::New])
+        ->create();
+
+    livewire(ListKnowledgeBaseItems::class)
+        ->assertTableColumnStateSet('health', false, $knowledgeBaseItem);
+});
```
Orrison, line 616: I think we need tests here regarding the health of the image / link scans.

Since this is just testing the rendering of whether it is saying true or false you don't have to do any mocking / creating of the article with these issues. Just set the columns to true or false imo

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/ViewKnowledgeBaseItemTest.php
```diff
@@ -190,3 +192,122 @@
...
+    livewire(ViewKnowledgeBaseItem::class, ['record' => $knowledgeBaseItem->getRouteKey()])
+        ->assertSchemaComponentStateSet('title_filled', true, 'infolist');
+});
+
+test('Health tab article_filled shows correct state', function () {
+    asSuperAdmin();
+
+    $knowledgeBaseItem = KnowledgeBaseItem::factory()
+        ->state(['article_details' => null])
+        ->create();
```
Orrison, line 219: I think having this check when it is `null` is fine here.

But if you can please save an article with no content locally and also ADD a step to this test checking that it returns false in that scenario.

I believe it is like a json object with an empty array in it kind of thing. If you can also test that on the list page test too.

Again, just what to say, let's ADD that check here to the checks already done with this test so we have good coverage of it being literally `null`, the json blob without any content in it, and then filled.

app-modules/knowledge-base/src/Jobs/CheckKnowledgeBaseArticleLinksJob.php
Orrison, line 1: Please add tests for this job

app-modules/knowledge-base/src/Jobs/CheckKnowledgeBaseArticleImagesJob.php
Orrison, line 1: Please add tests for this job

Orrison (changes requested):

Orrison (approved):

PR #1086, opened 2026-05-07
[AIDAPP-1123]: Enhance the AI support assistant by making the assistant more inquisitive during engagement

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1123

### Technical Description

> Enhance the AI Support Assistant to be more inquisitive and on-topic, and expose the system prompt as a configurable setting under Global Administration > Artificial Intelligence > AI Support Assistant

> No

### Cleanup Tasks

> Yes - it's documented in `.cleanup-tasks/2026_05_07_ai_support_assistant_default_instructions_feature.md`

---

### DIFF
--- a/.cleanup-tasks/2026_05_07_ai_support_assistant_default_instructions_feature.md
+++ b/.cleanup-tasks/2026_05_07_ai_support_assistant_default_instructions_feature.md
@@ -9,8 +9,10 @@ created: 2026-05-07
 
 ## Temporary Migrations
 
+- app-modules/ai/database/migrations/2026_05_13_120000_tmp_backfill_ai_support_assistant_default_instructions.php
+
 ## Additional Cleanup
 
 - In `app-modules/ai/src/Jobs/PortalAssistant/SendMessage.php`, remove the `AiSupportAssistantDefaultInstructionsFeature::active()` branch and keep only the active path that reads from `AiSupportAssistantSettings::instructions` (with the `defaultInstructions()` fallback).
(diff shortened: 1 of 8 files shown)


danharrin (changes requested):
My first concern is that we will need to maintain the "default" prompt" text in two places: the PHP class for resetting the value, and the migration.

The problem is that we can't just reference the PHP class inside the migration because that would break our rule about migrations not referencing app code.

I think we should have a `tmp_` migration that fills the prompt into the settings to start with, after the migration that creates the empty settings property. This can reference the PHP class that contains the default prompt since it is temporary.

Then, we should introduce a new seeder that is run for new tenants that sets the default prompt. This can reference app code since it is a permanent class, and we avoid the duplication.

What do you think?

payal-canyon:
> My first concern is that we will need to maintain the "default" prompt" text in two places: the PHP class for resetting the value, and the migration.
> 
> The problem is that we can't just reference the PHP class inside the migration because that would break our rule about migrations not referencing app code.
> 
> I think we should have a `tmp_` migration that fills the prompt into the settings to start with, after the migration that creates the empty settings property. This can reference the PHP class that contains the default prompt since it is temporary.
> 
> Then, we should introduce a new seeder that is run for new tenants that sets the default prompt. This can reference app code since it is a permanent class, and we avoid the duplication.
> 
> What do you think?

Yeah, you were right! I’ve added the changes as suggested. Thanks!

danharrin (approved):

PR #1087, opened 2026-05-07
[AIDAPP-1128]: Introduce consistent portal button component design and options based Filament admin panel buttons

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1128

### Technical Description

> Created a shared BaseButton Vue component as per Filament's button supporting `tag`, `color`, `size`, `icon`, `icon-position`, `loading`, and `disabled` props.
> Migrated all portal usages from the old portal-scoped ui/BaseButton to the new shared component, remapping `variant` → `color`, `as` → `tag`, and `:icon-left` → `:icon` across all pages and components in portals

> No

### Cleanup Tasks

> No

---


resources/js/components/BaseButton.vue
```diff
@@ -0,0 +1,288 @@
...
+        'border-radius': 'var(--rounding-md, 0.5rem)',
+    }));
+
+    const iconClasses = computed(() => [
+        'shrink-0 transition duration-75',
+        props.color === 'gray' && 'text-gray-400',
+        ['xs', 'sm'].includes(props.size) ? 'h-4 w-4' : 'h-5 w-5',
+    ]);
+
+    const resolvedIcon = computed(() => {
```
danharrin: I think we can simplify this.

In Vue, I think you can just pass a component into a prop without needing to find it by name.

So the consumer of the button should import the correct Heroicon, and then pass that component reference directly to the icon prop of the component.

Then we don't need to import all icons in this component and search for the correct one, which should improve rendering performance.

resources/js/components/BaseButton.vue
```diff
@@ -0,0 +1,288 @@
...
+<template>
+    <component
+        :is="componentTag"
+        v-bind="componentAttrs"
+        :class="buttonClasses"
+        :style="buttonStyle"
+        @click="handleClick"
+    >
+        <template v-if="iconPosition === 'before' && (loading || resolvedIcon)">
+            <ArrowPathIcon v-if="loading" :class="[...iconClasses, 'animate-spin']" aria-hidden="true" />
```
danharrin: Please copy the loading indicator from Filament into its own Vue component which you reference here, instead of `ArrowPathIcon`. https://github.com/filamentphp/filament/blob/e722cc2ce0789ca4a0ecfc27907fd910cbaf0e70/packages/support/src/View/DefaultLoadingIndicator.php#L13-L30

resources/js/components/BaseButton.vue
```diff
@@ -0,0 +1,288 @@
...
+-->
+<script setup>
+    import * as HeroiconsMini from '@heroicons/vue/20/solid';
+    import * as HeroiconsOutline from '@heroicons/vue/24/outline';
+    import * as HeroiconsSolid from '@heroicons/vue/24/solid';
+    import { ArrowPathIcon } from '@heroicons/vue/24/solid';
+    import { computed, useAttrs, useSlots } from 'vue';
+    import { RouterLink } from 'vue-router';
+
+    const COLOR_VARS = {
```
danharrin, line 38: Can you explain the color approach here please? We should be referencing the colors that are already defined / default values in the Tailwind config instead of redefining them here.

danharrin (changes requested):
Please merge the latest changes from `main`, and then refactor the portal assistant widget to use the new Button component too in all appropriate places.

nency-canyon:
> Please merge the latest changes from `main`, and then refactor the portal assistant widget to use the new Button component too in all appropriate places.

The original task scoped widgets as a future concern, so I have implemented in portals only. Let me also use this in widgets also.

danharrin:
@nency-canyon just to check, you are just processing the portal assistant widget and not the other widgets that are used outside the portal?

resources/js/components/BaseButton.vue
nency-canyon: Done. The icon prop now accepts [Object, Function] only, no more namespace imports or name-based lookup inside BaseButton. Each consumer imports the specific Heroicon they need and passes it directly.

resources/js/components/BaseButton.vue
nency-canyon: All non-primary colors now reference Tailwind v4 CSS variables directly. Initially I tried to match Tailwind's default palette. But I have now corrected the implementation.

nency-canyon:
> @nency-canyon just to check, you are just processing the portal assistant widget and not the other widgets that are used outside the portal?

Yes, @danharrin I have communicated with the Kevin, later I have only updated portal assistant.

widgets/assistant/src/components/ChatSignIn.vue
danharrin, line 1: It would be nice to be able to customise the loading text so that the buttons keep their custom loading text like before

resources/js/components/BaseButton.vue
danharrin: Please extract it into its own Vue component to avoid repetition

danharrin (changes requested):

widgets/assistant/src/components/ChatSignIn.vue
danharrin: There were other files that also used custom loading text before, please ensure those are preserved

nency-canyon: Added Loading text support. I have also added dedicated vue component for Loading state.
Please review and let me know if you find any issues. Thanks!

danharrin (changes requested):
Can you go through all the button usages and ensure that they use the `solid` Heroicon variants please, the 16x16 size should be used

nency-canyon:
> Can you go through all the button usages and ensure that they use the `solid` Heroicon variants please, the 16x16 size should be used

Done. All BaseButton :icon usages have been updated to use the solid 16×16 Heroicon variant, Please verify and let me know if you find any further issues.

danharrin (dismissed):

danharrin (approved):

PR #1098, opened 2026-05-11
[AIDAPP-1139]: Introduce the concept of presence of users

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1139

### Technical Description

Uses a Livewire persistent middleware for all requests, ignoring requests without calls or updates (pure component reloads) and requests that only contain chart widget update calls (triggered by polling while viewing a chart widget). Uses a new timestamp prop on the user to track when they were last active, and using the activity settings, converts that into an active status which is displayed in the user list and view page as an accessible icon or badge.

No

### Cleanup Tasks

.cleanup-tasks/user-presence.md

---

### DIFF
--- a/app/Filament/Pages/ManagePresenceSettings.php
+++ b/app/Filament/Pages/ManagePresenceSettings.php
@@ -17,7 +17,7 @@
       in the software, and you may not remove or obscure any functionality in the
       software that is protected by the license key.
     - You may not alter, remove, or obscure any licensing, copyright, or other notices
-      of the licensor in the software. Any use of the licensor's trademarks is subject
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
       to applicable law.
     - Canyon GBS Inc. respects the intellectual property rights of others and expects the
       same in return. Canyon GBS® and Aiding App® are registered trademarks of
--- a/app/Filament/Pages/ManagePresenceSettings.php
+++ b/app/Filament/Pages/ManagePresenceSettings.php
@@ -1,0 +1,120 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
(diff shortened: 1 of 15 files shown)


app/Http/Middleware/TrackPresence.php
```diff
@@ -0,0 +1,115 @@
...
+
+        if (! UserPresence::active()) {
+            return $response;
+        }
+
+        if ($this->isBackgroundRequest($request)) {
+            return $response;
+        }
+
+        $user->touchQuietly('last_activity_at');
```
payal-canyon: `touch($attribute)` sets the named attribute and calls `save()`, which still runs `updateTimestamps()` and updates `updated_at`. So, every authenticated user-initiated request ends up updating both `last_activity_at` and `updated_at` on the user row.

I think this is a real issue because the `Updated At` column is displayed in `ListUsers.php`, so it effectively starts behaving like a `Last Activity` column as well. Anywhere else relying on `updated_at` (caching, sync diffs, audits, etc.) could also be unintentionally affected.

So, I think we should address this. WDYT?

app/Filament/Pages/ManagePresenceSettings.php
```diff
@@ -0,0 +1,120 @@
...
+                    ->minValue(1)
+                    ->required()
+                    ->suffix('minutes'),
+                TextInput::make('inactive_threshold')
+                    ->label('Inactive threshold')
+                    ->helperText('Minutes of inactivity before a user is considered offline.')
+                    ->numeric()
+                    ->minValue(1)
+                    ->required()
+                    ->suffix('minutes'),
```
payal-canyon, line 101: The `presenceStatus()` cascade in `User.php` assumes the thresholds follow this order:
`active_threshold < idle_threshold < inactive_threshold`
But currently, the form only validates `numeric` and `minValue(1)`, so it’s possible to save values that break this logic and make certain presence states unreachable.

For example, if someone submits `active=30` and `idle=5`, the Idle state would never be returned because the range becomes invalid. Similarly, an incorrect inactive threshold could cause users to skip directly to Offline.

Since this feature is intended to support assignment/escalation based on availability, I think we should enforce the threshold ordering through validation - probably using Laravel’s built-in `gt:` / `lt:` rules.
WDYT?

payal-canyon (changes requested):

payal-canyon (approved):

PR #1102, opened 2026-05-12
[AIDAPP-1074]: Update the name of team to department to better reflect its departmental and organization chart structure

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1074

### Technical Description

> Rename Team to Department across the application.Updated all references from Team to Department in models, factories, tests, and migrations.Adjusted database migrations to rename tables and columns from teams to departments.Modified permission names and foreign key constraints to reflect the new Department structure.Removed the old Team module and its composer configuration.

> No

### Cleanup Tasks

> Yes - it's documented in `.cleanup-tasks/2026_05_15_team_rename_feature.md`

---


danharrin (changes requested):
Here are some places that I think are missing feature flag checks: `DepartmentPolicy` (permission names), `ProjectPolicy` & `ServiceRequestResource` & `CreateServiceRequest` & `ListServiceRequests` & `ManagedServiceRequestType` (`departments.id`), the morph map keys (for example audit logs for a team could be read before the migration has been run but would be empty?)

I think we want all new migrations combined so that they share a transaction and all are reverted if something goes wrong, otherwise there is time between migrations when the DB has changed but the feature flag is not active, so it leaves the deployment in a broken state should they fail. Alternative would be to use a feature flag per migration, but I think it's easier to combine them.

You might want to verify with @Orrison if any external systems or clients might use the data from `AllUsersController` via the API, and are expecting `team` and not `department`.

I think you might also want to merge `main` in to ensure that there aren't any new usages that would be broken or need migrating.

Orrison:
> Here are some places that I think are missing feature flag checks: `DepartmentPolicy` (permission names), `ProjectPolicy` & `ServiceRequestResource` & `CreateServiceRequest` & `ListServiceRequests` & `ManagedServiceRequestType` (`departments.id`), the morph map keys (for example audit logs for a team could be read before the migration has been run but would be empty?)
> 
> I think we want all new migrations combined so that they share a transaction and all are reverted if something goes wrong, otherwise there is time between migrations when the DB has changed but the feature flag is not active, so it leaves the deployment in a broken state should they fail. Alternative would be to use a feature flag per migration, but I think it's easier to combine them.
> 
> You might want to verify with @Orrison if any external systems or clients might use the data from `AllUsersController` via the API, and are expecting `team` and not `department`.
> 
> I think you might also want to merge `main` in to ensure that there aren't any new usages that would be broken or need migrating.

There is nothing currently consuming or expecting `team` as confirmed by @payal-canyon reviewing our other services calling this endpoint.
And no other client is consuming this endpoint.

.cleanup-tasks/2026_05_15_team_rename_feature.md
```diff
@@ -0,0 +1,14 @@
...
+created: 2026-05-15
+---
+
+## Feature Flags
+
+- App\Features\TeamRenameFeature
+
+## Temporary Migrations
+
+- database/migrations/2026_05_14_095810_activate_team_rename_feature.php
```
Orrison: I think the migration file had it's name changed?

Orrison (dismissed):
One comment on the cleanup file not having the filename updated.

Responded to @danharrin's question about consumers here. We should be all good.

@danharrin I did not review everything in this PR as I assume you did already. If you want me to conduct a full review please DM me and let me know. Other-wise if you think this is good then I am good with it.

.cleanup-tasks/2026_05_15_team_rename_feature.md
payal-canyon: Yeah! Actually, I had created separate migration and feature activation files earlier, but based on Dan’s suggestion, I updated the implementation and forgot to remove that line from the document.

Now, the feature flag is being activated directly within the migration file itself.

danharrin (changes requested):
Please resolve the merge conflict

danharrin (approved):

PR #1111, opened 2026-05-15
[AIDAPP-1115]: Enhance the user experience related to files associated with a service request

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1115

### Technical Description

- Implemented file upload tracking for service requests by adding created_by_id and created_by_type polymorphic columns to the media table via a new migration
- Created a Media observer that automatically stamps the authenticated user as the creator when uploading files
- Built a reusable `ServiceRequestMediaTable` Filament table widget showing file name, uploader info, date, and download action

> No

### Cleanup Tasks

> Feature Flag Added: `MediaCreatedByFeature`

---

### DIFF
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -54,7 +54,6 @@ use AidingApp\ServiceManagement\Models\ServiceRequest;
 use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
 use AidingApp\ServiceManagement\Models\TenantServiceRequestTypeDomain;
 use App\Features\MediaCreatedByFeature;
-use App\Models\Media;
 use App\Models\Tenant;
 use Aws\Crypto\KmsMaterialsProviderV3;
 use Aws\Kms\KmsClient;
@@ -339,7 +338,6 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
 
             foreach ($parser->getAttachments(false) as $attachment) {
                 try {
-                    /** @var Media $media */
                     $media = $serviceRequestUpdate->addMediaFromStream($attachment->getStream())
                         ->setName($attachment->getFilename())
                         ->setFileName($attachment->getFilename())
@@ -526,7 +524,6 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
 
                 foreach ($parser->getAttachments(false) as $attachment) {
                     try {
-                        /** @var Media $media */
                         $media = $serviceRequest->addMediaFromStream($attachment->getStream())
                             ->setName($attachment->getFilename())
                             ->setFileName($attachment->getFilename())
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -54,6 +54,7 @@ use AidingApp\ServiceManagement\Models\ServiceRequest;
 use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
 use AidingApp\ServiceManagement\Models\TenantServiceRequestTypeDomain;
 use App\Features\MediaCreatedByFeature;
+use App\Models\Media;
 use App\Models\Tenant;
 use Aws\Crypto\KmsMaterialsProviderV3;
 use Aws\Kms\KmsClient;
@@ -338,7 +339,7 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
 
             foreach ($parser->getAttachments(false) as $attachment) {
                 try {
-                    /** @var \App\Models\Media $media */
+                    /** @var Media $media */
                     $media = $serviceRequestUpdate->addMediaFromStream($attachment->getStream())
                         ->setName($attachment->getFilename())
                         ->setFileName($attachment->getFilename())
@@ -525,7 +526,7 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
 
                 foreach ($parser->getAttachments(false) as $attachment) {
                     try {
-                        /** @var \App\Models\Media $media */
+                        /** @var Media $media */
                         $media = $serviceRequest->addMediaFromStream($attachment->getStream())
                             ->setName($attachment->getFilename())
                             ->setFileName($attachment->getFilename())
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -344,6 +344,7 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
                         ->toMediaCollection('uploads');
 
                     $respondent = $serviceRequest->respondent;
+
                     if (MediaCreatedByFeature::active() && is_null($media->created_by_id)) {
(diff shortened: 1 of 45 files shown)


database/migrations/2026_05_12_254102_add_created_by_to_media_table.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        DB::transaction(function () {
+            Schema::table('media', function (Blueprint $table) {
+                if (! Schema::hasColumn('media', 'created_by_type')) {
+                    $table->string('created_by_type')->nullable();
+                }
+
+                if (! Schema::hasColumn('media', 'created_by_id')) {
+                    $table->uuid('created_by_id')->nullable()->index();
+                }
```
Orrison: This comment is both for here, and in the `down` method.

One should NEVER have conditionals like this in a migration, migrations should always be expected to run as they are written, as they should be the only time in which the database is modified.

If something was going on where this would fail when it ran, we would want to know and adjust accordingly.

So I would get rid of any `Schema::hasColumn` checks.

database/migrations/2026_05_12_254102_add_created_by_to_media_table.php
```diff
@@ -0,0 +1,77 @@
...
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('media', function (Blueprint $table) {
+                if (! Schema::hasColumn('media', 'created_by_type')) {
+                    $table->string('created_by_type')->nullable();
+                }
+
+                if (! Schema::hasColumn('media', 'created_by_id')) {
+                    $table->uuid('created_by_id')->nullable()->index();
```
Orrison: I would add an index for `created_by_type` and `created_by_id`, I would look at either using the morph method here or look at what it does to generate the columns and proper index/s.

In fact, I think there is a uuid morph helper that could be best to use here.

app/Observers/MediaObserver.php
```diff
@@ -0,0 +1,58 @@
...
+        if (! MediaCreatedByFeature::active() || filled($media->created_by_id)) {
+            return;
+        }
+
+        $user = Auth::user();
+
+        if ($user && ($key = $user->getKey()) !== null) {
+            $media->created_by_id = (string) $key;
+            $media->created_by_type = $user->getMorphClass();
+        }
```
Orrison: Just to cover all bases I would call the `$user` variable `$creator` as technically not all logged in entities this could be are `User`s.

Additionally, instead of setting these directly via the attributes. I would recommend using the `createdBy` relationship directly so we don't have to keep these column names updated in a bunch of places if it ever changes.

In almost ALL cases, using the relationship you established in the model is better than interacting directly with the columns in code.

app/Models/Media.php
```diff
@@ -0,0 +1,105 @@
...
+
+#[ObservedBy([MediaObserver::class])]
+class Media extends SpatieMedia
+{
+    /**
+     * @return MorphTo<Model, Media>
+     */
+    public function createdBy(): MorphTo
+    {
+        /** @var MorphTo<Model, Media> */
```
Orrison: I am not sure this `@var` comment is needed

app/Models/Media.php
```diff
@@ -0,0 +1,105 @@
...
+
+        // Handle User (has 'name')
+        if (isset($creator->name)) {
+            return (string) $creator->name;
+        }
+
+        // Handle Contact (has 'first_name', 'last_name')
+        $fullName = trim(trim($creator->first_name ?? '') . ' ' . trim($creator->last_name ?? ''));
+
+        return $fullName ?: 'N/A';
```
Orrison: I would maybe do a match statement here checking what `$creator` is an instance of (User or Contact).

Defaulting to `N/A`

And then that gives you type completion and understanding inside the nested functions of the match statement for what entity type it is.

app/Models/Media.php
```diff
@@ -0,0 +1,105 @@
...
+            $orgName = (string) ($creator->getAttribute('organization')?->getAttribute('name') ?? '');
+
+            if ($typeName && $orgName) {
+                return "{$typeName} - {$orgName}";
+            }
+
+            return $typeName ?: ($orgName ?: null);
+        }
+
+        return null;
```
Orrison: Same match statement comment as my other one.

app/Models/Media.php
```diff
@@ -0,0 +1,105 @@
...
+
+use App\Features\MediaCreatedByFeature;
+use App\Observers\MediaObserver;
+use Illuminate\Database\Eloquent\Attributes\ObservedBy;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\MorphTo;
+use Spatie\MediaLibrary\MediaCollections\Models\Media as SpatieMedia;
+
+#[ObservedBy([MediaObserver::class])]
+class Media extends SpatieMedia
```
Orrison, line 51: If you run all migrations and then `composer generate-helper-files` I am pretty sure this will generate the correct helper docblock needed for this.

app-modules/portal/src/Jobs/PersistServiceRequestUpload.php
```diff
@@ -83,10 +84,18 @@ public function handle(): void
...
                 ->usingName(pathinfo($this->originalFileName, PATHINFO_FILENAME))
                 ->toMediaCollection($this->collection);
+
+            $respondent = $this->serviceRequest->respondent;
+
+            if (MediaCreatedByFeature::active() && is_null($media->created_by_id)) {
+                $media->created_by_id = $respondent->getKey();
+                $media->created_by_type = $respondent->getMorphClass();
+                $media->saveQuietly();
+            }
```
Orrison, line 97: I would use the relationship here directory if possible, instead of the columns

app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
```diff
@@ -516,10 +525,16 @@ protected function handleServiceRequestType(TenantServiceRequestTypeDomain $serv
...
+                        $media = $serviceRequest->addMediaFromStream($attachment->getStream())
                             ->setName($attachment->getFilename())
                             ->setFileName($attachment->getFilename())
                             ->toMediaCollection('uploads');
+
+                        if (MediaCreatedByFeature::active() && is_null($media->created_by_id)) {
+                            $media->created_by_id = $contact->getKey();
+                            $media->created_by_type = $contact->getMorphClass();
+                            $media->saveQuietly();
+                        }
```
Orrison, line 538: I would use the relationship here directly if possible

app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
```diff
@@ -337,10 +338,18 @@ protected function handleServiceRequestReply(Tenant $tenant, Parser $parser, str
...
                         ->setFileName($attachment->getFilename())
                         ->toMediaCollection('uploads');
+
+                    $respondent = $serviceRequest->respondent;
+
+                    if (MediaCreatedByFeature::active() && is_null($media->created_by_id)) {
+                        $media->created_by_id = $respondent->getKey();
+                        $media->created_by_type = $respondent->getMorphClass();
+                        $media->saveQuietly();
+                    }
```
Orrison, line 354: I would use the relationship here directly if possible

Orrison (changes requested):

Orrison:
@monali4-canyon I have left some feedback. There is also a test that fails repeatedly in this PR. Though I am unsure if it totally related to your changes, I am not seeing other PRs fail this and it doesn't seem flaky as it fails every time I try and re-run it here

Orrison (dismissed):

Orrison (dismissed):

Orrison (approved):

PR #1135, opened 2026-05-21
[AIDAPP-1173]: Improve badge components in the portal for consistency

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1173

### Technical Description

> Introduces a new shared BaseBadge Vue component modeled on Filament's badge component.

> No

### Cleanup Tasks

> N/A

---


danharrin (changes requested):
I think we should rename the `hero` color to something like `transparent-white` as there are some places it is not used in the hero, like the chat header

I think the "Advisory History" page has badges on it that should be refactored too

Also, I think the "mono" option isn't being used currently and I don't think its in Filament, is there a specific reason you introduced this?

nency-canyon:
> I think we should rename the `hero` color to something like `transparent-white` as there are some places it is not used in the hero, like the chat header
> 
> I think the "Advisory History" page has badges on it that should be refactored too
> 
> Also, I think the "mono" option isn't being used currently and I don't think its in Filament, is there a specific reason you introduced this?

Thanks for the feedback! I have addressed all the feedbacks. 
I have removed mono prop usage entirely and has updated to use color="gray" instead, which achieves the same result without a custom prop outside of Filament's API. Let me know if you find any further issues, Thanks!

danharrin (approved):

PR #1139, opened 2026-05-22
[AIDAPP-1172]: Update Service Request Update portal form inputs for consistency with other inputs in the portal

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1172

### Technical Description

> Replaces the custom HTML form used to submit service request updates on the portal's "View Service Request" page with FormKit components.

> No

### Cleanup Tasks

> N/A

---


danharrin (changes requested):
Please rebase this so that it only includes the changes related to this ticket and not also the badge refactor. As far as I know, this form does not contain badges, so it can be merged without the other ticket blocking it. I can't really accurately review this change on its own before the badge one is merged otherwise.

nency-canyon:
> Please rebase this so that it only includes the changes related to this ticket and not also the badge refactor. As far as I know, this form does not contain badges, so it can be merged without the other ticket blocking it. I can't really accurately review this change on its own before the badge one is merged otherwise.

I have reverted changes. I have accidentally created branch from different branch.

portals/knowledge-management/src/Pages/ViewServiceRequest.vue
```diff
@@ -317,86 +253,26 @@
...
+                <FormKit
+                    type="textarea"
+                    name="description"
+                    label="Update"
+                    placeholder="Enter your update here..."
+                    validation="required"
+                    validation-visibility="submit"
+                    :classes="{ outer: 'mb-4', inner: 'max-w-full!', input: 'w-full h-32' }"
+                />
+                <FormKit
```
danharrin, line 266: I think we have a Filepond-powered upload field available, can we use that please?

danharrin: It is the same one we use for creating a service request currently

danharrin (changes requested):

nency-canyon (commented):
Allow me sometime.

portals/knowledge-management/src/Pages/ViewServiceRequest.vue
nency-canyon: They are different inputs. Let me re-verify and Update you.

nency-canyon: @danharrin There are two upload inputs available.

This PR uses the native FormKit file input because it submits standard multipart form data, which matches the current backend implementation (`$request->hasFile('files')`). No backend changes are needed.

The FilePond upload used in the "Create Service Request" flow uploads files directly to S3 using presigned URLs and submits only file paths. The current update flow does not support this and would require backend changes plus a new upload URL endpoint.

Let me know your thoughts on this, Thanks!

danharrin (approved):

PR #1140, opened 2026-05-24
[AIDAPP-1178]: Improve the internal handling of navigation labels

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1178

### Technical Description

- refactor navigation groups to use enum values

No

### Cleanup Tasks

No

---

### DIFF
--- a/app/Providers/Filament/AdminPanelProvider.php
+++ b/app/Providers/Filament/AdminPanelProvider.php
@@ -51,11 +51,11 @@ use Filament\Actions\ExportAction;
 use Filament\Actions\ImportAction;
 use Filament\Forms\Components\Field;
 use Filament\Http\Middleware\Authenticate;
+use App\Enums\NavigationGroup;
 use Filament\Http\Middleware\DisableBladeIconComponents;
 use Filament\Http\Middleware\DispatchServingFilamentEvent;
 use Filament\Infolists\Components\Entry;
 use Filament\Navigation\MenuItem;
-use Filament\Navigation\NavigationGroup;
 use Filament\Panel;
 use Filament\PanelProvider;
 use Filament\Support\Enums\Width;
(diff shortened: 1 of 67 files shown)


app/Providers/Filament/AdminPanelProvider.php
```diff
@@ -144,52 +143,6 @@ public function panel(Panel $panel): Panel
...
-                    ->collapsed(),
-                NavigationGroup::make()
-                    ->label('Settings')
-                    ->icon('heroicon-o-adjustments-vertical')
-                    ->collapsed(),
-                NavigationGroup::make()
-                    ->label('Global Admin')
-                    ->icon('heroicon-o-shield-check')
-                    ->collapsed(),
-            ])
```
payal-canyon, line 184: Looks like we're losing the "groups collapsed by default" behavior with this change. The old `->navigationGroups([...])` block was calling `->collapsed()` on each group, but now that we're going enum-based without registering the enum on the panel, Filament builds the groups through its discovery path which only reads `HasLabel` and `HasIcon` from the enum - it never checks `Collapsible`. So on a fresh browser every group renders expanded. Existing browsers still look fine because the old collapse state is cached in localStorage from before, which is why this is easy to miss locally.

Let's add `->navigationGroups(NavigationGroup::class)` to the panel config in this file. That routes the groups through `NavigationGroup::fromEnum()`, which respects the `Collapsible` interface, so we keep the previous default behavior.

payal-canyon (changes requested):

payal-canyon (dismissed):
LGTM overall. @danharrin could you give the `ProductHealth` changes a second look? We've stopped extending `HealthCheckResults` and reimplemented its internals - works, but want to make sure that's the right call vs. keeping the extension and routing nav config through the plugin setters. Just want a sanity check before merge.

danharrin:
@monali4-canyon why is it that, like Payal mentioned, we need to reimplement `ProductHealth`? The parent class even has the correct return type on the nav group method so I am a little confused as to what caused that: https://github.com/shuvroroy/filament-spatie-laravel-health/blob/0f506a53f464db62a9f6c43ec1fd5c199b8e2fde/src/Pages/HealthCheckResults.php#L39-L42

If you could help me understand why we need to copy it instead of extending from the parent that would be helpful so that I can either approve or request that it is changed back.

monali4-canyon:
@danharrin 

The third-party `HealthCheckResults` class defines `getNavigationGroup(): ?string` — it can only return a string. But Filament's enum-based navigation groups require the method to return the enum case itself (not its string label) so items are grouped correctly by serialization identity.

When one item returns the enum `(NavigationGroup::GlobalAdmin)` and another returns the string `('Global Admin')`, Filament serializes them differently and creates duplicate navigation groups in the sidebar.

Since I can't change the vendor class's return type, I extend Page directly and bring over the minimal functionality. 

I tried many ways to fix it but it was not working as expected. If you have any feasible solution please help.

danharrin:
@monali4-canyon are we using the latest version of the package then? If you look at the link I sent the return type appears correct

monali4-canyon:
@danharrin 
The vendor class `HealthCheckResults::getNavigationGroup()` has a `?string` return type, but in other files we use `$navigationGroup = NavigationGroup::GlobalAdmin` which returns an `enum`. Due to this conflict, it was duplicating the Global Admin navigation group.

danharrin:
@monali4-canyon I think you should update the package and check again?

payal-canyon (changes requested):
@monali4-canyon , Once Dan's suggestion regarding the package update has been addressed, could you please also review the suggestions I left on the Advising App PR?

Since both PRs touch the same area, I believe those suggestions may apply here as well. If applicable, could you please make the same updates in this PR too?

After that, please re-request my review. Thanks!

monali4-canyon:
@payal-canyon Sure will do. Since It is medium priority ticket and I do have High priority tickets in my bucket so will check priority wise.

monali4-canyon:
@payal-canyon You have added suggestion to remove `protected static string | BackedEnum | null $navigationIcon = '';` .
So for this PR I have added for those files where Dashboard is extended and it overrides `getNavigationIcon` method from dashboard.

It generates test case error:
Navigation group [Service Desk] has an icon but one or more of its items also have icons. Either the group or its items can have icons, but not both. This is to ensure a proper user experience.

Other fixes applied

payal-canyon (approved):

PR #1145, opened 2026-05-26
[AIDAPP-1190]: API Endpoint to List and Search Service Requests

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1190

### Technical Description

> Add API endpoint to fetch service request list and allow users to filter and sort.

> No

### Cleanup Tasks

> No

---


danharrin (changes requested):
Please implement tests for this endpoint similar to `app-modules/student-data-model/tests/Tenant/Http/Controllers/Api/V1/Students/ListStudentsControllerTest.php`

danharrin (changes requested):
Please have a look at that example test I sent, I think it's better to use datasets to test filters and sorts etc so that the test file is smaller and it's easier to add new ones. What do you think?

danharrin (dismissed):

danharrin (approved):

PR #1146, opened 2026-05-26
[AIDAPP-1195]: API Endpoint to View User Profile Details 

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1195

### Technical Description

> API Endpoint to View User Profile Details

> No

### Cleanup Tasks

> N/A

---


payal-canyon (dismissed):
This PR currently has merge conflicts. Could you please resolve them and then re-request my review? Thanks!

danharrin (changes requested):
Please take a look at https://github.com/canyongbs/aidingapp/pull/1145 and copy how the controller namespaces, API documentation through Scramble, and route registration works. https://github.com/canyongbs/aidingapp/pull/1147 is another example that might be closer to what you are doing since it is a "view" instead of "list" endpoint.

danharrin (changes requested):
Please check again for the route registration technique we use. Let me know if there is a reason we can't make this PR consistent

mittal-canyon:
> Please check again for the route registration technique we use. Let me know if there is a reason we can't make this PR consistent

@danharrin I have applied these changes because the user has no module, and so added route in root api.php. Now I have added some changes to match the exact route definition, like the existing one by adding api_v1.php

danharrin (approved):

PR #1155, opened 2026-05-28
[AIDAPP-1201]: Some users have reported that in emails, header images are not rendering at their intended size

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1201

### Technical Description

Ensure header logos are not restricted to 75px width by replicating styles used in other branches of header image determination

No

### Cleanup Tasks

N/A

---


payal-canyon (changes requested):
Looks good overall. Could we remove `.nvm` file? It doesn't seem to be related to the changes in this PR, and I think it may have been added accidentally.

payal-canyon (approved):

PR #1157, opened 2026-05-29
[AIDAPP-1166]: Some users have reported that duplicating a knowledge base article results in an error

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1166

### Technical Description

> Some users have reported that duplicating a knowledge base article results in an error

> NO

### Cleanup Tasks

> N/A

---


app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItemsTest.php
```diff
@@ -664,3 +665,60 @@
     livewire(ListKnowledgeBaseItems::class)
         ->assertTableColumnStateSet('health', false, $knowledgeBaseItem);
 });
+
+test('an authorised user can duplicate a knowledge base article', function () {
```
danharrin, line 670: Please add assertions to ensure that the newly created item has the correct properties and relationships in each case

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItemsTest.php
```diff
@@ -664,3 +665,60 @@
...
+    $user = User::factory()->create();
+    $user->givePermissionTo('knowledge_base_item.view-any');
+
+    actingAs($user);
+
+    livewire(ListKnowledgeBaseItems::class)
+        ->assertTableActionHidden('replicate', $knowledgeBaseItem->getKey());
+});
+
+test('duplicating a knowledge base article is gated with proper feature access control', function () {
```
danharrin: I don't think we need this test as it is not related to the duplication logic

app-modules/knowledge-base/src/Filament/Resources/KnowledgeBaseItems/Pages/ListKnowledgeBaseItems.php
```diff
@@ -174,7 +175,9 @@ public function table(Table $table): Table
             ], layout: FiltersLayout::BeforeContent)
             ->recordActions([
                 ReplicateAction::make()
+                    ->authorize(fn (): bool => auth()->user()->can('knowledge_base_item.create'))
```
danharrin: This is a direct permission check, we should instead use the `create` method on the policy

danharrin (changes requested):

danharrin (changes requested):
I tested this and it appears that tags attached while duplicating are not saved. Please fix this and ensure that all fields in the modal have test coverage to ensure the correct data is saved.

danharrin (approved):

PR #1162, opened 2026-05-29
[AIDAPP-1215]: Status cards should look consistent with the knowledge base category cards

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1215

### Technical Description

> Restyled the service monitor status cards to be visually consistent with the knowledge base category cards. 

> No

### Cleanup Tasks

> N/A

---


danharrin (changes requested):
Since the cards aren't clickable, I don't think we should have the BG color change on hover

Please use solid Heroicon components with the correct size for the icons in these cards

nency-canyon:
> Since the cards aren't clickable, I don't think we should have the BG color change on hover
> 
> Please use solid Heroicon components with the correct size for the icons in these cards

Addressed this changes, please review and let me know if you find any issues.

danharrin (approved):

PR #1163, opened 2026-05-29
[AIDAPP-1208]: Refactor the concept of custom forms

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1208

### Technical Description

> Refactor Custom Forms into Service Request Types.
> Forms are now managed inside a Service Request Type via a new Custom Form page instead of as standalone, embeddable entities. Editing a form with submissions archives it and creates a new active version (old service requests still render their original fields); only one active form exists per type. Removes the standalone form pages, embedding, and reCAPTCHA.

> No

### Cleanup Tasks

> N/A

---


app-modules/form/src/Actions/GenerateSubmissibleEmbedCode.php
Orrison, line 1: If we are deleting this, can we delete the actual widget code / api endpoints for the widget as well?

Or is any of that being used by the Form / Service Request flow in Filament, Portal, or in the Support Assistant Widget?

app-modules/service-management/database/factories/ServiceRequestFormFactory.php
Orrison, line 1: Was the RenderServiceRequestForm used to generate the preview? Because they still wanted to retain the ability to see the preview

app-modules/service-management/database/migrations/2026_05_29_200207_remove_embed_and_recaptcha_from_service_request_forms.php
```diff
@@ -34,34 +34,41 @@
...
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('service_request_forms', function (Blueprint $table) {
+                $table->dropColumn([
+                    'embed_enabled',
+                    'allowed_domains',
+                    'recaptcha_enabled',
+                    'primary_color',
+                    'rounding',
```
Orrison, line 54: It's possible that we could be asked to remove them later. But the current request does not mention removing any of the color / rounding settings, so let's keep that how it is.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypeCustomForm.php
```diff
@@ -0,0 +1,321 @@
...
+
+    public function getRelationManagers(): array
+    {
+        // Needed to prevent Filament from loading the relation managers on this page.
+        return [];
+    }
+
+    public function form(Schema $schema): Schema
+    {
+        return $schema
```
Orrison, line 89: We still need the description, color, and rounding settings retained. Anything not crossed out in the ticket should have been retained here.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypeCustomForm.php
```diff
@@ -0,0 +1,321 @@
...
+            $form = new ServiceRequestForm();
+            $form->type()->associate($type);
+        }
+
+        // Only new forms (a brand new type form, or a new version replacing an archived
+        // one) need a name. An in-place edit keeps the existing form's name. Each new form
+        // gets a distinct name so it never collides with a prior (archived) version.
+        if (! $form->exists) {
+            $form->fill(['name' => $this->resolveFormName($type)]);
+        }
```
Orrison: Honestly I am wondering if Service Request forms even need a name at this point or not. It made sense that they did in the past so you could identify them in the list page. But now I am not sure.

Do we even output the name anywhere after these changes?

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ViewServiceRequestType.php
```diff
@@ -75,7 +74,7 @@ public function infolist(Schema $schema): Schema
                         TextEntry::make('form.name')
                             ->label('Form')
                             ->hidden(fn (ServiceRequestType $record) => ! $record->form)
-                            ->url(fn (ServiceRequestType $record) => $record->form ? ServiceRequestFormResource::getUrl('edit', ['record' => $record->form]) : null)
+                            ->url(fn (ServiceRequestType $record) => ManageServiceRequestTypeCustomForm::getUrl(['record' => $record]))
                             ->color('primary'),
```
Orrison: Based on what I was talking about above, perhaps this can be turned into a bool field that shows a green check mark or a red x depending on whether or not the SR Type HAS a form or not. With the column linking to the form page like you have it now.

app-modules/service-management/src/Http/Controllers/ServiceRequestFormWidgetController.php
```diff
@@ -103,249 +72,24 @@ function () use ($stream) {
...
                 'is_authenticated' => false,
-                'recaptcha_enabled' => false,
                 'schema' => $generateSchema($serviceRequestForm),
-                'primary_color' => collect(Color::all()[$serviceRequestForm->primary_color ?? 'blue'])
+                'primary_color' => collect(Color::all()[$portalSettings->knowledge_management_portal_primary_color->value ?? 'blue'])
                     ->map(Color::convertToRgb(...))
                     ->map(fn (string $value): string => (string) str($value)->after('rgb(')->before(')'))
                     ->all(),
-                'rounding' => $serviceRequestForm->rounding,
+                'rounding' => $portalSettings->knowledge_management_portal_rounding,
```
Orrison: Interesting. I think it makes sense to have it pull from the KB settings like you are doing here. I think we just want to make sure Product approves before we do so since that wasn't originally asked for in the ticket.

Orrison (changes requested):
A few questions, though I think I had some of them answered as I reviewed more. It looks like we did retain the preview ability. So I just want to make sure we have everything properly cleaned up.

app-modules/form/src/Actions/GenerateSubmissibleEmbedCode.php
payal-canyon: I didn't remove it because the widget is still being used by the Preview functionality. Other than that, nothing in the Portal, Support Assistant Widget, or general Form flow depends on it, as each of those renders forms through its own implementation.

To fully remove the widget, we would first need to re-implement the Preview functionality. Since the purpose of the preview is to reflect what end users actually see-and those experiences are rendered using FormKit.

Given that, I would suggest one of the following options:
- Leave it as-is for now.
- Refactor the Preview as part of this PR and then remove the widget.
- Create a separate ticket to re-implement the FormKit-based Preview without relying on this specific widget bundle.

Any of these approaches would work. Please let me know which option you would prefer.

app-modules/service-management/database/factories/ServiceRequestFormFactory.php
payal-canyon: `RenderServiceRequestForm` wasn't the preview, so the preview is still intact.

`RenderServiceRequestForm` was the public "respond" page `(/service-request-forms/{id}/respond)`. Its blade rendered the form via `GenerateSubmissibleEmbedCode` - i.e. the real embeddable form for standalone public submission. That's the standalone form rendering we're intentionally removing.

Preview is a separate path and was kept:
Route `service-request-forms.preview` (+ preview-entry)
`ServiceRequestFormPreviewController` -> `service-request-form-preview.blade.php` -> renders the widget with `preview="true"`
The Custom Form page's Preview action points to it
So the only thing deleted was the standalone public-submission page, not the preview. The preview still works and is covered by the existing preview tests (`ServiceRequestFormPreviewControllerTest`, `ServiceRequestFormWidgetPreviewTest`), which assert it renders the `service-request-form-embed` element in `preview="true"` mode.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypeCustomForm.php
payal-canyon: after the refactor the form name is output in exactly two places, both admin-only:

Preview heading - `ServiceRequestFormWidgetController::preview` returns name (we strip the (2) suffix) -> shown as the preview's title.
The type's View page - `ViewServiceRequestType` has `TextEntry::make('form.name')->label('Form')`, a link to the Custom Form page that displays the form's name (e.g. "Password Reset Form (2)").
It is never shown to end users. So, could you please let me know if what we've done is correct, or if any changes are needed?

app-modules/form/src/Actions/GenerateSubmissibleEmbedCode.php
Orrison: Let's just leave it as it is for now. We can refactor later.

app-modules/service-management/database/migrations/2026_05_29_200207_remove_embed_and_recaptcha_from_service_request_forms.php
Orrison: Sorry, we can keep this removal.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypeCustomForm.php
Orrison: Yeah I think, if you agree, we just don't need the name any more. Any place we reference the name now based on what you told me, can just be a bool checkmark / X as to whether or not the type HAS a form.

Since the form is tied so close to the type now it has no need of an independent name.

And in the Preview we can just NOT display it

Orrison (changes requested):
Apologies @payal-canyon, like I sent in my message earlier today, let's actually move forward with removing the color and rounding settings from forms and having them inherit them from the portal settings / their own fallbacks if the portal settings are not set / configured.

I also responded back to a few of my earlier comments with additional requests.

Orrison (changes requested):
@payal-canyon I am still seeing the name column existing here, it looks like just the unique index was dropped. Not the actual column

Orrison (approved):

PR #1166, opened 2026-06-01
[AIDAPP-1199]: Add ability to manually define the portal display order of knowledge base article categories

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1199

### Technical Description

- Implement sorting feature for knowledge base categories and update related components

No

### Cleanup Tasks

.cleanup-tasks/2026_06_01_knowledge_base_category_sort_feature.md

---


app-modules/knowledge-base/database/migrations/2026_05_29_251423_add_sort_column_to_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,109 @@
...
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('knowledge_base_categories', function (Blueprint $table) {
+                $table->integer('sort')->nullable();
+                $table->index('sort');
+            });
+
+            // TODO: Cleanup - Remove this backfill block and change the column above to non-nullable with default 0
```
Orrison: You should add details of this to `.cleanup-tasks/2026_06_01_knowledge_base_category_sort_feature.md`

Really all that is needed is a reference in the 2nd or 3rd section of the cleanup file that this migration has cleanup that is needed. Then once a developer or AI does the cleanup they will know what to do from your comments here.

app-modules/knowledge-base/database/migrations/2026_05_29_251423_add_sort_column_to_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,109 @@
...
+                ->whereNull('parent_id')
+                ->whereNull('deleted_at')
+                ->orderBy('name')
+                ->get();
+
+            foreach ($parentCategories as $index => $category) {
+                DB::table('knowledge_base_categories')
+                    ->where('id', $category->id)
+                    ->update(['sort' => $index + 1]);
+            }
```
Orrison: Calling a `get()` like this inline and doing a foreach over it is not memory safe. If there are too many records you could crash the process by overloading memory.

Instead of `get` you could use `each` to do what you are doing with the `foreach` loop here which will automatically chunk for you.

Unless maybe that won't give you an index? Though I think it might

app-modules/knowledge-base/database/migrations/2026_05_29_251423_add_sort_column_to_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,109 @@
...
+                    ->whereNull('deleted_at')
+                    ->orderBy('name')
+                    ->get();
+
+                foreach ($subCategories as $index => $category) {
+                    DB::table('knowledge_base_categories')
+                        ->where('id', $category->id)
+                        ->update(['sort' => $index + 1]);
+                }
+            }
```
Orrison: Same as above there must be a way to better handle this without having to load all the parent IDs and records into memory.

@payal-canyon is really good at this kind of thing if you want some help.

app-modules/knowledge-base/database/migrations/2026_05_29_251423_add_sort_column_to_knowledge_base_categories_table.php
```diff
@@ -0,0 +1,109 @@
...
+            Schema::table('knowledge_base_categories', function (Blueprint $table) {
+                $table->integer('sort')->nullable();
+                $table->index('sort');
+            });
+
+            // TODO: Cleanup - Remove this backfill block and change the column above to non-nullable with default 0
+            // Backfill sort for parent categories (those without a parent_id)
+            $parentCategories = DB::table('knowledge_base_categories')
+                ->whereNull('parent_id')
+                ->whereNull('deleted_at')
```
Orrison: In many of your tests you account for soft deleted record. But in this migration here and below, we ignore them.

Which would result in existing soft deleting records I think having the "default" of `0`.

Is that intended and do you foresee any issues with that?

Orrison (changes requested):

app-modules/knowledge-base/database/migrations/2026_05_29_251423_add_sort_column_to_knowledge_base_categories_table.php
monali4-canyon: Not any specific issue. I just removed the ->whereNull('deleted_at') filters from the migration so all records (including soft-deleted) get proper sort values assigned.

Orrison (approved):

PR #1167, opened 2026-06-01
[AIDAPP-1212]: Use pre-signed upload URLs for service request update attachments submitted in portal

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1212

### Technical Description

> Use pre-signed upload URLs for service request update attachments submitted in portal

> No

### Cleanup Tasks

> No

---


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
```diff
@@ -82,4 +81,29 @@ public function __invoke(StoreServiceRequestUpdateRequest $request): JsonRespons
 
         return response()->json(['serviceRequestUpdates' => $serviceRequestUpdates], 201);
     }
+
+    /**
+     * @param Collection<string, mixed> $data
+     */
+    protected function dispatchFileUploads(
```
Orrison: Let's actually NOT dispatch the saving into a different job/s and do it inline the the request.

It was actually always a not intended that we do it in a different job and eventually I will make a ticket to remove that from Service Requests.

We just need to wrap all this controller in a DB transaction and roll back on errors so the update doesn't get saved etc. if a failure happens in the media processing. Which is preferred.

Orrison (changes requested):

Orrison (approved):

PR #1169, opened 2026-06-02
[AIDAPP-1211]: Enhance service request notifications

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1211

### Technical Description

>  Introduced ServiceRequestNotificationChannel enum to manage notification channels (Email, App).
> Enhanced ServiceRequestEmailTemplateType enum with methods for event slugs and view labels.
> Updated the notifications matrix view to utilize enums for roles and channels.
> Implemented ServiceRequestTypeEmailPreference model to manage email preferences in a pivot table.
> Refactored ServiceRequestType model to include email preferences relationship and preference checking logic.
> Updated observers to use the new preference checking method instead of legacy boolean columns.

> No

### Cleanup Tasks

> .cleanup-tasks/2026_06_02_data_migrate_service_request_type_email_preferences.md

---


.cleanup-tasks/2026_06_02_data_migrate_service_request_type_email_preferences.md
Orrison, line 1: Is there a particular reason for there to be two cleanup tasks?

You can / should define all of what needs to be cleaned up in the one single cleanup task. Or at least that was my intention with them

Orrison (dismissed):
I had one bit of feedback on the cleanup tasks. But I would like @danharrin to complete the review on this one as it was his original idea to split these and I ran out of time to get through this one today.

danharrin (changes requested):
I might be wrong, but I don't see anywhere where you are using the new "assigned manager" preferences to send a notification to that manager. Did I miss something?

jayushi-canyon:
> I might be wrong, but I don't see anywhere where you are using the new "assigned manager" preferences to send a notification to that manager. Did I miss something?

@danharrin No you are not missing anything, I assumed that for this ticket iteration we are only supposed to add the "Assigned manager" in Notifications page only. I have added the functionality now.

danharrin (approved):

PR #1170, opened 2026-06-02
[AIDAPP-1192]: API Endpoint to Update Service Request Details

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1192

### Technical Description

> API Endpoint to Update Service Request Details

> No

### Cleanup Tasks

> Yes, Feature Flag: `ServiceRequestAssignmentByTypeFeature`

---


app-modules/service-management/database/migrations/2026_06_04_105849_add_assigned_by_type_and_update_assigned_by_id_to_service_request_assignments_table.php
```diff
@@ -0,0 +1,84 @@
...
+use Illuminate\Support\Facades\DB;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('service_request_assignments', function (Blueprint $table) {
+                $table->string('assigned_by_type')->nullable()->after('assigned_by_id');
```
Orrison: `after` is not something available to use with Postgres

```suggestion
                $table->string('assigned_by_type')->nullable();
```

app-modules/service-management/database/migrations/2026_06_04_105849_add_assigned_by_type_and_update_assigned_by_id_to_service_request_assignments_table.php
```diff
@@ -0,0 +1,84 @@
...
+            });
+
+            DB::table('service_request_assignments')
+                ->whereNotNull('assigned_by_id')
+                ->update(['assigned_by_type' => (new User())->getMorphClass()]);
```
Orrison, line 55: I think we can add a comment here / note in the cleanup task to remove this after we deploy

Orrison (changes requested):

app-modules/service-management/src/Actions/UpdateServiceRequestAction.php
```diff
@@ -0,0 +1,69 @@
...
+                'assigned_by_id' => auth()->id(),
+                'assigned_at' => now(),
+                'status' => ServiceRequestAssignmentStatus::Active,
+            ];
+
+            if (ServiceRequestAssignmentByTypeFeature::active()) {
+                $assignmentData['assigned_by_type'] = auth()->user() ? auth()->user()->getMorphClass() : null;
+            }
+            $serviceRequest->assignments()->create($assignmentData);
+        }
```
Orrison, line 65: Do we need to do anything to deactivate an existing assignments here?

app-modules/service-management/src/Actions/UpdateServiceRequest.php
```diff
@@ -0,0 +1,69 @@
...
+        if ($dataArray['assigned_to_id'] ?? null) {
+            $assignmentData = [
+                'user_id' => $dataArray['assigned_to_id'],
+                'assigned_by_id' => auth()->id(),
+                'assigned_at' => now(),
+                'status' => ServiceRequestAssignmentStatus::Active,
+            ];
+
+            if (ServiceRequestAssignmentByTypeFeature::active()) {
+                $assignmentData['assigned_by_type'] = auth()->user() ? auth()->user()->getMorphClass() : null;
```
Orrison: Would this maybe work better?

```suggestion
                $assignmentData['assigned_by_type'] =  auth()->user()?->getMorphClass();
```

app-modules/service-management/src/DataTransferObjects/UpdateServiceRequestData.php
```diff
@@ -0,0 +1,70 @@
...
+use Spatie\LaravelData\Optional;
+
+class UpdateServiceRequestData extends Data
+{
+    public function __construct(
+        public string | Optional $status_id, // @phpstan-ignore MeliorStan.parameterNameNotCamelCase
+        public string | Optional $priority_id, // @phpstan-ignore MeliorStan.parameterNameNotCamelCase
+        public string | Optional $close_details, // @phpstan-ignore MeliorStan.parameterNameNotCamelCase
+        public string | Optional $assigned_to_id, // @phpstan-ignore MeliorStan.parameterNameNotCamelCase
+        public ServiceRequestCategory | Optional $category, // @phpstan-ignore MeliorStan.parameterNameNotCamelCase
```
Orrison: You actually should be able to make these camel case and I believe there are ways to map them to different keys in the package: https://spatie.be/docs/laravel-data/v4/as-a-resource/mapping-property-names

Also `$category` doesn't need the ignore anyways.

app-modules/service-management/src/Filament/Actions/ReclassifyServiceRequestAction.php
```diff
@@ -166,12 +167,18 @@ protected function setUp(): void
...
                             'user_id' => $data['assign_to'],
                             'assigned_by_id' => auth()->id(),
                             'assigned_at' => now(),
                             'status' => ServiceRequestAssignmentStatus::Active,
-                        ]);
+                        ];
+
+                        if (ServiceRequestAssignmentByTypeFeature::active()) {
+                            $assignmentData['assigned_by_type'] = auth()->user()?->getMorphClass() ?? null;
+                        }
```
Orrison, line 179: Same comments here as the other place we do this, see if there is anywhere else it would apply as well

app-modules/service-management/src/Http/Controllers/Api/V1/ServiceRequests/UpdateServiceRequestController.php
Orrison, line 1: Where is ther a different file / class that looks a lot like this one: app-modules/service-management/src/Http/Controllers/Api/V1/ServiceRequests/UpdateServiceRequestsController.php

app-modules/service-management/src/Http/Controllers/Api/V1/ServiceRequests/UpdateServiceRequestController.php
```diff
@@ -0,0 +1,78 @@
...
+    #[Group('Service Requests')]
+    public function __invoke(Request $request, UpdateServiceRequest $updateServiceRequest, ServiceRequest $serviceRequest): JsonResource
+    {
+        Gate::authorize('viewAny', ServiceRequest::class);
+        Gate::authorize('update', $serviceRequest);
+
+        $data = $request->validate([
+            'status_id' => ['nullable', 'uuid:4', Rule::exists(ServiceRequestStatus::class, 'id')],
+            'priority_id' => ['nullable', 'uuid:4', Rule::exists(ServiceRequestPriority::class, 'id')],
+            'assigned_to_id' => ['nullable', 'uuid:4', Rule::exists(User::class, 'id')],
```
Orrison, line 69: If you look at the logic for selecting a user to assign, the select input filters down to only certain people (people that are managers of this type of the service request and maybe some other things)

We need to make sure that is enforced via the API call as well so they couldn't assign someone we wouldn't expect.

It may actually make sense to do the check and throw a validation error from `UpdateServiceRequest` that way it gets checked anywhere someone attempts to update

Orrison (dismissed):

app-modules/service-management/src/Actions/UpdateServiceRequestAction.php
amit-canyon: No needed here as we already deactivated existing in `ServiceRequestAssignmentObserver`.

danharrin (dismissed):

danharrin (approved):

PR #1171, opened 2026-06-03
[AIDAPP-1202]: Enhance the service request list view by adding a filter for service request types

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1202

### Technical Description

- add filtering by service request type and implement type tree options

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/ListServiceRequests.php
+++ b/app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/ListServiceRequests.php
@@ -169,7 +169,8 @@ class ListServiceRequests extends ListRecords
                             fn (Builder $query) => $query->whereIn('type_id', $data['types'])
                         )
                     ))
-                    ->indicateUsing(fn (array $data): ?string => empty($data['types'])
+                    ->indicateUsing(
+                        fn (array $data): ?string => empty($data['types'])
                         ? null
                         : 'Types: ' . ServiceRequestType::whereIn('id', $data['types'])->pluck('name')->implode(', ')
                     ),
@@ -278,13 +279,6 @@ class ListServiceRequests extends ListRecords
             ->poll('60s');
     }
 
-    protected function getHeaderActions(): array
-    {
-        return [
-            CreateAction::make(),
-        ];
-    }
-
     /**
      * @return array<int, array{name: string, value: string, disabled: bool, children: array<int, mixed>}>
      */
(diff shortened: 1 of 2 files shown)


app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/ListServiceRequests.php
```diff
@@ -147,6 +151,29 @@ public function table(Table $table): Table
...
+                Filter::make('type')
+                    ->label('Type')
+                    ->form([
+                        SelectTree::make('types')
+                            ->label('Type')
+                            ->getTreeUsing(fn () => static::buildTypeTreeOptions())
+                            ->multiple()
+                            ->placeholder('All')
+                            ->independent(true)
+                            ->expandSelected(false),
```
Orrison: When I tested locally, the "It is not auto expand Area even though Type from that Area is selected (when you open after applying filter)" concern you raised didn't happen when I set this to `true`.

Why was this set to `false`?

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/ListServiceRequests.php
```diff
@@ -147,6 +151,29 @@ public function table(Table $table): Table
                     ->getOptionLabelFromRecordUsing(fn (ServiceRequestPriority $record) => "{$record->type->name} - {$record->name}")
                     ->multiple()
                     ->preload(),
+                Filter::make('type')
+                    ->label('Type')
+                    ->form([
```
Orrison: `form()` is deprecated and I believe should be `schema()` now.

Orrison (changes requested):
From a users perspective it looks to them like the uncategorized types are "above" the categorized area ones:

Image

But then your output here buts the Area (categories) groupings first:

Image

We may want to try and make sure we are matching visually what is in the settings. So putting uncategorized first. Then the area / categories ones.

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/ListServiceRequests.php
monali4-canyon: I set to true as well but somehow didn't work. Maybe a caching issue but now I set it again and it is working

Orrison (approved):

PR #1173, opened 2026-06-03
[AIDAPP-1210]: Enhance email creation by leveraging AI to create all the email templates at once

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1210

### Technical Description

Adds settings for example templates and an AI prompt to generate email templates with. Allows those templates to be generated in bulk from the service request type resource.

No

### Cleanup Tasks

`.cleanup-tasks/2026_06_03_service_request_notification_automation.md`

---


app-modules/service-management/database/migrations/2026_06_03_091401_create_service_request_notification_automation_email_templates_table.php
```diff
@@ -0,0 +1,80 @@
...
+return new class () extends SettingsMigration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::create('service_request_notification_automation_email_templates', function (Blueprint $table) {
+                $table->uuid('id')->primary();
+                $table->string('type');
+                $table->string('role');
+                $table->json('subject')->nullable();
+                $table->json('body')->nullable();
```
Orrison: Shouldn't these be `jsonb`?

app-modules/service-management/src/Models/ServiceRequestNotificationAutomationEmailTemplate.php
```diff
@@ -0,0 +1,96 @@
...
+use Filament\Forms\Components\RichEditor\Models\Concerns\InteractsWithRichContent;
+use Filament\Forms\Components\RichEditor\Models\Contracts\HasRichContent;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+/**
+ * @property ServiceRequestEmailTemplateType $type
+ * @property ServiceRequestTypeEmailTemplateRole $role
+ * @property array<string, mixed>|null $subject
+ * @property array<string, mixed>|null $body
+ * @property string|null $ai_instructions
```
Orrison: Is all of this needed as overrides?

Just curious if perhaps these were set before the DB was properly setup and `IdeHelperServiceRequestNotificationAutomationEmailTemplate` was generated.

If they are needed, that's okay, just wanting to have us confirm so we are not defining things we don't need to / may become out of date.

app-modules/service-management/src/Settings/ServiceRequestNotificationAutomationSettings.php
```diff
@@ -0,0 +1,111 @@
...
+                    ['type' => 'listItem', 'content' => [['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'Keep the subject concise (under 100 characters of visible text)']]]]],
+                    ['type' => 'listItem', 'content' => [['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'The body should be informative but not overly long']]]]],
+                ]],
+                ['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'Output format:']]],
+                ['type' => 'bulletList', 'content' => [
+                    ['type' => 'listItem', 'content' => [['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'Respond with only a JSON object containing "subject" and "body" keys']]]]],
+                    ['type' => 'listItem', 'content' => [['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'Both values must be valid HTML']]]]],
+                    ['type' => 'listItem', 'content' => [['type' => 'paragraph', 'content' => [['type' => 'text', 'text' => 'Example: {"subject": "<p>Your subject here</p>", "body": "<p>Your body here</p>"}']]]]],
+                ]],
+            ],
```
Orrison, line 108: I only ask just because I am curious, there may not be a need for a change here.

Writing this out like this feels to me like it introduces a slight brittleness if any of the json structure of TipTap changes we would need to carefully update this. Now, if that were to happen, we would be updating a lot of things, including data migrations.

But I just wanted to ask if we considered defining it in here with HTML / Markdown and then using the tool to convert it to it's proper JSON format.

WDYT?

app-modules/service-management/tests/Tenant/Filament/Pages/ManageServiceRequestNotificationAutomationSettingsTest.php
Orrison, line 1: There really should be a test here that validates access with the correct configuration / permission set rather than just testing that super admin makes it past.

database/seeders/NewTenantSeeder.php
```diff
@@ -86,6 +87,9 @@ public function run(): void
 
             // AI
             AiSupportAssistantSettingsSeeder::class,
+
+            // Service Request Notification Automation
+            ServiceRequestNotificationAutomationSettingsSeeder::class,
```
Orrison, line 92: I think we should include this to run in the migrations or a stand alone data migration as a part of this deployment so that current deployed Tenants get it filled as if it owuld be filled if the were new.

Orrison (dismissed):

app-modules/service-management/database/migrations/2026_06_03_091401_create_service_request_notification_automation_email_templates_table.php
danharrin: Changed

app-modules/service-management/src/Models/ServiceRequestNotificationAutomationEmailTemplate.php
danharrin: Cleaned this up

app-modules/service-management/src/Settings/ServiceRequestNotificationAutomationSettings.php
danharrin: I think it's better if we maintain version that corresponds with the actual storage format in general, and it's very unlikely that the TipTap JSON format would change in a breaking way for these simple elements. The merge tag nodes also look significantly worse in the HTML version too: `<span data-type="mergeTag" data-id="type name"></span>`. I think it's better if we stick to using JSON.

database/seeders/NewTenantSeeder.php
danharrin: Yes we have a temporary data migration that covers this already, right? Or am I misunderstanding the suggestion. `app-modules/service-management/database/migrations/2026_06_03_091402_tmp_data_seed_service_request_notification_automation_settings.php`

This seeder + temporary migration pattern has been used elsewhere where the seeded value lives in a separate PHP class, so we avoid referencing PHP classes in migrations.

app-modules/service-management/tests/Tenant/Filament/Pages/ManageServiceRequestNotificationAutomationSettingsTest.php
danharrin: There aren't permissions gating this page, this is part of the global administration settings. Or am I misunderstanding the suggestion?

Orrison: Oh no, you are right. Sorry about that

Orrison (approved):

database/seeders/NewTenantSeeder.php
Orrison: Nah you are right, I think I just misunderstood while reading at first.

PR #1176, opened 2026-06-04
[AIDAPP-1228]: Some users have experienced an issue where there are duplicate new service request statuses

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1228

### Technical Description

> Enforce unique knowledge base status names and merge existing duplicates.

> No

### Cleanup Tasks

> Yes. `.cleanup-tasks/2026_06_05_knowledge_base_and_service_request_status_name_uniqueness_feature.md`

---


Orrison (changes requested):
@payal-canyon apologies, I just updated the ticket description for this if you could please review and apply those changes as well.

TLDR it looks like there was some confusion and the actual issue for the customer was with Service Request Status.
But since you already did the work here for KB Status let's keep that and ALSO do the same changes for SR Status. (more details in the ticket as mentioned)

Orrison (approved):

PR #1181, opened 2026-06-05
[AIDAPP-1221]: Some users have reported that the filter for service request types display as unselectable when collapsed

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1221

### Technical Description

Removes the `disabled` state for categories so they are not disabled. Also removes two other config lines that are just setting the config options to their default values, so they don't do anything.

No

### Cleanup Tasks

N/A

---


payal-canyon (changes requested):
It looks like we may need to handle empty categories separately.
Currently, when selecting a category that has no types or sub-categories (e.g. "Human Resources"), the filter passes a value in the format `category_{uuid}` to the backend. This eventually gets used in a query that expects Service Request Type UUIDs, resulting in the following error:

`SQLSTATE[22P02]: invalid input syntax for type uuid`

We could either hide/disable empty categories in the Type filter or filter out `category_*` values before the type lookup is performed.

payal-canyon (approved):

PR #1182, opened 2026-06-05
[AIDAPP-1222]: Introduce support for file attachments for knowledge base articles

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1222

### Technical Description

- Adds the ability to upload file attachments to Knowledge Base articles and display them as downloadable resources on the portal frontend
- Added `article_attachments` Spatie Media Library collection to `KnowledgeBaseItem`
- Added `SpatieMediaLibraryFileUpload` in the Properties tab (before Notes) with `preserveFilenames() `and AI disclaimer helper text
-  returns a temporary signed URL; returns 403 if the article is not public at the time of request
- "Article Resources" section with paperclip icon, one file per row, hyperlinked for download. Hidden when no attachments exist.
- Covers download authorization (public/private), media collection integrity, and Filament form field presence

> No

### Cleanup Tasks

> No

---


Orrison (changes requested):
Please rebase this on `main` to only contain the changes you did for this ticket.

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItemTest.php
```diff
@@ -215,3 +217,56 @@
...
+
+    $article = KnowledgeBaseItem::factory()->create();
+
+    $article
+        ->addMedia(UploadedFile::fake()->create('test-document.pdf', 100))
+        ->toMediaCollection('article_attachments');
+
+    expect($article->getMedia('article_attachments'))->toHaveCount(1);
+    expect($article->getFirstMedia('article_attachments')->file_name)->toBe('test-document.pdf');
+});
```
Orrison: I don't think this test is needed nor does it belong here. We can just remove it.

This test for is about the edit page and this is testing that the model has media configured at all.

app-modules/knowledge-base/tests/Tenant/Filament/Resources/KnowledgeBaseItems/Pages/EditKnowledgeBaseItemTest.php
```diff
@@ -215,3 +217,56 @@
...
+
+test('multiple files can be added to article_attachments collection', function () {
```
Orrison: This test doesn't really belong here. We should have a test here that is actually testing the upload field on this edit page. That it can properly have media uploaded to it and that then once the edit is saved / executed, it is properly on the model.

So you can probably replace all 3 tests you added here with a test like that. That actually creates a fake file, uploads it via the input, submits, and asserts.

There should be ways in Filament / helpers to do all of that.

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/KnowledgeBaseArticleMediaDownloadControllerTest.php
```diff
@@ -0,0 +1,150 @@
...
+
+    $article = KnowledgeBaseItem::factory()->create(['public' => true]);
+    $article
+        ->addMedia(UploadedFile::fake()->create('document.pdf', 100))
+        ->toMediaCollection('article_attachments');
+
+    $media = $article->getFirstMedia('article_attachments');
+
+    get(route('api.portal.knowledge-base-article.media.download', ['media' => $media->getKey()]))
+        ->assertRedirect();
```
Orrison, line 69: Would `assertDownload` work here?

Orrison (changes requested):

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/KnowledgeBaseArticleMediaDownloadControllerTest.php
monali4-canyon: assertDownload() won't work here — the controller returns a 302 redirect to a temporary signed S3 URL (which has the Content-Disposition: attachment header). The response itself is a redirect, not a streamed download.

Orrison (approved):

PR #1191, opened 2026-06-09
[AIDAPP-1229]: Enhance communication templates for service request types by introducing the ability to dynamically reference users by name.

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1229

### Technical Description

> Refactor email template merge tags and enhance notification data handling

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/service-management/src/Models/ServiceRequest.php
+++ b/app-modules/service-management/src/Models/ServiceRequest.php
@@ -438,9 +438,10 @@ class ServiceRequest extends BaseModel implements Auditable, HasMedia
     /**
      * @return array<string, Closure>
      */
-    public function getTemplateMergeData(?string $timezone = null): array
+    public function getTemplateMergeData(?string $timezone = null, ?string $recipientName = null): array
     {
         return [
+            'recipients name' => fn () => $recipientName ?? '',
             'contact name' => fn () => $this->respondent->{$this->respondent::displayNameKey()},
             'service request number' => fn () => $this->service_request_number,
             'created date' => fn () => ! is_null($timezone) ? $this->created_at->setTimeZone($timezone)->format('M j, Y \a\t h:i A (T)') : $this->created_at->format('M j, Y \a\t h:i A (T)'),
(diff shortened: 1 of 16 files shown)


app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -430,9 +430,10 @@ public function isResolved(): bool
     /**
      * @return array<string, Closure>
      */
-    public function getTemplateMergeData(?string $timezone = null): array
+    public function getTemplateMergeData(?string $timezone = null, ?string $recipientName = null): array
     {
         return [
+            'recipients name' => fn () => $recipientName ?? '',
```
danharrin: While the label of the new tag should be `recipient's name`, lets call it `recipient name` internally to be consistent with `contact name` etc.

app-modules/service-management/src/Models/ServiceRequestTypeEmailTemplate.php
```diff
@@ -149,17 +149,20 @@ public function getSubject(array $mergeData): ?HtmlString
      */
     public static function getMergeTags(): array
```
danharrin, line 150: Just checking, have you looked at all usages of this function to ensure that it is okay that we are changing the data structure that is being returned to be the labels of the tags rather than the "fake merge data" like before?

danharrin (changes requested):

app-modules/service-management/src/Models/ServiceRequestTypeEmailTemplate.php
payal-canyon: Yes - only `setUpRichContent` received the old values, and it discards them (`array_keys` only). The page's editor field is the sole consumer of the values, mapping id => label. Stored keys are unchanged, so existing templates still render - verified against an old-structure template.

danharrin (approved):

PR #1201, opened 2026-06-11
[AIDAPP-1232]: Enhance the management of client types

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1232

### Technical Description

> Implement Contact Type Management feature Introduced a new feature flag for managing contact types. Removed the 'classification' field from the ContactType model and related components. Updated the ContactType model to include an 'is_default' boolean field. Added a migration to handle the addition of the 'is_default' field and the removal of legacy color classifications. Refactored the ContactTypeFactory, Seeder, and Filament resources to align with the new structure. Implemented a ContactTypeObserver to ensure only one contact type can be marked as default. Updated tests to reflect changes in the contact type structure and behavior.
> Also fixes a pre-existing bug where `ContactImporter` wrote to the dropped `assigned_to_id` column, breaking every import.

> No

### Cleanup Tasks

> `.cleanup-tasks/2026_06_11_contact_type_management_feature.md`

---

### DIFF
--- a/app-modules/contact/src/Imports/ContactImporter.php
+++ b/app-modules/contact/src/Imports/ContactImporter.php
@@ -140,7 +140,6 @@ class ContactImporter extends Importer
         /** @var User $user */
         $user = $this->import->user;
 
-        $record->assignedTo()->associate($user);
         $record->createdBy()->associate($user);
     }
 
--- a/app-modules/contact/src/Imports/ContactImporter.php
+++ b/app-modules/contact/src/Imports/ContactImporter.php
@@ -38,6 +38,7 @@ namespace AidingApp\Contact\Imports;
 
 use AidingApp\Contact\Models\Contact;
 use AidingApp\Contact\Models\ContactType;
+use App\Features\ContactTypeManagementFeature;
 use App\Models\User;
 use Filament\Actions\Imports\ImportColumn;
 use Filament\Actions\Imports\Importer;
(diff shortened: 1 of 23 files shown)


app-modules/contact/src/Imports/ContactImporter.php
```diff
@@ -127,7 +140,6 @@ public function beforeCreate(): void
         /** @var User $user */
         $user = $this->import->user;
 
-        $record->assignedTo()->associate($user);
```
danharrin, line 130: Is this removal deliberate?

danharrin (changes requested):

app-modules/contact/src/Imports/ContactImporter.php
payal-canyon: Yes, deliberate. The `assigned_to_id` column was already removed from the contacts table by `app-modules/contact/database/migrations/2026_01_12_234855_remove_assigned_to_id_from_contacts_table.php`, but `ContactImporter::beforeCreate()` still called `$record->assignedTo()->associate($user)`, which writes to that (now non-existent) column. Every contact import therefore failed with `SQLSTATE[42703]: column "assigned_to_id" does not exist`, and because Filament wraps the import rows in a single transaction, the first failure poisoned it and took all rows down with it. Removing the dead line aligns the insert with the current schema; `createdBy()` is kept since `created_by_id` still exists.

danharrin (approved):

PR #1205, opened 2026-06-12
[AIDAPP-1235]: Add failure retry logic to the Support Assistant

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1235

### Technical Description

> Add retry message functionality to assistant chat. Introduced `RetryMessageController` and `RetryMessage` job to handle message retries. Updated routes to include a retry endpoint for messages. Enhanced `PortalAssistantMessageChunk` to include rate limit reset information. Modified `useAssistantChat` to manage retry logic and state.

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
+++ b/app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
@@ -55,7 +55,6 @@ use Illuminate\Foundation\Bus\Dispatchable;
 use Illuminate\Queue\InteractsWithQueue;
 use Illuminate\Queue\SerializesModels;
 use Illuminate\Support\Facades\Log;
-use Prism\Prism\Exceptions\PrismException;
 use Prism\Prism\ValueObjects\Messages\UserMessage;
 use Prism\Prism\ValueObjects\ToolResult;
 use Throwable;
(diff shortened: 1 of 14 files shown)


app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
```diff
@@ -0,0 +1,229 @@
...
+            $aiService = app(AiIntegratedAssistantSettings::class)->getDefaultModel()->getService();
+
+            $nextRequestOptions = $this->thread->messages()->where('is_assistant', true)->latest()->value('next_request_options') ?? [];
+
+            $messages = [
+                new UserMessage($message->content),
+                ...(filled($message->internal_content) ? [new DeveloperMessage($message->internal_content)] : []),
+            ];
+
+            retry(3, function () use ($aiService, $context, $nextRequestOptions, $messages) {
```
danharrin: I don't think we do this in Advising App, and it's probably better that the user is not waiting for 3 attempts and has the feedback the request failed and is given the button to retry. Let me know if I missed where we do this in Advising App though.

app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
danharrin, line 1: Do we need `isStartOfConversation` logic here? Or was it deliberately left out?

widgets/assistant/src/composables/useAssistantChat.js
danharrin, line 1: In Advising App's widget, when a rate limit arrives:

```js
currentResponse.value = '';
resetWordQueue();
```

In this one, `handleStreamChunk` on rate limit only sets the error message on the existing assistant message object but doesn't clear its content. If partial text was already buffered (the stream yielded Text chunks before the Finish chunk), that partial content remains in the message object until `retryMessage()` fires and resets it.

Is this a deliberate decision?

danharrin (changes requested):

app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
payal-canyon: You're right - removed the `retry(3)` from both jobs so failures surface right away with the retry button.

app-modules/ai/src/Jobs/PortalAssistant/RetryMessage.php
payal-canyon: Deliberately omitted. That flag only gates full-history reseeding, which our portal jobs don't do - both `SendMessage` and `RetryMessage` rely on `previous_response_id` for context, not on replaying history. So there's nothing to gate, and keeping `RetryMessage` consistent with `SendMessage` is intentional.

widgets/assistant/src/composables/useAssistantChat.js
payal-canyon: Not deliberate - missed porting Advising App's `resetWordQueue()`. Now clearing message.content + the word queue on both rate limit and hard error (and on retry), so no partial text lingers or bleeds into the retried response.

danharrin (approved):

PR #1209, opened 2026-06-16
[Chore] Address Linting Errors

mittal-canyon:
### Ticket(s) or GitHub Issue

- N/A

### Technical Description

> Address linting errors. Began with 325, ended with 299

> N/A

### Cleanup Tasks

>N/A

---

### DIFF
--- a/app-modules/service-management/src/Models/ServiceRequestTypeEmailPreference.php
+++ b/app-modules/service-management/src/Models/ServiceRequestTypeEmailPreference.php
@@ -45,9 +45,7 @@ use Illuminate\Database\Eloquent\Relations\BelongsTo;
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 /**
- * @property ServiceRequestEmailTemplateType $service_request_email_template_type
- * @property ServiceRequestTypeEmailTemplateRole $service_request_email_template_role
- * @property ServiceRequestNotificationChannel $notification_channel
+ * @mixin IdeHelperServiceRequestTypeEmailPreference
  */
 class ServiceRequestTypeEmailPreference extends Model
 {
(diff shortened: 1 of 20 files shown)


app-modules/service-management/src/Models/ServiceRequestTypeEmailPreference.php
```diff
@@ -45,7 +45,9 @@
 use Illuminate\Database\Eloquent\SoftDeletes;
 
 /**
- * @mixin IdeHelperServiceRequestTypeEmailPreference
```
payal-canyon, line 48: Could you help me understand why the `mixin` docblock was removed? I may be missing some context. If it isn't intentionally being removed, could you please add it back?

payal-canyon (changes requested):

app-modules/service-management/src/Models/ServiceRequestTypeEmailPreference.php
mittal-canyon: @payal-canyon The @mixin was removed because IdeHelperServiceRequestTypeEmailPreference does not exist in _ide_helper_models.php — the IDE helper file is stale and was generated before this model was added. PHPStan was failing with:
PHPDoc tag @mixin contains unknown class AidingApp\ServiceManagement\Models\IdeHelperServiceRequestTypeEmailPreference
if we want to keep then it will need either a PHPStan baseline entry for that error or need to regenerate ide-helper:models

payal-canyon: @mittal-canyon , The `mixin` shouldn't be removed here. `_ide_helper_models.php` is gitignored and regenerated (locally + in CI), so a stale helper is the real cause - not the `mixin` itself. 
so, keep `@mixin IdeHelperServiceRequestTypeEmailPreference` (and drop the manual `@property` lines), then regenerate the helper with `composer generate-helper-files` so the missing class is added to `_ide_helper_models.php`. After that composer lint is clean.

payal-canyon (approved):

PR #1212, opened 2026-06-17
[AIDAPP-1243]: Some actions are not properly gated by permissions

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1243

### Technical Description

Upgrade common to v2.21; add bulk methods to policy stub; add missing policy methods, auth checks on bulk delete actions, and tests

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/alert/src/Filament/Resources/Alerts/Pages/ListAlerts.php
+++ b/app-modules/alert/src/Filament/Resources/Alerts/Pages/ListAlerts.php
@@ -121,7 +121,8 @@ class ListAlerts extends ListRecords
             ])
             ->toolbarActions([
                 BulkActionGroup::make([
-                    DeleteBulkAction::make(),
+                    DeleteBulkAction::make()
+                        ->authorizeIndividualRecords('delete'),
                 ]),
             ])
             ->defaultSort('created_at', 'desc');
--- a/app-modules/alert/src/Policies/AlertPolicy.php
+++ b/app-modules/alert/src/Policies/AlertPolicy.php
@@ -82,6 +82,14 @@ class AlertPolicy
         );
     }
 
+    public function deleteAny(Authenticatable $authenticatable): Response
+    {
+        return $authenticatable->canOrElse(
+            abilities: ['alert.*.delete'],
+            denyResponse: 'You do not have permission to delete any alert.'
+        );
+    }
+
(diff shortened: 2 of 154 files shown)


danharrin (changes requested):
I created a user with a singular role. That role contained every permission group but only viewAny/view permissions (read only).

- I can still associate and dissociate organisation contacts.
- I can also attach and detach roles from any user.
- I can add and remove users to/from departments.
- I can associate/dissociate departments from divisions.
This is probably not an exhaustive list, I just noticed a common pattern and decided to feed back early.

The common pattern here is that the associate/dissociate/attach/detach actions do not user the "update" policy method for the parent record. In Filament out of the box, relation manager associate/dissociate/attach/detach actions are hidden on view pages, so this is not an issue normally. We disable the "Read only relation managers" feature in the panel configuration which introduces this problem.

I think we might want to consider a `configureUsing()` on those actions (and their bulk-equivalents) that calls `->authorize()` with a function that injects the `ManageRelatedRecords | RelationManager $livewire` and authorises `update` on the current user for `$livewire->getOwnerRecord()`.

What do you think @Orrison @payal-canyon?

Orrison:
> I created a user with a singular role. That role contained every permission group but only viewAny/view permissions (read only).
> 
> * I can still associate and dissociate organisation contacts.
> * I can also attach and detach roles from any user.
> * I can add and remove users to/from departments.
> * I can associate/dissociate departments from divisions.
>   This is probably not an exhaustive list, I just noticed a common pattern and decided to feed back early.
> 
> The common pattern here is that the associate/dissociate/attach/detach actions do not user the "update" policy method for the parent record. In Filament out of the box, relation manager associate/dissociate/attach/detach actions are hidden on view pages, so this is not an issue normally. We disable the "Read only relation managers" feature in the panel configuration which introduces this problem.
> 
> I think we might want to consider a `configureUsing()` on those actions (and their bulk-equivalents) that calls `->authorize()` with a function that injects the `ManageRelatedRecords | RelationManager $livewire` and authorises `update` on the current user for `$livewire->getOwnerRecord()`.
> 
> What do you think @Orrison @payal-canyon?

I think this makes sense, proceed with this here and in the other related tickets in the other projects please @payal-canyon. CC @danharrin

danharrin (approved):

PR #1218, opened 2026-06-19
[AIDAPP-1194]: Enhance the API to support viewing a list of users

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1194

### Technical Description

> Enhance the API to support viewing a list of users

> NO

### Cleanup Tasks

> N/A
---

### DIFF
--- a/app/Http/Controllers/Api/V1/Users/ListUsersController.php
+++ b/app/Http/Controllers/Api/V1/Users/ListUsersController.php
@@ -17,7 +17,7 @@
       in the software, and you may not remove or obscure any functionality in the
       software that is protected by the license key.
     - You may not alter, remove, or obscure any licensing, copyright, or other notices
-      of the licensor in the software. Any use of the licensor's trademarks is subject
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
       to applicable law.
     - Canyon GBS Inc. respects the intellectual property rights of others and expects the
       same in return. Canyon GBS® and Aiding App® are registered trademarks of
--- a/app/Http/Controllers/Api/V1/Users/ListUsersController.php
+++ b/app/Http/Controllers/Api/V1/Users/ListUsersController.php
@@ -1,0 +1,88 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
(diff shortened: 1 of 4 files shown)


tests/Tenant/Http/Controllers/Api/V1/Users/ListUsersControllerTest.php
```diff
@@ -0,0 +1,274 @@
...
+    $user = SystemUser::factory()->create();
+    $user->givePermissionTo('user.view-any');
+    Sanctum::actingAs($user, ['api']);
+
+    $targetUser = User::factory()->create();
+
+    $response = getJson(route('api.v1.users.index', [], false));
+    $response->assertOk();
+
+    $item = collect($response['data'])->firstWhere('id', $targetUser->id);
```
payal-canyon: The Advising App tests access the response data via direct indexing (e.g. `$response['data'][0][$responseKey]`) and do not require any PHPStan baseline entries. In contrast, these tests use `collect($response['data'])->firstWhere('id', ...)`, which appears to be the reason for the two new baseline entries.

Could we align this implementation with the Advising App approach instead? You can use `ListProspectsControllerTest` as a reference.

app/Http/Controllers/Api/V1/Users/ListUsersController.php
```diff
@@ -0,0 +1,88 @@
...
+        Gate::authorize('viewAny', User::class);
+
+        return QueryBuilder::for(
+            User::query()->tap(new WithoutAnyAdmin())->with(['roles', 'department', 'permissionsFromRoles'])
+        )
+            ->allowedFilters(
+                AllowedFilter::partial('name'),
+                AllowedFilter::callback('department', function (Builder $query, string $value): void {
+                    $query->whereHas('department', fn (Builder $deptQuery) => $deptQuery->where('name', 'ilike', '%' . $value . '%'));
+                }),
```
payal-canyon: Could we use `AllowedFilter::partial('department', 'department.name')` here instead? Please test it after applying the change to confirm it behaves as expected.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1226, opened 2026-06-23
[AIDAPP-1197]: Ensure Audit Logging Captures Programmatic User as Change Author

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1197

### Technical Description

> Ensure Audit Logging Captures Programmatic User as Change Author
> Add test to check system user audits are stored properly.

> No

### Cleanup Tasks

> No

---

### DIFF
--- a/app-modules/audit/src/Filament/Resources/Audits/Pages/ListAudits.php
+++ b/app-modules/audit/src/Filament/Resources/Audits/Pages/ListAudits.php
@@ -39,6 +39,7 @@ namespace AidingApp\Audit\Filament\Resources\Audits\Pages;
 use AidingApp\Audit\Actions\Finders\AuditableModels;
 use AidingApp\Audit\Filament\Resources\Audits\AuditResource;
 use App\Filament\Tables\Columns\IdColumn;
+use App\Models\SystemUser;
 use App\Models\User;
 use Filament\Actions\ViewAction;
 use Filament\Resources\Pages\ListRecords;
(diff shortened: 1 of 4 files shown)


app-modules/audit/src/Filament/Resources/Audits/Pages/ListAudits.php
```diff
@@ -61,7 +61,7 @@ public function table(Table $table): Table
                     ->sortable(),
                 TextColumn::make('user.name')
                     ->label('Change Agent (User)')
-                    ->sortable(),
+                    ->sortable(query: fn (Builder $query, string $direction): Builder => $query->orderBy(User::select('name')->whereColumn('users.id', 'audits.change_agent_id'), $direction)),
```
danharrin: Can you please let me know what the difference between this query and the Filament default one is? Is the problem that user is a morph relationship or something?

danharrin (changes requested):

app-modules/audit/src/Filament/Resources/Audits/Pages/ListAudits.php
amit-canyon: Hey @danharrin , I checked it again and noticed the query was only looking at the users table. Since change_agent is a polymorphic relationship, I've updated the query to handle the morph types correctly.

app-modules/audit/src/Filament/Resources/Audits/Pages/ListAudits.php
```diff
@@ -54,14 +55,31 @@ class ListAudits extends ListRecords
...
                     ->label('Auditable')
                     ->sortable(),
-                TextColumn::make('user.name')
-                    ->label('Change Agent (User)')
-                    ->sortable(),
+                TextColumn::make('changeAgent.name')
+                    ->label('Change Agent')
+                    ->sortable(
+                        query: fn (Builder $query, string $direction): Builder => $query->orderByRaw(
+                            'COALESCE((' .
```
danharrin: Instead of a coalesce which would do two table lookups for system users (check if user exists first, could be slow on a large table like audits), maybe do a `case` that checks the agent type and does a different query based on that?

danharrin (changes requested):

app/Providers/AppServiceProvider.php
```diff
@@ -101,6 +102,7 @@ public function register(): void
     public function boot(): void
     {
         Relation::morphMap([
+            'user' => User::class,
```
danharrin: Seems like my database already has `user` in the audits table, can you explained what you saw here? I don't see the morph map entry elsewhere which is strange, but would have expected it

danharrin (changes requested):

app/Providers/AppServiceProvider.php
amit-canyon: removed.

danharrin (approved):

PR #1231, opened 2026-06-24
[AIDAPP-1258]: Optimize based on prior release: Contact Type Management Feature

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1258

### Technical Description

> Optimize based on prior release: Contact Type Management Feature

> NO

### Cleanup Tasks

> N/A

---


app-modules/contact/database/migrations/2026_06_11_183351_add_is_default_to_contact_types_table.php
```diff
@@ -56,31 +55,19 @@ public function up(): void
...
-                 * write/read classification while the feature is being rolled out.
-                 */
-                $table->string('classification')->nullable()->change();
             });
 
             foreach ($this->colorMap as $from => $to) {
                 DB::table('contact_types')
                     ->where('color', $from)
                     ->update(['color' => $to]);
             }
```
Orrison: @payal-canyon can this color re-mapping here, in the down, and the test for this in `tests/TenantMigrationTests.php` be removed now?

It wasn't mentioned in the cleanup task. But would it be needed anymore? Wouldn't new Tenants already have the correct values?

Orrison (changes requested):

app-modules/contact/database/migrations/2026_06_11_183351_add_is_default_to_contact_types_table.php
payal-canyon: @Orrison yeah, you're right! We don't need it anymore. I forgot to include that in the cleanup file.

Orrison: @mittal-canyon then let's move forward with removing that from the migration as well and remove the test in `tests/TenantMigrationTests.php`

Orrison (approved):

PR #1244, opened 2026-07-01
[AIDAPP-1267]: Notify and restrict access for expired subscriptions

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1267

### Technical Description

> Notify and restrict access for expired subscriptions

> No

### Cleanup Tasks

> Yes. `.cleanup-tasks/2026_07_08_subscription_expiration_feature.md`

---


Orrison (changes requested):
@payal-canyon I think it would be much more efficient for me to have the Olympus related ticket settled and approved first before I dug too much into this one. As there may be changes you need to make from my feedback from that one anyway.

So once we get that one settled approved / merged, please re-request my review here.

Orrison (changes requested):
Requested some changes in the related Advising App PR, I would say once the Advising App one is settled, make the same changes here, then re-request my review.

Orrison (changes requested):
Just left one final feedback comment on https://github.com/canyongbs/advisingapp/pull/2659 which would apply here as well.

Orrison (approved):

PR #1249, opened 2026-07-03
[AIDAPP-1274]: Introduce the Knowledge Base report

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1274

### Technical Description

- Implemented `KnowledgeBaseArticlesOverTimeBarChart` for visualizing article creation over time.
- Created `KnowledgeBaseArticlesTable` for displaying articles with filtering and export functionality.
- Developed `KnowledgeBaseConcernsByStatusDonutChart` to show concerns categorized by status.
- Added `KnowledgeBaseStats` widget for displaying key statistics related to knowledge base articles.
- Introduced tests for each widget to ensure correct functionality and data handling, including filtering by categories and status.

No

### Cleanup Tasks

No

---


app-modules/report/src/Filament/Widgets/KnowledgeBaseStats.php
```diff
@@ -0,0 +1,116 @@
...
+            : Cache::tags(["{$this->cacheTag}"])->remember(
+                'total-views-count',
+                now()->addHours(24),
+                fn () => KnowledgeBaseItem::query()->sum('portal_view_count')
+            );
+
+        return [
+            Stat::make('Total Articles', Number::abbreviate($totalArticles, maxPrecision: 2)),
+            Stat::make('Healthy Articles', Number::abbreviate($healthyArticles, maxPrecision: 2)),
+            Stat::make('Total Views', Number::abbreviate((int) $totalViews, maxPrecision: 2)),
```
rebecca-canyon: Because these are all counts and not percentages, we don't need precision for them.

app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesByCategoryDonutChart.php
```diff
@@ -0,0 +1,150 @@
...
+            'rgb(34, 197, 94)',
+            'rgb(249, 115, 22)',
+            'rgb(168, 85, 247)',
+            'rgb(236, 72, 153)',
+            'rgb(14, 165, 233)',
+            'rgb(245, 158, 11)',
+            'rgb(99, 102, 241)',
+            'rgb(20, 184, 166)',
+            'rgb(239, 68, 68)',
+        ];
```
rebecca-canyon: Rather than defining the colors, we can assign them randomly; an example can be found in `app-modules/report/src/Filament/Widgets/StudentOverviewGenderDoughnutChart.php` in Advising App

app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesTable.php
```diff
@@ -0,0 +1,194 @@
...
+    public function mount(string $cacheTag): void
+    {
+        $this->cacheTag = $cacheTag;
+    }
+
+    #[On('refresh-widgets')]
+    public function refreshWidget(): void
+    {
+        $this->dispatch('$refresh');
+    }
```
rebecca-canyon, line 80: I believe that we'll want to add this snippet to the other widgets as well

app-modules/report/tests/Tenant/Filament/Widgets/KnowledgeBaseArticlesOverTimeBarChartTest.php
```diff
@@ -0,0 +1,94 @@
...
+use AidingApp\Report\Filament\Widgets\KnowledgeBaseArticlesOverTimeBarChart;
+
+it('returns correct articles over time data without filters', function () {
+    $status = KnowledgeBaseStatus::factory()->create();
+    $category = KnowledgeBaseCategory::factory()->create();
+
+    KnowledgeBaseItem::factory()->count(2)->state([
+        'status_id' => $status->id,
+        'category_id' => $category->id,
+        'created_at' => now()->subMonths(2),
```
rebecca-canyon: I think that `subMonths()` could potentially lead to issues at the end of the month (e.g., July 31 sub month would be June 31, which doesn't exist, and I think translates to July 1); but I think there's a `subMonthsWithoutOverflow()` that you could use here that would avoid that.

app-modules/report/tests/Tenant/Filament/Widgets/KnowledgeBaseArticlesOverTimeBarChartTest.php
```diff
@@ -0,0 +1,94 @@
...
+    ])->create();
+
+    KnowledgeBaseItem::factory()->count(3)->state([
+        'status_id' => $status->id,
+        'category_id' => $category->id,
+        'created_at' => now()->subMonth(),
```
rebecca-canyon: Same as above

app-modules/report/tests/Tenant/Filament/Widgets/KnowledgeBaseArticlesOverTimeBarChartTest.php
```diff
@@ -0,0 +1,94 @@
...
+
+it('filters articles over time by category', function () {
+    $status = KnowledgeBaseStatus::factory()->create();
+    $categoryA = KnowledgeBaseCategory::factory()->create();
+    $categoryB = KnowledgeBaseCategory::factory()->create();
+
+    KnowledgeBaseItem::factory()->count(2)->state([
+        'status_id' => $status->id,
+        'category_id' => $categoryA->id,
+        'created_at' => now()->subMonth(),
```
rebecca-canyon: Same as above

app-modules/report/tests/Tenant/Filament/Widgets/KnowledgeBaseArticlesOverTimeBarChartTest.php
```diff
@@ -0,0 +1,94 @@
...
+    ])->create();
+
+    KnowledgeBaseItem::factory()->count(3)->state([
+        'status_id' => $status->id,
+        'category_id' => $categoryB->id,
+        'created_at' => now()->subMonth(),
```
rebecca-canyon: Same as above

app-modules/report/tests/Tenant/Filament/Widgets/KnowledgeBaseArticlesTableTest.php
```diff
@@ -0,0 +1,131 @@
...
+
+    livewire(KnowledgeBaseArticlesTable::class, [
+        'cacheTag' => 'test-kb-articles-table-export-notification',
+        'pageFilters' => [],
+    ])
+        ->callTableAction(ExportAction::class)
+        ->assertNotified();
+});
+
+it('filters records by status using the table-level status filter', function () {
```
rebecca-canyon, line 109: This looks good, but we might want to add a little more to verify that filtering for `$statusB` also works after having already filtered for `$statusA` and that we're still only seeing the expected results.

There should also be tests for the other filters on the table, as well, similar to this one.

rebecca-canyon (changes requested):

app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesTable.php
monali4-canyon: It is already present on all KB report widgets, either directly (`KnowledgeBaseArticlesTable`) or inherited from their base classes (`ChartReportWidget`, `BarChartReportWidget`, `StatsOverviewReportWidget`).

What do you mean by other widgets?

rebecca-canyon: Ah, you're correct; I was referring to the other widgets you had created, but you're right, they do inherit `refreshWidget` from their parents.

However, refreshing doesn't appear to update the report on my end; is it working for you? And if not, could you investigate that?

rebecca-canyon (changes requested):

rebecca-canyon (approved):

PR #1251, opened 2026-07-06
[AIDAPP-1275]: Introduce a new report category for Artificial Intelligence

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1275

### Technical Description

> Introduce a new report category for Artificial Intelligence

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/resources/views/filament/pages/coming-soon.blade.php
+++ b/resources/views/filament/pages/coming-soon.blade.php
@@ -31,4 +31,30 @@
     
     </COPYRIGHT>
 --}}
-<x-filament-panels::page>Coming Soon!</x-filament-panels::page>
+<x-filament-panels::page>
+    <div class="flex min-h-[60vh] flex-col items-center justify-center gap-6 text-center">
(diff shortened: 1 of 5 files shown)


resources/views/filament/pages/coming-soon.blade.php
danharrin, line 1: Was this new design requested in the ticket? We reuse this view across multiple pages so this change would mean that non-report pages start showing this too

danharrin (changes requested):

resources/views/filament/pages/coming-soon.blade.php
mittal-canyon: No nothing reuested in ticket i just took reference from olympus for this. because simple comming soon text was there , should i keep the older version only  or we can go with new version ? and yes all pages look like this.

mittal-canyon: Reverted to the existing one!

danharrin (approved):

PR #1256, opened 2026-07-08
[AIDAPP-1279]: Eliminate tasks as a concept in Aiding App

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1279

### Technical Description

- Deleted the entire `app-modules/task/` module (models, enums, policies, observers, notifications, Filament resources, factories, seeders, tests)
- Created a drop migration to remove `tasks`, `confidential_task_users`, `confidential_task_departments`, and `confidential_task_projects` tables
- Moved existing task migrations to migrations to preserve migration history
- Removed task relationships from `User`, `Contact`, and `Project` models
- Removed task-related Filament pages/widgets from Contact, Project, and Report modules
- Removed `MyTasks` dashboard widget
- Removed tasks count from the Utilization Metrics API and its test
- Removed task navigation, routes, and feature flags from the Knowledge Management Portal (Vue)
- Removed `has_tasks` from portal API responses (3 controllers)
- Removed `canyon-gbs/aiding-app-task` from composer.json and updated composer.lock
- Cleaned up `DatabaseSeeder` and legacy UUID model list

No

### Cleanup Tasks

No

---


database/migrations/2026_07_08_254132_drop_task_tables.php
```diff
@@ -34,20 +34,20 @@
...
-    public function title(): string
+return new class () extends Migration {
+    public function up(): void
     {
-        return 'My Tasks for Contacts';
+        DB::transaction(function () {
+            Schema::dropIfExists('confidential_task_projects');
+            Schema::dropIfExists('confidential_task_departments');
+            Schema::dropIfExists('confidential_task_users');
+            Schema::dropIfExists('tasks');
```
payal-canyon, line 52: There are just a few cleanup items remaining that we should remove as part of this PR:

1. There's still one route left in `app-modules/portal/routes/web.php`. Could we also remove the `/portal/tasks` route?
2. Could we also add a migration to remove the related audit records? For example:
   `DELETE FROM audits WHERE auditable_type = 'task';`
3. We should also remove the associated permission and permission group. Could you add a migration to drop those as well?
4. Remove the `PermissionDescription::Task` enum case.
5. Remove the orphaned `task_management` and `tasks` language keys from:
        `resources/lang/en/global.php`
        `resources/lang/es/global.php`
6. Remove the task-related entries from `phpstan-baseline.php`.
7. In `app-modules/engagement/resources/views/components/filters.blade.php`, remove the now-unused **"Open Tasks"** checkbox from the engagement filters.

payal-canyon (changes requested):

database/migrations/2026_07_10_213642_delete_task_audit_records.php
```diff
@@ -34,20 +34,16 @@
...
-class ContactTasks extends TasksWidget
-{
-    public function title(): string
+return new class () extends Migration {
+    public function up(): void
     {
-        return 'My Tasks for Contacts';
+        DB::table('audits')
+            ->where('auditable_type', 'task')
+            ->delete();
```
payal-canyon: Let's remove this file and move its contents into the existing migration: `database/migrations/2026_07_08_254132_drop_task_tables.php`.

database/migrations/2026_07_10_546219_delete_task_permissions_and_group.php
```diff
@@ -34,38 +34,45 @@
...
-        if (is_null($task->created_by)) {
-            $user = auth()->user();
+            if ($groupId) {
+                DB::table('model_has_permissions')
+                    ->whereIn('permission_id', function (Builder $query) use ($groupId) {
+                        $query->select('id')
+                            ->from('permissions')
+                            ->where('group_id', $groupId);
+                    })
+                    ->delete();
```
payal-canyon: This looks good overall, but we already have a trait for handling this: `CanModifyPermissions`. We should use that here instead, as it's responsible for both creating and deleting permissions.

For reference, you can take a look at: `database/migrations/2026_05_08_085249_delete_unused_permissions_and_permission_groups.php`

One small thing to note: in that migration, the permission changes aren't wrapped in a `DB::transaction()`. For this implementation, could we wrap those operations in a transaction to ensure they're executed atomically?

payal-canyon (changes requested):

payal-canyon (approved):

PR #1266, opened 2026-07-14
[AIDAPP-1319]: Introduce a managed contact for users

danharrin:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1319

### Technical Description

Introduces managed contacts for users, which get immediately created and synced. They are read-only and allow the user to sign in to the portal via the user menu.

No

### Cleanup Tasks

`.cleanup-tasks/2026_07_13_managed_contact_feature.md`

---


app-modules/contact/src/Filament/Resources/ContactResource/Pages/EditContact.php
danharrin, line 1: This is a page-level check since super admins should not be able to bypass it, which would be the case if it was in a policy

app-modules/contact/src/Services/ManagedContactService.php
danharrin, line 1: Opted for a service here instead of actions since the create/sync logic should be kept in sync and the methods have shared helpers

Orrison (approved):

Orrison (dismissed):

Orrison (approved):

PR #1268, opened 2026-07-14
[AIDAPP-1315]: Introduce pipeline stage classifications

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1315

### Technical Description

- New enum: `PipelineStageClassification` with Planning, InProgress, Complete cases
- Migration: Adds `classification` column to `pipeline_stages` table with Planning as the default for existing rows (uses initial() modifier)
- Feature flag: `PipelineStageClassificationFeature` gates the field visibility; activated during migration
- Model: Updated `PipelineStage` with classification in fillable and casts
- UI: Added classification Select (native) to the stages Repeater on Create, Edit, and Manage Pipelines pages; added badge display on View Pipeline
- Defaults: New pipelines are seeded with one stage per classification (Planning → In Progress → Complete)
- Factory/Request factories: Updated to include classification data
- Tests: `PipelineStageClassificationTest` covers enum values, model casting, DB persistence, and labels

No

### Cleanup Tasks

Feature Flag Added: `.cleanup-tasks/2026_07_14_pipeline_stage_classification_feature.md`

---

### DIFF
--- a/app-modules/project/database/migrations/2026_07_14_514313_add_classification_to_pipeline_stages_table.php
+++ b/app-modules/project/database/migrations/2026_07_14_514313_add_classification_to_pipeline_stages_table.php
@@ -17,7 +17,7 @@
       in the software, and you may not remove or obscure any functionality in the
       software that is protected by the license key.
     - You may not alter, remove, or obscure any licensing, copyright, or other notices
-      of the licensor in the software. Any use of the licensor's trademarks is subject
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
       to applicable law.
     - Canyon GBS Inc. respects the intellectual property rights of others and expects the
       same in return. Canyon GBS® and Aiding App® are registered trademarks of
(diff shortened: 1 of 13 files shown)


Orrison (changes requested):
@rebecca-canyon please request my final review on this one once you have approved it.

Note that @monali4-canyon is going to be OOO tomorrow, so if there are any changes needed can you please make them directly on this branch?

rebecca-canyon (dismissed):

app-modules/project/database/migrations/2026_07_14_514313_add_classification_to_pipeline_stages_table.php
```diff
@@ -0,0 +1,68 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('pipeline_stages', function (Blueprint $table) {
+                $table->string('classification')
+                    ->initial(PipelineStageClassification::Planning->value)
+                    ->default(PipelineStageClassification::Planning->value);
```
Orrison: If you are setting the default to be the same as the `initial` then there is no point in have the `initial`. You would only define both if they were different.

Initial just allows you to add a column where all existing column should have it created set as X, but you don't actually want the default of that column to be that for new records.

If you are already setting a default for the same value, it will set them all to be that for you anyways.

So I would get rid of the `initial` here

Orrison (changes requested):

Orrison (approved):

PR #1269, opened 2026-07-14
[AIDAPP-1317]: Introduce address auto complete for organization and contact profiles

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1317

### Technical Description

Create address input, use in place of existing address inputs for contacts and organizations; drop now extraneous address columns

Three new ENV variables have been added:
AWS_GEO_PLACES_ACCESS_KEY_ID
AWS_GEO_PLACES_SECRET_ACCESS_KEY
AWS_GEO_PLACES_REGION

### Cleanup Tasks

N/A

---


danharrin (dismissed):
Unfortunately, I think we may have gone down the wrong path here: I don't think the data model should change, it should act like the address autocomplete from Advising App: https://github.com/canyongbs/advisingapp/blob/898d8a3aa6d231577d3dc96f5154f7d7c5cc0e36/app-modules/prospect/src/Filament/Resources/Prospects/Pages/EditProspect.php#L209

I checked with Kevin and he believes that the data model here should not change.

app-modules/contact/src/Filament/Resources/ContactResource/Pages/CreateContact.php
```diff
@@ -133,6 +138,36 @@ public function form(Schema $schema): Schema
...
+                                        ->send();
+
+                                    session()->put('has_aws_geo_places_error_notification_sent', true);
+                                }
+
+                                report($exception);
+                            }
+                        })
+                        ->saved(false),
                     TextInput::make('address')
```
payal-canyon: This is the one page where autocomplete writes fields the form doesn't show. `city`, `state`, and `postal` have no `TextInput` here, so the new `onItemSelected` sets values the user can't see or correct, and `mutateFormDataBeforeCreate()` is needed to persist them.

Since `resolveCity()` returns when AWS doesn't return a Locality, a contact can be created with a blank city that can only be fixed by re-opening it in Edit.

Could we add the missing `city`/ `state` / `postal` inputs to match `EditContact` and drop the hook? That brings all four pages in line with the Advising App pattern, which relies on real form fields rather than a mutate hook.

app/Services/AwsGeoPlacesService.php
```diff
@@ -0,0 +1,108 @@
...
+
+namespace App\Services;
+
+use App\DataTransferObjects\AutocompletedAddress;
+use Aws\GeoPlaces\GeoPlacesClient;
+use Exception;
+use Throwable;
+
+class AwsGeoPlacesService
+{
```
payal-canyon, line 45: `AwsGeoPlacesService` is new in this project but doesn't currently have any test coverage. The Advising App already has `tests/Tenant/Unit/AwsGeoPlacesServiceTest.php`, which covers the `resolveCity()` country branches (US/CA/MX/GB), the `Code3`/`Region.Name` fallbacks, and the missing-fields case.

Could we add similar test coverage here as well?

payal-canyon (changes requested):

payal-canyon (changes requested):
Image

The **Address** field is displayed twice when creating or editing a Contact or Organization. From a user perspective, this is confusing. Could we remove the duplicate so the address is only shown once?

payal-canyon (approved):

PR #1281, opened 2026-07-20
[AIDAPP-1324]: Optimize based on prior release: Service Request Type Visibility Restrictions Cleanup

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1324

### Technical Description

- Optimize based on prior release: Service Request Type Visibility Restrictions Cleanup

No

### Cleanup Tasks

No

---


danharrin (changes requested):
I think we should remove the `visibilityRestrictionsEnabled` param completely from `BuildContactServiceRequestTypeTree`, unless you can find a place where it is ever false?

danharrin (approved):

PR #1287, opened 2026-07-20
[AIDAPP-1331]: Optimize based on prior release: Reporting Feature

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1331

### Technical Description

> Optimize based on prior release: Reporting Feature

> NO

### Cleanup Tasks

> N/A

---


Orrison (changes requested):
One thing to look at before we merge this in that was in the cleanup of the other app changes similar to this is. Are we able to remove the `report-library` permissions now?

IF nothing is using them, then we should drop those permissions from the DB.

If you are finding them being used still. Please let me know here.

mittal-canyon:
> One thing to look at before we merge this in that was in the cleanup of the other app changes similar to this is. Are we able to remove the `report-library` permissions now?
> 
> IF nothing is using them, then we should drop those permissions from the DB.
> 
> If you are finding them being used still. Please let me know here.

@Orrison Thanks for calling this out. I audited the codebase and there are no active authorization checks using report-library permissions anymore. Current access checks now rely on reporting permissions only. The only remaining report-library references were historical migration artifacts. so now adding a missing cleanup migration to remove stale DB rows for `report-library` across web and api guards.

app-modules/report/database/migrations/2026_07_21_000004_data_remove_report_library_permissions.php
Orrison, line 1: We typically use the trait `database/migrations/Concerns/CanModifyPermissions.php` to handle permission creation and deletion.

We should use that here unless there is a specific reason not to

Orrison (changes requested):

Orrison (approved):

PR #1306, opened 2026-07-24
[AIDAPP-1338]: Some users have reported that the portal is not detecting their timezone

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1338

### Technical Description

> Implement timezone handling and formatting across the portal.

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/portals/knowledge-management/src/Pages/Licenses.vue
+++ b/portals/knowledge-management/src/Pages/Licenses.vue
@@ -41,6 +41,7 @@
     import Page from '../Components/Page.vue';
     import PageCard from '../Components/PageCard.vue';
     import { consumer } from '../Services/Consumer';
+    import formatDateTime from '../Services/FormatDateTime.js';
 
     const productLicenses = ref({});
     const { get } = consumer();
@@ -146,7 +147,7 @@
                                 <p class="mb-2 font-semibold text-gray-800">
                                     Expiration:
                                     <span class="font-normal text-gray-700">
-                                        {{ activeLicense.formatted_expiration_date ?? 'No Expiration' }}
+                                        {{ formatDateTime(activeLicense.expiration_date, { dateOnly: true, utc: true }) ?? 'No Expiration' }}
                                     </span>
                                 </p>
                                 <p class="flex gap-2 font-semibold text-gray-800">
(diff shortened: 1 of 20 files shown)


portals/knowledge-management/src/Pages/Licenses.vue
```diff
@@ -193,7 +199,12 @@
                                 <p class="mb-2 font-semibold text-gray-800">
                                     Expiration:
                                     <span class="font-normal text-gray-700">
-                                        {{ expiredLicense.formatted_expiration_date }}
+                                        {{
+                                            formatDateTime(expiredLicense.expiration_date, {
+                                                dateOnly: true,
+                                                utc: true,
+                                            })
+                                        }}
```
Orrison: Is there a particular reason we need this one to be in UTC? Tracing back through where we get this from I see that we are just getting it as UTC. But if the expectation is that all dates and datetimes in the portal are in the users timezone I think we need to make sure this is too.

portals/knowledge-management/src/Stores/timezone.js
```diff
@@ -0,0 +1,58 @@
...
+    const displayTimezone = ref(null);
+    const browserTimezone = ref(null);
+
+    function setDisplayTimezone(value) {
+        displayTimezone.value = value ?? null;
+    }
+
+    function detectBrowserTimezone() {
+        browserTimezone.value = Intl.DateTimeFormat().resolvedOptions().timeZone;
+    }
```
Orrison: I would like to simplify this by NOT storing the `browserTimezone` in the store.

If/when the browser time is needed in `portals/knowledge-management/src/Services/FormatDateTime.js` or anywhere else we should just call `Intl.DateTimeFormat().resolvedOptions().timeZone`.

Caching it adds a lot of overhead when the `Intl.DateTimeFormat().resolvedOptions().timeZone` is synchronous and cheap already.

Orrison (changes requested):

portals/knowledge-management/src/Pages/Licenses.vue
Orrison: Discussed offline, good to keep

Orrison (approved):

PR #1307, opened 2026-07-24
[AIDAPP-1357]: Update the service requests report to include a new filter for one or more service request types

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1357

### Technical Description

- Added a grouped "Service Request Types" multi-select filter
- Added "My Affiliated Types" button to auto-select types the user manages/audits.
- Added a "Clear" button, disabled when nothing is selected.
- Wired the filter into all report widgets (tables, stats, charts).
- Added shared `getServiceRequestTypes()` accessor to the filters trait.

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/report/src/Filament/Pages/ServiceRequests.php
+++ b/app-modules/report/src/Filament/Pages/ServiceRequests.php
@@ -44,17 +44,25 @@ use AidingApp\Report\Filament\Widgets\ServiceRequestsStats;
 use AidingApp\Report\Filament\Widgets\ServiceRequestsTable;
 use AidingApp\Report\Filament\Widgets\ServiceRequestStatusDistributionDonutChart;
 use AidingApp\Report\Filament\Widgets\ServiceRequestTypesTable;
+use AidingApp\ServiceManagement\Models\ServiceRequestType;
+use AidingApp\ServiceManagement\Models\ServiceRequestTypeCategory;
 use App\Enums\Feature;
 use App\Enums\ReportLibraryNavigationGroup;
 use App\Filament\Clusters\ReportLibrary;
 use App\Models\User;
 use BackedEnum;
+use Filament\Actions\Action;
 use Filament\Forms\Components\DatePicker;
+use Filament\Forms\Components\Select;
 use Filament\Pages\Dashboard;
 use Filament\Pages\Dashboard\Concerns\HasFiltersForm;
+use Filament\Schemas\Components\Actions;
 use Filament\Schemas\Components\Section;
 use Filament\Schemas\Components\Utilities\Get;
+use Filament\Schemas\Components\Utilities\Set;
 use Filament\Schemas\Schema;
+use Filament\Support\Icons\Heroicon;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Facades\Gate;
 use UnitEnum;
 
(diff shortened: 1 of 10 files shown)


app-modules/report/src/Filament/Pages/ServiceRequests.php
```diff
@@ -119,6 +153,93 @@ public function filtersForm(Schema $schema): Schema
...
+     */
+    public function getAffiliatedServiceRequestTypeIds(): array
+    {
+        /** @var User $user */
+        $user = auth()->user();
+
+        $departmentId = $user->department?->getKey();
+
+        return ServiceRequestType::query()
+            ->where(function (Builder $query) use ($user): void {
```
payal-canyon: This is missing `->withoutArchived()`, which the options query above it already applies. Clicking `My Affiliated Types` puts archived type IDs into the filter state - the widgets filter by them, but they don't appear in the dropdown, so the user can't see or deselect them individually.

app-modules/report/src/Filament/Pages/ServiceRequests.php
```diff
@@ -119,6 +153,93 @@ public function filtersForm(Schema $schema): Schema
...
+            return implode(' › ', $segments);
+        };
+
+        $grouped = [];
+        $uncategorized = [];
+
+        ServiceRequestType::query()
+            ->withoutArchived()
+            ->with('categories:id')
+            ->orderBy('name')
```
payal-canyon: The dropdown doesn't follow the service catalog hierarchy - group labels are `ksorted` alphabetically and types use `orderBy('name')`, so both `service_request_type_categories.sort` and the `service_request_category_types.sort` pivot are ignored (sort is selected in the categories query but never read). `BuildContactServiceRequestTypeTree` already walks the catalog in the right order via `categorySortMap()` - reuse it.

app-modules/report/tests/Tenant/Filament/Pages/ServiceRequestTest.php
```diff
@@ -89,3 +120,304 @@
...
+    $user = User::factory()->create(['timezone' => 'UTC']);
+
+    grantServiceRequestsReportAccess($user);
+
+    actingAs($user);
+
+    $type = ServiceRequestType::factory()->create();
+
+    livewire(ServiceRequests::class)
+        ->set('filters.serviceRequestTypes', [$type->getKey()])
```
payal-canyon, line 452: The Clear button's disabled state is a stated requirement but has no test - `assertActionDisabled` when the filter is empty, `assertActionEnabled` after setting a type. There's also no test that "My Affiliated Types" excludes archived types. Let's cover both.

payal-canyon (changes requested):

app-modules/report/src/Filament/Pages/ServiceRequests.php
```diff
@@ -95,6 +103,32 @@ public static function canAccess(): bool
     public function filtersForm(Schema $schema): Schema
     {
         return $schema->components([
+            Section::make()
+                ->schema([
+                    Select::make('serviceRequestTypes')
+                        ->label('Service Request Types')
+                        ->options(fn (): array => $this->getServiceRequestTypeOptions())
```
payal-canyon: Could we reuse the existing Service Request Types tree here? The Service Requests list already implements this filter - `SelectTree` + `ListServiceRequests::buildTypeTreeOptions()` - and it would keep the behavior consistent.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1321, opened 2026-07-28
[AIDAPP-1346]: Introduce Kanban board for pipelines and pipeline tasks

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1346

### Technical Description

> Introduce Kanban board for pipelines and pipeline tasks

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/project/src/Livewire/PipelineEntryKanban.php
+++ b/app-modules/project/src/Livewire/PipelineEntryKanban.php
@@ -146,8 +146,20 @@ class PipelineEntryKanban extends Component implements HasForms, HasActions
                     ->columnSpanFull(),
                 ...PipelineEntryForm::components($this->pipeline),
             ])
-            ->action(function (array $data, array $arguments) {
-                $stage = $this->pipeline->stages()->whereKey($arguments['stage'])->firstOrFail();
+            ->action(function (array $data, array $arguments, Action $action) {
+                $stage = $this->pipeline->stages()->whereKey($arguments['stage'] ?? null)->first();
+
+                if (! $stage) {
+                    Notification::make()
+                        ->danger()
+                        ->title('Pipeline entry could not be added')
+                        ->body('The selected stage does not belong to this pipeline.')
+                        ->send();
+
+                    $action->halt();
+
+                    return;
+                }
 
                 $dataArray = [
                     'name' => $data['name'],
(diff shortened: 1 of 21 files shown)


app-modules/project/src/Livewire/PipelineEntryKanban.php
payal-canyon: The `movedEntry()` method updates `pipeline_stage_id` without any policy check, so a user holding only `pipeline.view-any` (which already renders this page, per the existing test) can move entries. Could you please add an `authorize('update', $this->pipeline)` check before the update?

Also, The `$pipeline` parameter is accepted but never used, and neither `$fromStage` nor `$toStage` is validated against the current pipeline, so a crafted request can move any entry into any other project's stage. Could you please resolve the entry and target stage through `$this->pipeline->entries()` and `$this->pipeline->stages()` instead?

payal-canyon: `moveEntry()` is an exact duplicate of `movedEntry()` and nothing calls it, but as a public Livewire method it is still remotely callable and carries the same gap. Could you please remove it?

payal-canyon: The `addEntry` action writes the client-supplied `$arguments['stage']` straight to `pipeline_stage_id` and has no `->authorize()`, unlike the equivalent action in `ProjectWorkPipelineWidget`. Could you please validate the stage belongs to this pipeline and add the authorize check?

payal-canyon: `getStages()` plucks stages without an `orderBy('order')`, so the Kanban columns can appear in a different order on each load. The rest of the module already orders by order - could you please apply the same here?

payal-canyon: `getPipelineEntries()` loads entries with no eager loading, and each card renders `$entry->organizable?->full_name`, producing one extra query per card. Could you please add `->with('organizable')` and an explicit order, matching how the widget's table loads its data?

payal-canyon, line 58: This component currently has no test coverage. Could you please create test cases for:
- addEntry creates a pipeline entry with the expected attributes.
- An entry created from a column's Add button is assigned to that column's stage, since the stage comes from the action arguments rather than a form field.
- Entry cards render their expected content on the board.
- Stages render in order sequence.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1322, opened 2026-07-29
[AIDAPP-1350] & [AIDAPP-1347]: Introduce the ability to preload new service request types with the base service request types templates

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1350
- https://canyongbs.atlassian.net/browse/AIDAPP-1347

### Technical Description

> Introduce the ability to preload new service request types with the base service request types templates

> No

### Cleanup Tasks

> Yes, FeatureFlag: `ProloadServiceRequestTypeFeature`

---


app-modules/service-management/src/Observers/ServiceRequestTypeObserver.php
```diff
@@ -50,6 +53,13 @@ public function creating(ServiceRequestType $serviceRequestType): void
         $serviceRequestType->sort = DB::raw('(select coalesce(max(sort), 0) + 1 from service_request_types)'); /** @phpstan-ignore assign.propertyType (Assigns a DB raw Expression so the next global sort value is computed atomically in SQL, which the property's int|null type does not allow.) */
     }
 
+    public function created(ServiceRequestType $serviceRequestType): void
+    {
+        if (ProloadServiceRequestTypeFeature::active() && (app(ServiceRequestNotificationAutomationSettings::class)->preload_new_service_request_types || app(ServiceRequestNotificationAutomationSettings::class)->use_custom_templates)) {
+            app(PreloadServiceRequestTypeBaseTemplates::class)->execute($serviceRequestType);
+        }
+    }
```
Orrison: I think this should run ONLY if `preload_new_service_request_types` is `true`.

Both the global admin version and the customer admin version should have that as a toggle, which can be tied to the same value stored in settings.

The customer admin version of the templates pages purpose is to get the customers a place to override the global settings.
I think it makes sense that the templates themselves be stored separate in the DB between global and partner admin. But I think it should be fine that both pages set and use the same setting `preload_new_service_request_types`

app-modules/service-management/src/Models/ServiceRequestCustomEmailTemplate.php
```diff
@@ -0,0 +1,87 @@
...
+use AidingApp\ServiceManagement\Filament\Blocks\SurveyResponseEmailTemplateTakeSurveyButtonBlock;
+use App\Models\BaseModel;
+use Filament\Forms\Components\RichEditor\Models\Concerns\InteractsWithRichContent;
+use Filament\Forms\Components\RichEditor\Models\Contracts\HasRichContent;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+
+/**
+ * @mixin IdeHelperServiceRequestCustomEmailTemplate
+ */
+class ServiceRequestCustomEmailTemplate extends BaseModel implements HasRichContent
```
Orrison: Let's not use the base model here and let's use the new HasUuids v7 trait

app/Features/ProloadServiceRequestTypeFeature.php
```diff
@@ -0,0 +1,47 @@
...
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace App\Features;
+
+use App\Support\AbstractFeatureFlag;
+
+class ProloadServiceRequestTypeFeature extends AbstractFeatureFlag
```
Orrison, line 41: This is misspelled, but honestly since it will only live for a week it is not a big deal. Don't change it unless we misspell this anywhere the is going to be permanent.

app/Filament/Clusters/ServiceManagementAdministration/Pages/ServiceRequestTypeTemplates.php
```diff
@@ -0,0 +1,208 @@
...
+            return false;
+        }
+
+        return (bool) auth()->user()?->can('settings.view-any');
+    }
+
+    public function form(Schema $schema): Schema
+    {
+        return $schema
+            ->components([
```
Orrison, line 96: This also needs a toggle to be able to change the preload setting, under the `use_custom_templates` one and only visible if `use_custom_templates` is true like everything else.

Orrison (changes requested):

app-modules/service-management/tests/Tenant/Filament/Clusters/ServiceManagementAdministration/Pages/ServiceRequestTypeTemplatesTest.php
```diff
@@ -0,0 +1,193 @@
...
+    get(ServiceRequestTypeTemplates::getUrl())
+        ->assertForbidden();
+});
+
+test('it allows access for a user with permission', function () {
+    asSuperAdmin();
+
+    get(ServiceRequestTypeTemplates::getUrl())
+        ->assertSuccessful();
+});
```
Orrison, line 75: Doesn't this file check if they have settings permissions? Should we be checking that instead of just making them a super admin?

app-modules/service-management/tests/Tenant/Filament/Clusters/ServiceManagementAdministration/Pages/ServiceRequestTypeTemplatesTest.php
```diff
@@ -0,0 +1,193 @@
...
+
+test('it saves the override flag and custom template content', function () {
```
Orrison, line 77: We may also want to be testing that it properly sets the preload setting

app-modules/service-management/tests/Tenant/Filament/Clusters/ServiceManagementAdministration/Pages/ServiceRequestTypeTemplatesTest.php
```diff
@@ -0,0 +1,193 @@
...
+    livewire(ServiceRequestTypeTemplates::class)
+        ->fillForm([
+            'use_custom_templates' => true,
+            'templates.' . ServiceRequestEmailTemplateType::Created->value . '.' . ServiceRequestTypeEmailTemplateRole::Customer->value . '.subject' => serviceRequestTypeTemplatesDoc('Hello'),
+            'templates.' . ServiceRequestEmailTemplateType::Created->value . '.' . ServiceRequestTypeEmailTemplateRole::Customer->value . '.body' => serviceRequestTypeTemplatesDoc('Body'),
+        ])
+        ->call('save')
+        ->assertHasNoFormErrors();
+
+    expect(app(ServiceRequestNotificationAutomationSettings::class)->use_custom_templates)->toBeTrue();
```
Orrison, line 91: To be safe, you should have an expect BEFORE you call the form and save that it is `false`

app/Filament/Clusters/ServiceManagementAdministration/Pages/ServiceRequestTypeTemplates.php
```diff
@@ -0,0 +1,217 @@
...
+                Toggle::make('preload_new_service_request_types')
```
Orrison: This should be swapped with the `use_custom_templates` below and only be visible if `use_custom_templates` is `true`.

The page should essentially be empty if the `use_custom_templates` toggle is off

Orrison (changes requested):

Orrison (approved):

PR #1335, opened 2026-08-03
[AIDAPP-1372]: Update base templates so they preload global administrator base templates on override toggle confirmation

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1372

### Technical Description

> Update base templates so they preload global administrator base templates on override toggle confirmation

> No

### Cleanup Tasks

> N/A

---


app-modules/service-management/src/Filament/Actions/PreloadBaseTemplatesAction.php
```diff
@@ -0,0 +1,137 @@
...
+
+class PreloadBaseTemplatesAction extends Action
+{
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this
+            ->modalHeading('Override Defaults')
+            ->modalDescription('Turning on the override will load the base templates set by the global administrator as the starting point, replacing any templates already entered below.')
```
Orrison: This message won't make sense if the override was already off and they never entered any templates nor do the see them in the settings below

app-modules/service-management/src/Filament/Actions/PreloadBaseTemplatesAction.php
```diff
@@ -0,0 +1,137 @@
...
+
+namespace AidingApp\ServiceManagement\Filament\Actions;
+
+use AidingApp\ServiceManagement\Enums\ServiceRequestEmailTemplateType;
+use AidingApp\ServiceManagement\Enums\ServiceRequestTypeEmailTemplateRole;
+use AidingApp\ServiceManagement\Models\ServiceRequestNotificationAutomationEmailTemplate;
+use Filament\Actions\Action;
+use Filament\Pages\SettingsPage;
+
+class PreloadBaseTemplatesAction extends Action
```
Orrison: I don't think we need this action or confirmation.

Let's just have it always override when it is toggled on on this page

Orrison (changes requested):

Orrison (approved):

PR #1345, opened 2026-08-05
[AIDAPP-1360]: Some users have reported that sorting in the knowledge base report is not working as expected

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1360

### Technical Description

- change default sorting to views in KnowledgeBaseArticlesTable

No

### Cleanup Tasks

No

---

### DIFF
--- a/app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesTable.php
+++ b/app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesTable.php
@@ -188,6 +188,10 @@ class KnowledgeBaseArticlesTable extends BaseWidget
             ->paginated([5, 10, 25])
             ->defaultPaginationPageOption(5)
             ->paginationPageOptions([5, 10, 25])
-            ->defaultSort('updated_at', 'desc');
+            ->defaultSort(function (Builder $query): Builder {
+                return $query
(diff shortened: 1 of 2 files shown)


app-modules/report/src/Filament/Widgets/KnowledgeBaseArticlesTable.php
```diff
@@ -189,6 +188,10 @@ public function table(Table $table): Table
             ->paginated([5, 10, 25])
             ->defaultPaginationPageOption(5)
             ->paginationPageOptions([5, 10, 25])
-            ->defaultSort('updated_at', 'desc');
+            ->defaultSort(function (Builder $query): Builder {
+                return $query
+                    ->orderBy('portal_view_count', 'desc')
+                    ->orderBy('updated_at', 'desc');
+            });
```
Orrison: The removal of the `orderBy` in the query above was very likely the culprit of why sorting toggles on the table were not working for testers.

So that removal alone is all that is needed.

We should revert this back just to `->defaultSort('updated_at', 'desc')`, imo

Orrison (changes requested):

Orrison (approved):

PR #1347, opened 2026-08-05
[AIDAPP-1375]: Introduce the ability to automate reporting from service monitors

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1375

### Technical Description

Add new fields, create pivot models, and form updates for automatic service monitoring reporting; add new notification classes; add jobs and tests

No

### Cleanup Tasks

.cleanup-tasks/2026_08_05_service_monitoring_report_feature.md

---

### DIFF
--- a/.cleanup-tasks/2026_08_05_service_monitoring_report_feature.md
+++ b/.cleanup-tasks/2026_08_05_service_monitoring_report_feature.md
@@ -1,0 +1,12 @@
+---
+title: Service Monitoring Report Feature
+created: 2026-08-05
+---
+
+## Feature Flags
+
+- App\Features\ServiceMonitoringReportFeature
+
+## Temporary Migrations
+
+## Additional Cleanup
--- a/app-modules/service-management/database/factories/ServiceMonitoringTargetFactory.php
+++ b/app-modules/service-management/database/factories/ServiceMonitoringTargetFactory.php
@@ -37,6 +37,7 @@
 namespace AidingApp\ServiceManagement\Database\Factories;
 
 use AidingApp\ServiceManagement\Enums\ServiceMonitoringFrequency;
+use AidingApp\ServiceManagement\Enums\ServiceMonitoringReportFrequency;
 use AidingApp\ServiceManagement\Models\ServiceMonitoringTarget;
 use Illuminate\Database\Eloquent\Factories\Factory;
 
@@ -59,6 +60,10 @@ class ServiceMonitoringTargetFactory extends Factory
             'frequency' => $this->faker->randomElement(ServiceMonitoringFrequency::cases()),
             'is_notified_via_database' => $this->faker->boolean(),
             'is_notified_via_email' => $this->faker->boolean(),
+            'is_reporting_active' => $this->faker->boolean(),
+            'report_frequency' => $this->faker->randomElement(ServiceMonitoringReportFrequency::cases()),
+            'is_reported_via_database' => $this->faker->boolean(),
+            'is_reported_via_email' => $this->faker->boolean(),
             'is_confidential' => false,
         ];
     }
--- a/app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 use Illuminate\Database\Migrations\Migration;
 use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
 use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
--- a/app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_user', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('user_id')->constrained()->cascadeOnDelete();
+
+            $table->timestamps();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('service_monitoring_target_report_user');
+    }
+};
--- a/app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 use Illuminate\Database\Migrations\Migration;
 use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
 use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
--- a/app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_department', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('department_id')->constrained()->cascadeOnDelete();
+
+            $table->timestamps();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('service_monitoring_target_report_department');
+    }
+};
--- a/app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php
@@ -1,5 +1,39 @@
 <?php
 
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
 use Illuminate\Database\Migrations\Migration;
 use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
 use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
--- a/app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php
+++ b/app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php
@@ -1,0 +1,24 @@
+<?php
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_contact', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('contact_id')->constrained()->cascadeOnDelete();
+
+            $table->timestamps();
+        });
+    }
+
+    public function down(): void
+    {
+        Schema::dropIfExists('service_monitoring_target_report_contact');
+    }
+};
--- a/app-modules/service-management/resources/views/mail/service-monitoring-report.blade.php
+++ b/app-modules/service-management/resources/views/mail/service-monitoring-report.blade.php
@@ -1,3 +1,36 @@
+{{--
+    <COPYRIGHT>
+    
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+    
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+    
+    Notice:
+    
+    - You may not provide the software to third parties as a hosted or managed
+    service, where the service provides users with access to any substantial set of
+    the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+    in the software, and you may not remove or obscure any functionality in the
+    software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+    of the licensor in the software. Any use of the licensor’s trademarks is subject
+    to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+    same in return. Canyon GBS® and Aiding App® are registered trademarks of
+    Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+    vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+    Software as a Service (SaaS) by Canyon GBS Inc.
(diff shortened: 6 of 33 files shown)


app-modules/service-management/src/Models/ServiceMonitoringTargetReportUser.php
```diff
@@ -0,0 +1,76 @@
...
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Models;
+
+use AidingApp\ServiceManagement\Database\Factories\ServiceMonitoringTargetReportUserFactory;
+use App\Models\User;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
```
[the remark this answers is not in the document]
rebecca-canyon: New models should not use legacy HasVersion4Uuids

app-modules/service-management/src/Models/ServiceMonitoringTargetReportDepartment.php
```diff
@@ -0,0 +1,76 @@
...
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Models;
+
+use AidingApp\Department\Models\Department;
+use AidingApp\ServiceManagement\Database\Factories\ServiceMonitoringTargetReportDepartmentFactory;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
```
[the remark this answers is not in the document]
rebecca-canyon: New models should not use legacy HasVersion4Uuids

app-modules/service-management/src/Models/ServiceMonitoringTargetReportContact.php
```diff
@@ -0,0 +1,76 @@
...
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Models;
+
+use AidingApp\Contact\Models\Contact;
+use AidingApp\ServiceManagement\Database\Factories\ServiceMonitoringTargetReportContactFactory;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
```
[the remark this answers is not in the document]
rebecca-canyon: New models should not use legacy HasVersion4Uuids

app-modules/service-management/database/factories/ServiceMonitoringTargetFactory.php
```diff
@@ -59,6 +60,10 @@ public function definition(): array
             'frequency' => $this->faker->randomElement(ServiceMonitoringFrequency::cases()),
             'is_notified_via_database' => $this->faker->boolean(),
             'is_notified_via_email' => $this->faker->boolean(),
+            'is_reporting_active' => $this->faker->boolean(),
+            'report_frequency' => $this->faker->randomElement(ServiceMonitoringReportFrequency::cases()),
+            'is_reported_via_database' => $this->faker->boolean(),
+            'is_reported_via_email' => $this->faker->boolean(),
```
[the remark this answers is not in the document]
rebecca-canyon: report_frequency and report channels could be enabled, and then is_reporting_active could be set to false, which would not clear out those values. So this creates possible states that this could be. Tests that need specific values here will control them.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function toDatabase(User $notifiable): array
+    {
+        return Notification::make()
+            ->title('Your ' . Str::lower($this->serviceMonitoringTarget->report_frequency->value ?? ServiceMonitoringReportFrequency::Monthly->value) . ' service monitor report for ' . $this->serviceMonitoringTarget->name . ' is ready.')
+            ->body(
```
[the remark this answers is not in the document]
rebecca-canyon: Unnecessary according to PHPStan.

app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php
```diff
@@ -0,0 +1,58 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_user', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('user_id')->constrained()->cascadeOnDelete();
```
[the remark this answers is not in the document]
rebecca-canyon: Matches current conventions for pivot tables

app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php
```diff
@@ -0,0 +1,58 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_contact', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('contact_id')->constrained()->cascadeOnDelete();
```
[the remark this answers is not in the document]
rebecca-canyon: Matches current conventions for pivot tables

app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php
```diff
@@ -0,0 +1,58 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::create('service_monitoring_target_report_department', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+
+            $table->foreignUuid('service_monitoring_target_id')->constrained()->cascadeOnDelete();
+            $table->foreignUuid('department_id')->constrained()->cascadeOnDelete();
```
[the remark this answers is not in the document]
rebecca-canyon: Matches current conventions for pivot tables

app-modules/service-management/src/Jobs/ServiceMonitoringReportJob.php
```diff
@@ -0,0 +1,85 @@
...
+        return 30 * 60;
+    }
+
+    public function handle(): void
+    {
+        if (! app(LicenseSettings::class)->data?->addons?->serviceMonitoring) {
+            return;
+        }
+
+        ServiceMonitoringTarget::where('report_frequency', $this->frequency)
```
payal-canyon, line 79: The dispatch query only checks `report_frequency`, not `is_reporting_active`. Since turning reporting off in the UI doesn't clear the stored frequency/channel values, the monitor can continue sending reports even after reporting is disabled.

Could you please add `->where('is_reporting_active', true)` to the query (and/or an early return in `ServiceMonitoringReportNotifyJob`)? It may also be worth clearing the frequency/channel values when reporting is disabled so the stored data stays consistent with the UI.

app-modules/service-management/src/Jobs/ServiceMonitoringReportJob.php
```diff
@@ -0,0 +1,85 @@
...
+     * Return the period for which this job should be unique for, half an hour, in seconds
+     */
+    public function uniqueFor(): int
+    {
+        return 30 * 60;
+    }
+
+    public function handle(): void
+    {
+        if (! app(LicenseSettings::class)->data?->addons?->serviceMonitoring) {
```
payal-canyon: I think we should add the feature flag check here as well. If it isn't active, we should return early.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,218 @@
...
+    {
+        return match ($this->channel) {
+            DatabaseChannel::class => ['database'],
+            MailChannel::class => ['mail'],
+            'both' => ['database', 'mail'],
+            default => throw new InvalidArgumentException("Unsupported notification channel: {$this->channel}"),
+        };
+    }
+
+    public function toMail(User $notifiable): MailMessage
```
payal-canyon: `toMail()` and `toDatabase()` are type-hinted as `User $notifiable`, but this notification can also be sent to `Contact` recipients (`via()` already supports `User|Contact`). Since the channels pass the actual notifiable into these methods, a `Contact` recipient would cause a `TypeError` and the notification would fail.

Could you please change both signatures to `object $notifiable`? This matches the pattern used by other notifications that support multiple recipient types and ensures contacts work as expected.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,218 @@
...
+            ->histories()
+            ->where('created_at', '>=', now()->subDays($this->getUptimeDays()))
+            ->where('succeeded', false)
+            ->get();
+
+        if ($serviceChecks->isEmpty()) {
+            return 'No incidents were detected during this reporting period.';
+        }
+
+        return $serviceChecks->count() . str('incident')->plural($serviceChecks->count()) . ' were detected during this reporting period.';
```
payal-canyon: I think we can simplify this a bit. Something like:
```javascript
private function getIncidentSummary(): string
{
    $incidentCount = $this->getFailedChecks();

    if ($incidentCount === 0) {
        return 'No incidents were detected during this reporting period.';
    }

    return $incidentCount . ' ' . str('incident')->plural($incidentCount) . ' ' . ($incidentCount === 1 ? 'was' : 'were') . ' detected during this reporting period.';
}
```

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,218 @@
...
+                'Failed checks: ' . $this->getFailedChecks() . "\n" .
+                'Incident summary: ' . $this->getIncidentSummary()
+            )
+            ->getDatabaseMessage();
+    }
+
+    /**
+     * @return array<string>
+     */
+    private function getReportPeriod(): array
```
payal-canyon, line 126: `getReportPeriod()` returns calendar boundaries in the tenant timezone, but every metric - `getSuccessfulChecks`, `getFailedChecks`, `getAverageResponseTime`, `getTotalDowntime`, `getIncidentSummary` and `getUptimePercentage($this->getUptimeDays())` - filters on a rolling `now()->subDays(N)` window in the app timezone. The reported numbers therefore don't match the period the email displays, and the gap widens with timezone differences, queue delays, and monthly reports.

Could you please have `getReportPeriod()` return the Carbon instances (formatted for display separately) and pass the same `[$start, $end]` into each metric via `whereBetween('created_at', [$start, $end])`? `getUptimePercentage()` needs a range-based version too, since a day count has no start anchor and can't express the calendar period - after which getUptimeDays() can go.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,218 @@
...
+
+        if ($averageResponseTime === null) {
+            return 'N/A';
+        }
+
+        return number_format((float) $averageResponseTime, 2) . ' s';
+    }
+
+    private function getTotalDowntime(): string
+    {
```
payal-canyon: Each metric currently runs its own query, and some (`getTotalDowntime()`, `getUptimePercentage()`, and `getIncidentSummary()`) hydrate the full result set just to calculate aggregates. `getUptimeDays()` is also re-evaluated multiple times, and `toMail()` runs twice per recipient.

For a monthly report with 20 recipients, this could result in hundreds of thousands of models being hydrated. Could you please compute the statistics once into a property using a single aggregate query with conditional counts? Combined with the period fix above, this should reduce the report to one query per notification and avoid repeating the same work for each recipient.

app-modules/service-management/src/Jobs/ServiceMonitoringReportNotifyJob.php
```diff
@@ -0,0 +1,98 @@
...
+    {
+        return 30 * 60;
+    }
+
+    public function handle(): void
+    {
+        $recipientUsers = $this->serviceMonitoringTarget->reportUsers()->get();
+        $recipientContacts = $this->serviceMonitoringTarget->reportContacts()->get();
+
+        $this->serviceMonitoringTarget->reportDepartments()->each(function (Department $department) use (&$recipientUsers) {
```
payal-canyon: This runs one query per department, and the same merge-and-dedupe logic already exists in `ServiceMonitoringCheckJob` -  over `users()` / `departments()` there rather than the report relations here.

Could you please move recipient resolution onto `ServiceMonitoringTarget` - e.g. `getReportRecipients()` alongside a `getNotificationRecipients()`, both sharing one private helper - and eager-load with `reportDepartments()->with('users')`? That removes the per-department queries and gives the two jobs a single implementation to maintain.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoring.php
```diff
@@ -97,6 +104,62 @@ public function form(Schema $schema): Schema
                             ->default(false),
                     ])
                     ->columns(2),
+                Section::make('Automated Reporting')
+                    ->schema([
```
payal-canyon: The automated reporting form is currently duplicated across both pages (~55 lines with six closures), with only a few small differences such as `->default(false)` and `afterStateHydrated()`.

Could you please extract the shared schema into an `AutomatedReportingSection::make(): Section` under `Resources/ServiceMonitorings/Schemas/Components/`, with the channels checkbox as its own component? Then each page could only apply its page-specific differences. This would also ensure fixes to things like toggle-off handling or channel validation only need to be made in one place.

payal-canyon (changes requested):

.cleanup-tasks/2026_08_05_service_monitoring_report_feature.md
```diff
@@ -0,0 +1,12 @@
...
+created: 2026-08-05
+---
+
+## Feature Flags
+
+- App\Features\ServiceMonitoringReportFeature
+
+## Temporary Migrations
+
+## Additional Cleanup
```
payal-canyon, line 12: Could you please add the filenames and a short note on what needs to be deleted from each file? You can use the other cleanup files as a reference.

app-modules/service-management/src/Jobs/ServiceMonitoringReportNotifyJob.php
rebecca-canyon: I've reworked the logic on both jobs to avoid the department queries. But I do think that, despite their similarity, they are distinct enough that a shared helper method would offer little benefit.

Creating `getNotificationRecipients` and `getReportRecipients` would just be moving the logic from one spot to another, so I don't see anything gained from that either.

But there might be something I'm not considering, too.

.cleanup-tasks/2026_08_05_service_monitoring_report_feature.md
rebecca-canyon: Because all of the usages of the feature flag are very direct- activating it in the migration and otherwise checking if it is active as a conditional- I do not believe that there is any benefit to describing that in the cleanup document. Searching for the feature name will correctly bring up all usages when it is time to clean this feature up, and there is no other cleanup or modification that will need to happen then.

app-modules/service-management/src/Jobs/ServiceMonitoringReportNotifyJob.php
payal-canyon: That makes sense. I checked both jobs and the rework looks right.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,223 @@
...
+     */
+    private function getStatistics(): array
+    {
+        if ($this->statistics !== null) {
+            return $this->statistics;
+        }
+
+        [$reportPeriodStart, $reportPeriodEnd] = $this->getReportPeriod();
+
+        $result = DB::table('historical_service_monitorings')
```
payal-canyon: The statistics query uses `DB::table('historical_service_monitorings')`, which bypasses the `SoftDeletes` scope applied by `$this->serviceMonitoringTarget->histories()`. This means soft-deleted checks are now included in the report, even though `ResetMonitoringAction::execute()` soft-deletes those histories.

Could you please build the query from `$this->serviceMonitoringTarget->histories()->selectRaw(...)` instead? That would automatically exclude soft-deleted records and also apply the `service_monitoring_target_id` filter from the relationship.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,223 @@
...
+
+    private function getIncidentSummary(): string
+    {
+        $incidentCount = $this->getStatistics()['failed_checks'];
+
+        if ($incidentCount === 0) {
+            return 'No incidents were detected during this reporting period.';
+        }
+
+        return $incidentCount . str('incident')->plural($incidentCount) . ($incidentCount === 1 ? ' was' : ' were') . ' detected during this reporting period.';
```
payal-canyon: The verb agreement is fixed, but the separator is still missing, so `$incidentCount . str('incident')->plural($incidentCount)` renders as **"3incidents were detected..."**.

Could you please add `. ' ' .` after `$incidentCount`? This also matches the spacing pattern used in the exporters.

app-modules/service-management/src/Filament/Components/ReportChannelCheckboxList.php
```diff
@@ -0,0 +1,75 @@
...
+            ->options([
+                'is_reported_via_email' => 'Email',
+                'is_reported_via_database' => 'Application',
+            ])
+            ->live()
+            ->afterStateUpdated(function (Set $set, array $state = []): void {
+                $set('is_reported_via_email', in_array('is_reported_via_email', $state, true));
+                $set('is_reported_via_database', in_array('is_reported_via_database', $state, true));
+            })
+            ->dehydrated(false)
```
payal-canyon: `->dehydrated(false)` is applied twice in the same builder chain. Could you please remove one of the duplicate calls? It's harmless, but removing it will keep the chain clean and avoid confusion.

app-modules/service-management/src/Filament/Components/AutomatedReportingSection.php
```diff
@@ -0,0 +1,97 @@
...
+use Filament\Forms\Components\Hidden;
+use Filament\Forms\Components\Radio;
+use Filament\Forms\Components\Select;
+use Filament\Forms\Components\Toggle;
+use Filament\Schemas\Components\Section;
+use Filament\Schemas\Components\Utilities\Get;
+
+class AutomatedReportingSection
+{
+    public static function make(bool $isCreatePage = false): Section
```
payal-canyon: The extraction looks good, but I think the `bool $isCreatePage` parameter can be removed. The three `->default($isCreatePage ? false : null)` calls can simply use `->default(false)`, since Edit already hydrates the existing record values.

The `afterStateHydrated` guard is still needed for Create, though. Could you please change the closure to accept `?ServiceMonitoringTarget $record = null` and use a `filled($record)` check?

That would let us remove `$isCreatePage` entirely and have both pages call `AutomatedReportingSection::make()` without arguments.

app-modules/service-management/src/Filament/Components/ReportChannelCheckboxList.php
```diff
@@ -0,0 +1,75 @@
...
+namespace AidingApp\ServiceManagement\Filament\Components;
+
+use AidingApp\ServiceManagement\Models\ServiceMonitoringTarget;
+use Filament\Forms\Components\CheckboxList;
+use Filament\Schemas\Components\Utilities\Get;
+use Filament\Schemas\Components\Utilities\Set;
+
+class ReportChannelCheckboxList
+{
+    public static function make(bool $isCreatePage = false): CheckboxList
```
payal-canyon: same as above.

app-modules/service-management/resources/views/mail/service-monitoring-report.blade.php
```diff
@@ -0,0 +1,74 @@
...
+Uptime: {{ $uptimePercentage }}<br>
+Successful Checks: {{ $successfulChecks }}<br>
+Failed Checks: {{ $failedChecks }}<br>
+Average Response Time: {{ $averageResponseTime }}<br>
+Total Downtime: {{ $totalDowntime }}
+
+## Incidents
+{{ $incidentSummary }}
+
+Last Checked: {{ is_null($serviceMonitoringTarget->latestHistory) ? 'N/A' : $serviceMonitoringTarget->latestHistory->created_at->format('M j, Y g:i a (T)') }}
```
payal-canyon: The email template currently ends after `Last Checked:` and goes straight to the footer, but the spec includes one more line explaining why the recipient is receiving the report.

Could you please add this below `Last Checked`?

> You are receiving this email because you are subscribed to {{ REPORT_FREQUENCY | LOWERCASE }} reports for this service monitor.

Please use the same `strtolower($serviceMonitoringTarget->report_frequency?->value ?? 'monthly')` expression already used in the template so the wording stays consistent.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,223 @@
...
+    {
+        if ($this->statistics !== null) {
+            return $this->statistics;
+        }
+
+        [$reportPeriodStart, $reportPeriodEnd] = $this->getReportPeriod();
+
+        $result = DB::table('historical_service_monitorings')
+            ->where('service_monitoring_target_id', $this->serviceMonitoringTarget->id)
+            ->whereBetween('created_at', [$reportPeriodStart, $reportPeriodEnd])
```
payal-canyon: `getReportPeriod()` builds the boundaries in the tenant's timezone, but `created_at` stores UTC values. Passing those timezone-aware Carbon instances directly to `whereBetween()` means Laravel binds the local time as-is, so the query can cover the wrong UTC range for non-UTC tenants.

Could you please pass `$start->copy()->utc()` and `$end->copy()->utc()` to `whereBetween()`, while keeping the tenant-timezone values for display? It's a small change and would ensure the report covers the correct period for non-UTC tenants.

app-modules/service-management/resources/views/mail/service-monitoring-report.blade.php
payal-canyon: `latestHistory->created_at` is formatted in the app timezone, while the Reporting Period is shown in the tenant timezone. This means the email can display two different timezones and make the last check appear outside the reported period.

Could you please format `latestHistory->created_at` using the instance timezone as well? Since `getReportPeriod()` already resolves the timezone, the cleanest approach would be to pass it to the view and use it when formatting `Last Checked`, rather than resolving `Tenant::current()` in Blade.

payal-canyon (changes requested):

app-modules/service-management/src/Filament/Components/ReportChannelCheckboxList.php
payal-canyon: I think this suggestion may have been missed. Could you please take a look and address it?

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,226 @@
...
+
+    private function getIncidentSummary(): string
+    {
+        $incidentCount = $this->getStatistics()['failed_checks'];
+
+        if ($incidentCount === 0) {
+            return 'No incidents were detected during this reporting period.';
+        }
+
+        return $incidentCount . str('incident')->plural($incidentCount) . ' ' . ($incidentCount === 1 ? ' was' : ' were') . ' detected during this reporting period.';
```
payal-canyon: Could you please recheck the spacing here? It currently renders as `3incidents  were detected` - there’s no space after the count and an extra space before `were`.

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -0,0 +1,226 @@
...
+use Carbon\CarbonInterface;
+use Filament\Notifications\Notification;
+use Illuminate\Bus\Queueable;
+use Illuminate\Contracts\Queue\ShouldQueue;
+use Illuminate\Notifications\Notification as BaseNotification;
+use Illuminate\Support\Str;
+use InvalidArgumentException;
+
+class ServiceMonitoringReportNotification extends BaseNotification implements ShouldQueue
+{
```
payal-canyon, line 56: Could you please add test coverage for the reporting feature?

- Add `ServiceMonitoringReportNotificationTest` following the existing notification test pattern, covering the report logic and edge cases.
- Extend `CreateServiceMonitoringTest` / `EditServiceMonitoringTest` with a `report_frequency` required case in the existing validation dataset.
- Add coverage for `ReportChannelCheckboxList::afterStateHydrated()` using `assertSchemaStateSet`. 
- Extend The correct details are displayed on the `ViewServiceMonitoring` page to cover the Automated Reporting section, and add a case verifying it is hidden when `is_reporting_active` is false.
- If possible, add a period-boundary test using a non-UTC `DisplaySettings::$timezonee`

payal-canyon (changes requested):

payal-canyon (changes requested):
Apologies, but it looks like there are currently some merge conflicts. Could you please resolve them and then re-request my review? Thanks!

Orrison (dismissed):
@rebecca-canyon given that @payal-canyon is done with her shift, if you get this ready for re-review please request my re-review AND Payal's.

That way if I am able to get it in, I can. If not @payal-canyon can pick it up in the morning

payal-canyon (approved):

PR #1348, opened 2026-08-05
[AIDAPP-1374]: Introduce the ability to configure confidential service monitors

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1374

### Technical Description

> Introduce the ability to configure confidential service monitors

> NO

### Cleanup Tasks

>N/A

---


app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoring.php
```diff
@@ -97,6 +99,46 @@ public function form(Schema $schema): Schema
                             ->default(false),
                     ])
                     ->columns(2),
+                Section::make('Confidentiality')
```
payal-canyon: Could you please create a feature flag and make sure the feature is properly gated behind it?

app-modules/service-management/src/Observers/ServiceMonitoringTargetObserver.php
```diff
@@ -0,0 +1,60 @@
...
+use AidingApp\ServiceManagement\Models\ServiceMonitoringTarget;
+
+class ServiceMonitoringTargetObserver
+{
+    public function creating(ServiceMonitoringTarget $serviceMonitoringTarget): void
+    {
+        $this->assignCreatedByIfBlank($serviceMonitoringTarget);
+    }
+
+    public function updating(ServiceMonitoringTarget $serviceMonitoringTarget): void
```
payal-canyon, line 53: The backfill currently runs on every update, not just when confidentiality is enabled.

This means any non-admin editing a legacy monitor for an unrelated change can become its permanent creator and retain access even after being removed from the grant lists. It also creates an audit entry on the first edit of every existing record.

Could you please narrow this to when confidentiality is actually enabled?
```php
if ($serviceMonitoringTarget->isDirty('is_confidential') && $serviceMonitoringTarget->is_confidential)
```

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoring.php
```diff
@@ -97,6 +99,46 @@ public function form(Schema $schema): Schema
...
+                    ->schema([
```
payal-canyon: Could you please extract this into `Resources/ServiceMonitorings/Schemas/Components/ConfidentialitySection.php` with a `public static function make(): Section` ? The section is multi-line and shared across pages, so having it as a reusable component would keep the schema consistent and easier to maintain.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/EditServiceMonitoring.php
```diff
@@ -102,6 +104,46 @@ public function form(Schema $schema): Schema
                             ->default(false),
                     ])
                     ->columns(2),
+                Section::make('Confidentiality')
```
payal-canyon: same as above.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/EditServiceMonitoring.php
```diff
@@ -102,6 +104,46 @@ public function form(Schema $schema): Schema
...
+                            ->afterStateUpdated(function (bool $state, Set $set): void {
+                                if ($state) {
+                                    return;
+                                }
+
+                                $set('confidentialUsers', []);
+                                $set('confidentialDepartments', []);
+                                $set('confidentialContacts', []);
+                            }),
+                        Select::make('confidentialUsers')
```
payal-canyon: same as above

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoring.php
```diff
@@ -97,6 +99,46 @@ public function form(Schema $schema): Schema
...
+                            ->afterStateUpdated(function (bool $state, Set $set): void {
+                                if ($state) {
+                                    return;
+                                }
+
+                                $set('confidentialUsers', []);
+                                $set('confidentialDepartments', []);
+                                $set('confidentialContacts', []);
+                            }),
+                        Select::make('confidentialUsers')
```
payal-canyon: Could you please use `UserSelect` here instead of a plain `Select`? The same form already uses `UserSelect::make('user')`, which searches by name/email and applies `WithoutAnyAdmin`.

That admin filtering is important here since admins already have access to all confidential monitors, so listing them as grantees would provide options that don't actually change access.

app-modules/service-management/database/factories/ServiceMonitoringTargetFactory.php
```diff
@@ -59,6 +59,14 @@ public function definition(): array
...
             'is_notified_via_database' => $this->faker->boolean(),
             'is_notified_via_email' => $this->faker->boolean(),
+            'is_confidential' => false,
         ];
     }
+
+    public function confidential(): static
+    {
+        return $this->state(fn (array $attributes) => [
+            'is_confidential' => true,
```
payal-canyon, line 69: Could you please switch the six factory call sites to `ServiceMonitoringTarget::factory()->confidential()->create(...)`, keeping any other attributes in the `create()` array? The `confidential()` state is currently unused, while these tests manually set `is_confidential => true`.

The three `fillForm()` cases can stay as-is since those are testing the form input, not factory setup.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/ListServiceMonitorings.php
```diff
@@ -57,7 +60,11 @@ public function table(Table $table): Table
                 TextColumn::make('name')
                     ->label('Name')
                     ->searchable()
-                    ->sortable(),
+                    ->sortable()
+                    ->icon(fn (ServiceMonitoringTarget $record): ?Heroicon => $record->is_confidential ? Heroicon::LockClosed : null)
+                    ->iconColor('gray')
+                    ->iconPosition(IconPosition::After)
+                    ->tooltip(fn (ServiceMonitoringTarget $record): ?string => $record->is_confidential ? 'This service monitor is confidential and only visible to admins, its creator, and permitted users, departments, and contacts.' : null),
```
payal-canyon: Could you please add assertions for the lock `icon()` and `tooltip()`? Currently, nothing verifies that these are configured correctly.

app-modules/service-management/src/Models/ServiceMonitoringTarget.php
```diff
@@ -140,4 +192,16 @@ public function getUptimePercentage(int $days): string
...
+
+    /**
+     * @return array<string, string>
+     */
+    protected function casts(): array
+    {
+        return [
+            'is_notified_via_database' => 'boolean',
+            'is_notified_via_email' => 'boolean',
+            'is_confidential' => 'boolean',
```
payal-canyon, line 201: This adds a `casts()` method alongside the existing `$casts` property. They merge correctly, so there’s no bug, Could you please move `frequency` into `casts()` and remove the `$casts` property?

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoringTest.php
```diff
@@ -227,3 +227,26 @@
...
+        ->fillForm([
+            ...ServiceMonitoringTargetRequestFactory::new()->create(),
+            'is_confidential' => true,
+            'confidentialUsers' => [$grantedUser->getKey()],
+            'confidentialDepartments' => [$grantedDepartment->getKey()],
+        ])
+        ->call('create')
+        ->assertHasNoFormErrors();
+
+    $serviceMonitoringTarget = ServiceMonitoringTarget::latest('created_at')->first();
```
payal-canyon: Could you please avoid ServiceMonitoringTarget::latest('created_at')->first() here? It’s order-dependent and doesn’t follow the project’s query convention. Could you use `assertDatabaseHas(...)`

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/EditServiceMonitoring.php
```diff
@@ -128,6 +170,21 @@ public function getBreadcrumbs(): array
...
+        $record = $this->getRecord();
+
+        if ($record->is_confidential) {
+            return;
+        }
+
+        // Filament skips saving hidden relationship fields, so clear stale grants explicitly
+        $record->confidentialUsers()->sync([]);
+        $record->confidentialDepartments()->sync([]);
+        $record->confidentialContacts()->sync([]);
```
payal-canyon, line 154: Could you please guard these `sync([])` calls with `if (! $record->wasChanged('is_confidential')) { return; }`? Otherwise, all three are executed on every save of a non-confidential monitor.

payal-canyon (changes requested):
In ServiceMonitoringNotification Down-alerts can fail for confidential monitors.

Because the notification is queued, `loadMissing('serviceMonitoringTarget')` runs without an authenticated user. The new global scope then filters out confidential targets, so `$target` becomes `null` and `$target->name` throws. The same issue affects `->domain` in the mail template.

Could you please load the target without `ServiceMonitoringTargetVisibilityScope`, similar to the fix in `ServiceMonitoringJob`?

```php
$this->historicalServiceMonitoring->loadMissing([
    'serviceMonitoringTarget' => fn ($query) => $query->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class),
]);
```

Also, could you add a `ServiceMonitoringCheckJobTest` case for a confidential target with a recipient and assert that the notification is sent?

__________________________________________________________________________________
In ServiceMonitorTable and ServiceMonitoringTargetExporter CSV export can silently drop confidential rows, even for a super admin.

`total_rows` is counted while the admin is authenticated, but `PrepareCsvExport` re-runs the serialized query in the queue without a user, so the visibility scope filters out confidential rows. This can result in a shorter CSV while the completion notice still reports the full count.

Could you please add `->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class)` in the exporter's `modifyQuery()`?

app-modules/report/src/Filament/Exports/ServiceMonitoringTargetExporter.php
```diff
@@ -36,15 +36,28 @@
...
+    /**
+     * @param  Builder<ServiceMonitoringTarget>  $query
+     *
+     * @return Builder<ServiceMonitoringTarget>
+     */
+    public static function modifyQuery(Builder $query): Builder
+    {
+        // The export re-runs this query in a queued job with no authenticated user, so the confidentiality scope must be bypassed
+        return $query->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class);
+    }
```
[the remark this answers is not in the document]
payal-canyon: I think I was wrong here, and Copilot was right. Apologies for the confusion.

Could we please remove `modifyQuery()` and the related exporter test? The current tests still pass without it because they test the model's global scope rather than the exporter's `modifyQuery()`, so they don't actually cover this method.

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoringTest.php
```diff
@@ -227,3 +227,31 @@
             return ! empty($field->getSearchResults($adminUser->name));
         });
 });
+
+test('creating a confidential service monitor persists the granted users, departments, and contacts', function () {
```
payal-canyon, line 232: Could we also add test coverage for `confidentialContacts` in the admin panel? The current Create test only covers users and departments, and the Edit clearing test only asserts `confidentialUsers`. A regression in the contacts or departments sync could therefore still go unnoticed.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ConfidentialitySection.php
```diff
@@ -0,0 +1,90 @@
...
+use Filament\Forms\Components\Toggle;
+use Filament\Schemas\Components\Section;
+use Filament\Schemas\Components\Utilities\Get;
+use Filament\Schemas\Components\Utilities\Set;
+
+class ConfidentialitySection
+{
+    public static function make(): Section
+    {
+        return Section::make('Confidentiality')
```
payal-canyon, line 55: Could we also add a test for the `ConfidentialitySection` visibility behavior? Currently, there’s no coverage to verify that the user, department, and contact selects are hidden when the confidentiality toggle is off.

Since this is now a shared component, one `assertFormFieldHidden` test should be enough to cover the behavior.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1350, opened 2026-08-06
[AIDAPP-1303]: Introduce support for automated status changes during assignment.

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1303

### Technical Description

Adds per–Service Request Type support for automatically changing a ticket's status when it is auto-assigned via the automated assignment system.

On a Service Request Type's **Assignments** admin page, when an automated assignment type is selected, an optional **Automated Status Change** toggle appears. When enabled, a required status picker is shown (the same `ServiceRequestStatusSelect` used when managing service requests, grouped by classification), defaulting to the classification's default "new" status. When a ticket is subsequently auto-assigned (via the Individual, Round Robin, or Workload assignment types), its status is set to the configured status.

Key decisions:

- **Single nullable FK, no separate boolean.** The feature adds one nullable `automated_status_id` FK to `service_request_types` (nullable, `nullOnDelete`). Enabled state is derived purely from whether the FK is set — set = on, `null` = off — rather than persisting a separate boolean flag. The Filament **Automated Status Change** toggle is a form-only field (not persisted); `mutateFormDataBeforeFill` derives it from `filled(automated_status_id)` and `mutateFormDataBeforeSave` nulls the FK when the toggle is off and strips the toggle key.
- **Runtime hook.** The status change is applied by the auto-assigners themselves — not an observer. `ServiceRequestTypeAssigner` is an abstract base class whose `final execute()` resolves the assignee (via each subclass's `resolveAssignee()`) and then, in one place, creates the assignment and applies the status change. This makes the status change fire **only** during the automated-assignment flow (Individual/Round Robin/Workload) and impossible for a future assigner to bypass, while never triggering for other system-created assignments (seeder, import, manual reclassify override, or any direct `assignments()->create()`). It is guarded by the feature flag, resolves the type via the request's priority, and short-circuits when the ticket is already at the target status — avoiding a redundant update and any extra notification churn.
- **Zero-downtime.** All new behaviour is gated behind the `AutomatedStatusChangeOnAssignmentFeature` Pennant flag, which is activated inside the migration that adds the column.

No.

### Cleanup Tasks

Yes, Feature Flag: `AutomatedStatusChangeOnAssignmentFeature` — `.cleanup-tasks/2026_08_06_automated_status_change_on_assignment_feature.md`

---


app-modules/service-management/database/migrations/2026_08_06_121719_add_automated_status_change_columns_to_service_request_types_table.php
```diff
@@ -0,0 +1,69 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::table('service_request_types', function (Blueprint $table) {
+                if (! Schema::hasColumn('service_request_types', 'automated_status_id')) {
```
Orrison: We should never have conditionals like this in migrations. Our databases are fully controlled by these migrations so if we are ever being surprised that something already exists, something went wrong.

You also don't want a migrations to NOT do something you expect.

So let's just remove the conditional here.

app-modules/service-management/database/migrations/2026_08_06_121719_add_automated_status_change_columns_to_service_request_types_table.php
```diff
@@ -0,0 +1,69 @@
...
+        });
+    }
+
+    public function down(): void
+    {
+        DB::transaction(function () {
+            AutomatedStatusChangeOnAssignmentFeature::deactivate();
+
+            Schema::table('service_request_types', function (Blueprint $table) {
+                if (Schema::hasColumn('service_request_types', 'automated_status_id')) {
```
Orrison: Same as the above comment

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/EditServiceRequestTypeAssignments.php
```diff
@@ -121,6 +131,24 @@ public function form(Schema $schema): Schema
...
+                            ->columnSpanFull()
+                            ->live()
+                            ->visible(fn (Get $get): bool => AutomatedStatusChangeOnAssignmentFeature::active()
+                                && filled($get('assignment_type'))
+                                && $get('assignment_type') !== ServiceRequestTypeAssignmentTypes::None)
+                            ->afterStateUpdated(function (bool $state, Get $get, Set $set) use ($defaultStatusId): void {
+                                if ($state && blank($get('automated_status_id'))) {
+                                    $set('automated_status_id', $defaultStatusId);
+                                }
+                            }),
```
Orrison, line 145: I think in this scenario it would actually be preferred in this situation that we don't load the default in, so that they are required to think about it and select one.

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
```diff
@@ -153,10 +154,37 @@ public function created(ServiceRequestAssignment $serviceRequestAssignment): voi
...
+
+    private function applyAutomatedStatusChange(ServiceRequestAssignment $serviceRequestAssignment): void
+    {
+        if (! AutomatedStatusChangeOnAssignmentFeature::active()) {
+            return;
+        }
+
+        if (filled($serviceRequestAssignment->assigned_by_id)) {
+            return;
+        }
```
Orrison: Imo having this in this observer and having it trigger mostly just based on whether or not `assigned_by_id` is empty is a little brittle.

Currently the auto assigner for is the primary way that assignment can happen when there isn't a logged in user. But it isn't the only way now and definitely won't be the only way in the future.

The database seeder (ServiceRequestSeeder) creates assignments via factory with no assigned_by_id.
Reclassification flows create assignments — the ReclassifyServiceRequestActionTest shows assignments created with 'assigned_by_id' => null even when a super admin is doing the reclassify manually.
Any future programmatic assignment (imports, migrations, API) that doesn't populate assigned_by_id would silently trigger a status change.

Auto-assignment is triggered from two places, both delegating to [AssignServiceRequestToDepartment.php:46](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html):

[CreateServiceRequestAction.php:46-62](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — when a new service request is created, assignServiceRequestToDepartment->execute($serviceRequest) runs inside the same DB transaction as the creation, right before manager/auditor notifications are sent.
[SubmitServiceRequestDraft.php:54-63](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) — when a draft SR is submitted, assignToDepartment($draft) calls the same action (unless $assignToDepartment is passed as false).
Additionally, [ReclassifyServiceRequestAction.php:156-181](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) re-triggers assignment when a manager changes an SR's priority/type — it clears existing active assignments, then either manually assigns (override) or calls $record->priority->type->assignment_type->getAssignerClass()?->execute($record) directly.

So we have a bit of a disjointed problem here I think we need to solve and I am open to suggestions.
We need this feature to reliably only be triggering during the auto-assignment flow, there is a bit of a disjointed issue there because the auto assignment flow is broken up between different auto assigner classes that implement `ServiceRequestTypeAssigner` so there could be more options but my first thoughts is finding some way to make it so there is some sort of central point for them all.

Orrison (changes requested):

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/EditServiceRequestTypeAssignments.php
Orrison: Never mind this was asked for

app-modules/service-management/database/migrations/2026_08_06_121719_add_automated_status_change_columns_to_service_request_types_table.php
mxm1070: Removed the existence conditionals — the migration now adds the columns unguarded.

app-modules/service-management/database/migrations/2026_08_06_121719_add_automated_status_change_columns_to_service_request_types_table.php
mxm1070: Removed here too — same fix as above.

app-modules/service-management/src/Observers/ServiceRequestAssignmentObserver.php
mxm1070: Agreed on the concern, and we went with the abstract-class approach you preferred. The trigger is out of this observer entirely — the `assigned_by_id` heuristic and the feature check here are both gone.

**What changed structurally:**

- The `ServiceRequestTypeAssigner` interface is now an **abstract base class**. It owns the assignment creation and the automated status change in a `private assign(ServiceRequest, User)` method (previously the triplicated `create([...])` block in each assigner).
- `execute()` is now **`final`** on the base and follows a template-method shape: it resolves the assignee and, if one is found, calls `assign()` itself:

  ```php
  final public function execute(ServiceRequest $serviceRequest): void
  {
      $user = $this->resolveAssignee($serviceRequest);

      if ($user !== null) {
          $this->assign($serviceRequest, $user);
      }
  }

  abstract protected function resolveAssignee(ServiceRequest $serviceRequest): ?User;
  ```

- The three assigners (`Individual`/`RoundRobin`/`Workload`) no longer implement `execute()`. They only implement `resolveAssignee()` — deciding *who* to assign (Individual returns the configured manager; RoundRobin/Workload do their selection + `last_assigned_id` bookkeeping and return the chosen user, or `null`).

This gives us the **central point** you were after: assignment (and the status change) happen in exactly one place, the base class, funnelled through `getAssignerClass()?->execute()` for all three call sites (create, submit-draft, reclassify).

**On the "future implementer forgets to call it" worry** — that's now impossible by construction. A new assigner can't forget to call `assign()` because it never calls it; the `final execute()` does. `execute()` can't be overridden to bypass it either. Subclasses only pick the user. And every non-auto-assignment path (seeder, import, reclassify manual override, any direct `assignments()->create()`) never runs an assigner, so it never triggers the status change — which is exactly the reliability guarantee we needed.

**Why not a PHPStan rule or Pest arch test** (both considered, both discarded):

- *Pest arch* can only assert class structure (extends/implements/traits/naming/dependencies) — it can't inspect a method body to verify that `execute()` actually calls `assign()`. So it can't express this rule at all.
- A *custom PHPStan rule* could walk each subclass's AST looking for a `$this->assign(...)` call, but it's brittle: it only proves the call *appears* syntactically, not that it's reachable (someone could wrap it in `if (false)`), and it'd be the only hand-written rule in the repo (everything else comes from `orrison/meliorstan`), so it cuts against convention and adds maintenance.

The restructure enforces the same guarantee through the type system with no rule to maintain, so we went with that instead.

app-modules/service-management/tests/Tenant/Models/ServiceRequestTypeTest.php
Orrison, line 1: I am not sure if I find that either of these tests really provide value.

I would ask that we remove them.

app-modules/service-management/tests/Tenant/Observers/ServiceRequestAssignmentObserverTest.php
```diff
@@ -665,3 +667,29 @@ function enablePreference(
         expect($assignedManagerEmails->first()->emailTemplate?->is($managerTemplate))->toBeTrue();
     });
 });
+
+describe('automated status change on assignment', function () {
+    it('does not change the request status when an assignment is created outside the auto-assignment flow', function () {
```
Orrison: This is a type of test that I think AI is notorious about writing and stems from when you work in a thread where you did one thing, then went back on it.

I don't think we really NEED to be testing here that a status change doesn't happen from normal assignment.
That leads down a bit of a slippery slope imo.

app-modules/service-management/tests/Tenant/Services/ServiceRequestType/IndividualAssignerTest.php
```diff
@@ -36,17 +36,72 @@
...
+    ): void {
+        ServiceRequestTypeEmailPreference::create([
+            'service_request_type_id' => $type->getKey(),
+            'service_request_email_template_type' => $templateType,
+            'service_request_email_template_role' => $role,
+            'notification_channel' => $channel,
+            'is_enabled' => $isEnabled,
+        ]);
+    }
+}
```
Orrison: If we are defining these like this i.e with the `! function_exists('enablePreference')` checks because we define them the same in other test files. I think we shouldn't do this and these should be moved to a shared pest helper file.

https://pestphp.com/docs/custom-helpers

Orrison (changes requested):

app-modules/service-management/tests/Tenant/Models/ServiceRequestTypeTest.php
mxm1070: Removed `ServiceRequestTypeTest.php` — both tests were low value (a plain belongsTo/FK persistence check, and asserting the feature flag is active, which it is by default in the suite since migrations run).

app-modules/service-management/tests/Tenant/Observers/ServiceRequestAssignmentObserverTest.php
mxm1070: Removed the `automated status change on assignment` describe block. Agreed — with the trigger moved out of the observer entirely (into the assigner base class), asserting the observer *doesn't* change status was guarding absence in the wrong place.

app-modules/service-management/tests/Tenant/Services/ServiceRequestType/IndividualAssignerTest.php
mxm1070: Centralized `enablePreference` into the shared `tests/Helpers.php` (Pest auto-loads it via BootFiles), imported where used with `use function Tests\enablePreference;`. That single source removes the need for the `function_exists` guard entirely. Left the two single-use helpers (`individualAutoAssignType`/`serviceRequestForType`) in place since they're local to this one file.

app-modules/service-management/src/Services/ServiceRequestType/ServiceRequestTypeAssigner.php
```diff
@@ -36,9 +36,57 @@
...
+                'user_id' => $user->getKey(),
+                'assigned_by_id' => null,
+                'assigned_by_type' => null,
+                'assigned_at' => now(),
+                'status' => ServiceRequestAssignmentStatus::Active,
+            ]);
+
+            $this->applyAutomatedStatusChange($serviceRequest);
+        });
+    }
```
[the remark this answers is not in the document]
mxm1070: Good catch, and technically accurate — `resolveAssignee()` in the round-robin/workload assigners commits `last_assigned_id` before `assign()` opens its transaction, so an `assign()` failure would advance the rotation pointer without recording an assignment.

That said, this is pre-existing behavior, not something this PR introduces: on the base branch (`release/1.42.1`) the pointer save and the assignment create were already separate un-transactioned writes, so an assignment-create failure left the pointer advanced there too. The transaction added here only narrows the assignment-vs-status gap; it does not open this one.

Wrapping `resolveAssignee()` in the transaction so the rotation pointer only advances on full success is a deliberate change to round-robin/workload ordering semantics, which is outside the scope of this ticket (automated status changes). The impact is also low — the failure path is rare and the effect is skipping one person in the rotation once, which self-heals. We may still fix it, but as a separate follow-up rather than expanding scope here.

app-modules/service-management/tests/Tenant/Services/ServiceRequestType/IndividualAssignerTest.php
```diff
@@ -36,16 +36,53 @@
...
+use AidingApp\ServiceManagement\Notifications\ServiceRequestStatusChanged;
 use AidingApp\ServiceManagement\Services\ServiceRequestType\IndividualAssigner;
+use App\Features\AutomatedStatusChangeOnAssignmentFeature;
 use App\Models\User;
+use Illuminate\Support\Facades\Notification;
 
 use function Tests\asSuperAdmin;
+use function Tests\enablePreference;
+
+if (! function_exists('individualAutoAssignType')) {
```
Orrison: I am fine with having methods local to the files if they are only used in this one file. But if we are going to do that, let's NOT have the `! function_exists` checks. That way we can catch if we start defining this in other places for some reason

Orrison (changes requested):

app-modules/service-management/tests/Tenant/Services/ServiceRequestType/IndividualAssignerTest.php
mxm1070: Done — dropped the `! function_exists` guards on `individualAutoAssignType` and `serviceRequestForType`. Confirmed both are used only in this file, so they stay local; without the guards we'll now get a redeclaration error if either name starts being defined elsewhere.

Orrison (approved):

PR #1352, opened 2026-08-10
[AIDAPP-1300]: Enhance service request by allowing configuration of default priority levels per request type

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1300

### Technical Description

> Enhance service request by allowing configuration of default priority levels per request type

> No

### Cleanup Tasks

> Yes, FeatureFlag: `DefaultPriorityFeature`

---


Orrison (changes requested):
@amit-canyon please rebase this on and change the PR base to `release/1.43.1`

app-modules/ai/src/Http/Controllers/AssistantWidget/StoreServiceRequestController.php
```diff
@@ -80,7 +81,10 @@ public function __invoke(Request $request, ServiceRequestType $type): JsonRespon
         $data = $request->validate([
             'title' => ['required', 'string', 'max:255'],
             'description' => ['required', 'string'],
-            'priority_id' => ['required', 'uuid'],
+            'priority_id' => [
+                DefaultPriorityFeature::active() && $type->defaultPriority()->exists() ? 'nullable' : 'required',
+                'uuid',
+            ],
```
Orrison, line 89: Hmm, I think this should actually work via the [prohibited_if](https://laravel.com/docs/13.x/validation#rule-prohibited-if) closure version.

It's not a default for them to select and see on the front end, it is the priority it MUST be.

So the user would not see it or select it on the frontend, if they somehow found a way to try and submit the form including a priority_id we would reject that request. And in here after that validation if there is a default set, then we set it here.

It's intended to make it so the end customer is not the one to select the priority, but instead it is defined by this setting.

app-modules/service-management/database/migrations/2026_08_07_103227_add_default_priority_id_service_request_types_table.php
```diff
@@ -0,0 +1,59 @@
...
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('service_request_types', function (Blueprint $table) {
+            $table->foreignUuid('default_priority_id')
+                ->nullable()
+                ->constrained('service_request_priorities')
+                ->nullOnDelete();
```
Orrison: Let's remove the `nullOnDelete` here and make it so it will error if they try and delete a used one. Then we should add archiving to `ServiceRequestPriority` soon (not in this increment)

app-modules/service-management/database/migrations/2026_08_07_164555_data_activate_default_priority_feature.php
Orrison, line 1: This could be and would be better to be, wrapped into the other migration

Orrison (changes requested):

app-modules/ai/tests/Tenant/Http/Controllers/AssistantWidget/StoreServiceRequestControllerTest.php
Orrison, line 1: I think we need a test here for it being required and that it is used when no default is set

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestController.php
```diff
@@ -88,7 +89,9 @@ public function __invoke(Request $request, ServiceRequestType $type): JsonRespon
             ], Response::HTTP_UNPROCESSABLE_ENTITY);
         }
 
-        $priority = $type->priorities()->findOrFail($data->pull('Main.priority'));
+        $priority = DefaultPriorityFeature::active() && $type->defaultPriority()->exists()
+            ? $type->defaultPriority()->first()
+            : $type->priorities()->findOrFail($data->pull('Main.priority'));
```
Orrison, line 95: Is there anything in this flow that is doing the required / prohibited validation that the similar flow for the assistant is doing for priority?

Orrison (changes requested):

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/GetServiceRequestFormControllerTest.php
```diff
@@ -49,6 +51,41 @@
...
+    try {
+        $response = getJson(route('api.portal.service-request.create', ['type' => $type]));
+
+        $response->assertOk();
+
+        expect(json_encode($response->json('schema')))->toContain('priority');
+    } finally {
+        DefaultPriorityFeature::activate();
+    }
+});
```
Orrison, line 97: I think we need a test here that shows that (assuming the FF is active because the migration activated it and it will be deleted eventually) when there is NOT a default priority set, then priority is passed.

Or is this already covered by a different test here?

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestControllerTest.php
```diff
@@ -0,0 +1,101 @@
...
+            route('api.portal.service-request.store', ['type' => $type]),
+            ['Main' => ['title' => 'Portal priority test', 'description' => 'A priority is required.']],
+        );
+
+        $response->assertUnprocessable()
+            ->assertJsonValidationErrors('Main.priority');
+    } finally {
+        DefaultPriorityFeature::activate();
+    }
+});
```
Orrison, line 123: Same as my other comment above, it is required if there is no default, right? So we should have a test for that assuming the ff is active.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypePriorities.php
```diff
@@ -124,4 +157,85 @@ public function table(Table $table): Table
...
+        assert($form instanceof Schema);
+
+        return $form;
+    }
+
+    protected function getServiceRequestType(): ServiceRequestType
+    {
+        $type = $this->getOwnerRecord();
+
+        assert($type instanceof ServiceRequestType);
```
Orrison: I would actually do an if here and have it throw if it isn't. Since this is something we can control and it IS possible to put this manager in the wrong spot, an actual exception would make more sense here.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypePriorities.php
```diff
@@ -124,4 +157,85 @@ public function table(Table $table): Table
...
+            ->title('Configuration saved')
+            ->success()
+            ->send();
+    }
+
+    protected function getConfigurationForm(): Schema
+    {
+        $form = $this->getSchema('configurationForm');
+
+        assert($form instanceof Schema);
```
Orrison: Same as my other throw comment here.

app-modules/service-management/src/Filament/Resources/ServiceRequestTypes/Pages/ManageServiceRequestTypePriorities.php
```diff
@@ -124,4 +157,85 @@ public function table(Table $table): Table
...
+                'default_priority_id' => $data['customer_defined_priorities']
+                    ? null
+                    : $data['default_priority_id'],
+            ]);
+        });
+
+        Notification::make()
+            ->title('Configuration saved')
+            ->success()
+            ->send();
```
Orrison: Instead of calling `DB::transaction` I would try/catch this all and start, commit, and rollback the transaction manually with the `DB` facade, and if there is an error then report the error, rollback the transaction, and send a notification that something went wrong.

Orrison (changes requested):

Orrison (approved):

PR #1354, opened 2026-08-10
[AIDAPP-1385]: Modify the Kanban view to use the slide out modal to view or edit items on click

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1385

### Technical Description

- Deleted `ViewPipelineEntry`.php page and its associated tests.
- Updated `PipelineEntryResource` to remove references to `ViewPipelineEntry`.
- Introduced `EditPipelineEntryAction` and `ViewPipelineEntryAction` for better action handling in the `PipelineEntryKanban` component.
- Adjusted tests to ensure proper functionality of viewing and editing pipeline entries through the Kanban interface.
- Ensured permissions are correctly checked for viewing and editing actions.

No

### Cleanup Tasks

No

---


payal-canyon (changes requested):
Could you please resolve the merge conflicts and then re-request my review? Thanks!

payal-canyon (changes requested):
Could you please resolve the merge conflicts and then re-request my review? Thanks!

app-modules/project/src/Filament/Resources/Pipelines/Pages/ManagePipelineEntries.php
```diff
@@ -112,13 +112,10 @@ public function table(Table $table): Table
...
             ])
             ->recordActions([
-                ViewAction::make()
-                    ->label('View')
-                    ->url(fn (PipelineEntry $record): string => PipelineEntryResource::getUrl('view', [
-                        'record' => $record,
-                        'pipeline' => $pipeline,
-                        'project' => $pipeline->project,
-                    ])),
+                ViewPipelineEntryAction::make()
```
payal-canyon: Could you please verify the intended surface here? `ManagePipelineEntries` uses a custom Blade view that doesn't render `{{ $this->table }}`, so the `ViewPipelineEntryAction` and `EditPipelineEntryAction` defined in `recordActions()` are currently unreachable.

Could you remove the record actions if the table is intentionally unused, and move the two authorization tests to the Kanban actions.

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -289,3 +294,136 @@
...
+        'assigned_to_type' => $contact->getMorphClass(),
+        'assigned_to_id' => $contact->getKey(),
+    ]);
+
+    livewire(PipelineEntryKanban::class, ['pipeline' => $pipeline])
+        ->mountAction('viewPipelineEntry', ['entry' => $entry->getKey()])
+        ->assertActionMounted('viewPipelineEntry')
+        ->assertSee('View Modal Task')
+        ->assertSee($contact->full_name)
+        ->assertSee($pipeline->stages->first()->name);
```
payal-canyon: Could you please replace the three `assertSee()` calls with `assertSchemaStateSet([...])` against the mounted action? Currently, those assertions can pass because the same values are already visible on the Kanban card, so they don't actually verify `PipelineEntryInfolist`.

app-modules/project/src/Filament/Resources/Pipelines/Actions/EditPipelineEntryAction.php
```diff
@@ -0,0 +1,73 @@
...
+    public static function make(?Pipeline $pipeline = null, ?string $name = null, ?Closure $after = null): EditAction
+    {
+        return EditAction::make($name ?? 'edit')
+            ->slideOver()
+            ->modalHeading('Edit Pipeline Task')
+            ->schema(PipelineEntryForm::components($pipeline))
+            ->after(function (PipelineEntry $record, array $data) use ($after): void {
+                $record->milestones()->sync($data['milestones'] ?? []);
+                $record->assets()->sync($data['assets'] ?? []);
+                $record->serviceRequests()->sync($data['serviceRequests'] ?? []);
```
payal-canyon, line 65: Could you please add coverage for these two cases?

- Add a test confirming that mounting `editPipelineEntry` with an entry belonging to a different pipeline fails. This is the only protection in `PipelineEntryKanban`, and the deleted 404 tests provide a good template.
- Add a test that editing an entry with existing milestones, assets, and service requests without changing them preserves those relationships. Also cover the case where the type is changed to none and those relationships are cleared.

app-modules/project/resources/views/components/entry-card.blade.php
```diff
@@ -64,20 +64,29 @@
...
 
 <div
+    role="button"
+    tabindex="0"
+    @can("update", $pipeline)
+        wire:click="mountAction('editPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
+        wire:keydown.enter="mountAction('editPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
+    @else
+        wire:click="mountAction('viewPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
+        wire:keydown.enter="mountAction('viewPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
```
payal-canyon: Could you please restore an icon button in the existing top-right wrapper? The open button was replaced with a whole-card click, but there’s no visual indication that the card is clickable, and the current cursor-move makes it look draggable rather than clickable.

payal-canyon (changes requested):

app-modules/project/resources/views/components/entry-card.blade.php
```diff
@@ -87,11 +88,22 @@ class="z-10 flex max-w-md transform cursor-move flex-col rounded-lg bg-white p-5
...
+                label="Open pipeline task"
+            />
+        @else
+            <x-filament::icon-button
+                class="fi-primary-color"
+                wire:click="mountAction('viewPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
+                icon="heroicon-m-arrow-top-right-on-square"
+                label="Open pipeline task"
+            />
+        @endcan
```
payal-canyon: Could you please take a reference from the Olympus sales pipeline Kanban and add the same **ellipsis dropdown** to this card?

The current card only has the open icon, so users can't **view** a task read-only or **remove it from the pipeline**. The previous **Remove from Pipeline** action was lost when the `ViewPipelineEntry` page was removed.

The existing Kanban pattern uses `heroicon-m-ellipsis-horizontal` with **View**, **Edit** (when the user has update permission), and **Remove** (`color="danger"`). Reusing that pattern here would keep the UX consistent and restore the missing actions.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1355, opened 2026-08-10
[AIDAPP-1383]: Reopen a Service Request if a update responds after a Request is closed

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1383

### Technical Description

> Reopen a Service Request if a update responds after a Request is closed

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -364,14 +364,14 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
                 return;
             }
 
+            app(ReopenServiceRequestAction::class)->execute($serviceRequest);
+
             $serviceRequestUpdate = $serviceRequest->serviceRequestUpdates()->create([
                 'update' => $parser->getMessageBody('text') ?: $parser->getMessageBody('html'),
                 'internal' => false,
                 'created_by_id' => $serviceRequest->respondent->getKey(),
                 'created_by_type' => $serviceRequest->respondent->getMorphClass(),
             ]);
-
-            app(ReopenServiceRequestAction::class)->execute($serviceRequest);
 
             foreach ($parser->getAttachments() as $attachment) {
                 try {
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -364,7 +364,6 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
                 return;
             }
 
-            // Create SR Update
             $serviceRequestUpdate = $serviceRequest->serviceRequestUpdates()->create([
                 'update' => $parser->getMessageBody('text') ?: $parser->getMessageBody('html'),
                 'internal' => false,
@@ -372,7 +371,6 @@ class ProcessSesS3InboundEmail implements ShouldQueue, ShouldBeUnique, NotTenant
                 'created_by_type' => $serviceRequest->respondent->getMorphClass(),
             ]);
 
-            // Reopen the service request if it was closed when this reply arrived.
             app(ReopenServiceRequestAction::class)->execute($serviceRequest);
 
             foreach ($parser->getAttachments() as $attachment) {
--- a/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
@@ -47,6 +47,7 @@ use AidingApp\Engagement\Exceptions\UnableToRetrieveContentFromSesS3EmailPayload
 use AidingApp\Engagement\Models\UnmatchedInboundCommunication;
 use AidingApp\Engagement\Notifications\IneligibleContactSesS3InboundEmailServiceRequestNotification;
 use AidingApp\Notification\Models\OutboundEmailMessageId;
+use AidingApp\ServiceManagement\Actions\ReopenServiceRequestAction;
 use AidingApp\ServiceManagement\Enums\EmailAutomaticCreationContactCreateCondition;
 use AidingApp\ServiceManagement\Enums\SystemServiceRequestClassification;
 use AidingApp\ServiceManagement\Models\ServiceRequest;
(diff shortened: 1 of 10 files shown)


app-modules/service-management/src/Actions/ReopenServiceRequestAction.php
```diff
@@ -0,0 +1,68 @@
...
+use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
+use Carbon\CarbonImmutable;
+
+class ReopenServiceRequestAction
+{
+    public function execute(ServiceRequest $serviceRequest): void
+    {
+        if ($serviceRequest->status?->classification !== SystemServiceRequestClassification::Closed) {
+            return;
+        }
```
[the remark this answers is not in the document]
mittal-canyon: Verified this locally — lazy loading isn't disabled anywhere in this app (no Model::preventLazyLoading()), so $serviceRequest->status resolves correctly even without eager loading. Confirmed the reopen works end-to-end via both the portal and email flows. Not making a change here.

app-modules/service-management/tests/Tenant/Actions/ReopenServiceRequestActionTest.php
```diff
@@ -0,0 +1,95 @@
...
+});
+
+it('reports an error and leaves the status unchanged when no open status exists', function () {
+    Exceptions::fake();
+
+    // The seeded "New" status is system protected, so bypass row-protection triggers to clear all statuses.
+    $connection = (new ServiceRequestStatus())->getConnection();
+    $connection->statement('SET session_replication_role = replica');
+    ServiceRequestStatus::query()->forceDelete();
+    $connection->statement('SET session_replication_role = DEFAULT');
```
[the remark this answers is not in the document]
mittal-canyon: General point, but this app only ever runs on PostgreSQL, so there's no other driver to guard against here. Leaving as-is.

app-modules/service-management/src/Actions/ReopenServiceRequestAction.php
```diff
@@ -34,32 +34,35 @@
...
 
-            ProjectArchivingFeature::activate();
-        });
-    }
+        $openStatus = ServiceRequestStatus::query()
+            ->where('classification', SystemServiceRequestClassification::Open)
+            ->orderBy('sort')
```
payal-canyon, line 54: Could you please make the ordering deterministic by adding `->orderBy('created_at')->orderBy('id')` after `->orderBy('sort')`? Since multiple statuses can have the same `sort` value, the current query can return an arbitrary status and may even pick a newly created status over the seeded `New` status.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
payal-canyon, line 87: Could you please update the portal UI after a successful reopen? The current response only returns `serviceRequestUpdates`, so the status badge remains "Closed" until the page is refreshed.

Either return the updated `statusName`/`statusColor` and update `serviceRequest` in `submitUpdate`, or simply call `getData(currentPage.value)` after the post succeeds.

app-modules/service-management/tests/Tenant/Actions/ReopenServiceRequestActionTest.php
```diff
@@ -0,0 +1,95 @@
...
+it('reopens a closed service request to the first open status ordered by sort', function () {
+    ServiceRequestStatus::factory()->create([
+        'classification' => SystemServiceRequestClassification::Open,
+        'sort' => 999,
+    ]);
+
+    $expectedFirstOpenStatus = ServiceRequestStatus::query()
+        ->where('classification', SystemServiceRequestClassification::Open)
+        ->orderBy('sort')
+        ->firstOrFail();
```
payal-canyon: The test currently re-runs the same `classification + orderBy('sort')` query, so it doesn't really verify the ordering rule.

Could we create two Open statuses with explicit `sort` values and assert that the expected specific status is selected? That would actually cover the ordering behavior.

Also, could we use the existing `->open()` factory state instead of setting the classification manually?

app-modules/service-management/src/Actions/ReopenServiceRequestAction.php
```diff
@@ -34,32 +34,35 @@
...
+
+            return;
+        }
 
-            Schema::table('projects', function (Blueprint $table) {
-                $table->dropColumn('archived_at');
-            });
-        });
+        $serviceRequest->status()->associate($openStatus);
+        $serviceRequest->status_updated_at = CarbonImmutable::now();
```
payal-canyon: Could we remove the explicit `status_updated_at` assignment here? `ServiceRequestObserver::saving()` already updates it whenever `status_id` changes, so this line is redundant.

app-modules/engagement/src/Jobs/ProcessSesS3InboundEmail.php
```diff
@@ -363,14 +364,17 @@ protected function handleServiceRequestReply(Tenant $tenant, Parser $parser, str
...
+            // Create SR Update
             $serviceRequestUpdate = $serviceRequest->serviceRequestUpdates()->create([
                 'update' => $parser->getMessageBody('text') ?: $parser->getMessageBody('html'),
                 'internal' => false,
                 'created_by_id' => $serviceRequest->respondent->getKey(),
                 'created_by_type' => $serviceRequest->respondent->getMorphClass(),
             ]);
 
+            // Reopen the service request if it was closed when this reply arrived.
+            app(ReopenServiceRequestAction::class)->execute($serviceRequest);
```
payal-canyon: Could we remove these comments? They just restate what the code is already doing.

In particular, `// Reopen the service request...` duplicates `ReopenServiceRequestAction::execute()`, and the reduced `// Create SR Update` comment no longer adds useful context.

payal-canyon (changes requested):
Could you please add `$this->afterCommit = true` to `SendEducatableServiceRequestStatusChangeNotification`, matching the other service request notifications?

This notification is dispatched inside the reopen transaction, and all queue connections have `after_commit` disabled by default. Without this, a worker could process it before the status update commits and send the contact the old "Closed" status. The sibling notifications already use `afterCommit`, so this would keep the behavior consistent.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
payal-canyon, line 87: Could we also refresh the request status after submitting an update? when a contact updates a **Closed** request, it reopens correctly, but the portal still shows **Closed** until the page is refreshed.

**Steps to reproduce:**

1. Open a contact's request in the portal with **Closed** status.
2. Submit a new update.
3. The update appears, but the status badge still shows **Closed**.
4. Refresh the page - it changes to **New**.

Could you please test this scenario and fix the stale status in the UI?

payal-canyon (changes requested):
In ServiceRequestStatusChanged could we also set `$this->afterCommit = true;` in this notification's constructor?

Since it's `ShouldQueue` and uses the updated `serviceRequest->status` when sending, there's a chance the worker could process it before the portal transaction commits and send the old status (e.g. **Closed**) after the request has been reopened.

This is the same fix already applied to `SendEducatableServiceRequestStatusChangeNotification`.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/StoreServiceRequestUpdateController.php
mittal-canyon: Yeah, it was an existing pattern, so I haven't updated.

payal-canyon (approved):

PR #1357, opened 2026-08-10
[AIDAPP-1371]: Enhance the capabilities of the pipeline selection modal

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1371

### Technical Description

Moves pipeline management into the slide-over pipeline switcher and adds the ability to archive a pipeline, with the whole change gated behind the `PipelineArchivingFeature` flag.

**Archiving model**

- Adds an `archived_at` column to `pipelines` and `pipeline_stages` (via the `CanBeArchived` trait), so both are archivable.
- Archiving cascades along ownership: archiving a `Pipeline` archives its `PipelineStage`s, which archive their `PipelineEntry`s (tasks). Each entry then re-evaluates its linked `ProjectMilestone`s — a milestone is archived once it has no remaining active linked task. There is no user-facing unarchive flow, so the cascade is one-directional (archive only).
- `ProjectMilestone` gains a `reevaluateArchivedState()` invariant helper that owns this milestone-follows-its-tasks rule, using a quiet archive to avoid feedback loops.
- Ownership cascades iterate with `eachById()` so large graphs are chunked rather than loaded into memory all at once.

**Atomicity**

- The vendor `CanBeArchived::archive()` fires the cascade synchronously and is not itself transactional, so both archive entry points wrap it in `DB::transaction()` — the switcher footer action and the relation-manager row action — ensuring a mid-cascade failure rolls the whole graph back.

**UI**

- The pipeline switcher slide-over hides archived pipelines from its list, exposes an **Archive** footer action (confirmation modal), and clears the current selection / advances to the next pipeline when the active one is archived or none remain.
- Archived milestones and tasks are hidden from their list surfaces — the project milestones page and widget, and the pipeline task (kanban) board — consistent with the switcher and milestone-selector filtering.
- Archived milestones are excluded from the task form's milestone selector.
- The projects `PipelinesRelationManager` hides archived pipelines and gains an **Archive** row action (with an archive-box icon) consistent with the switcher.

All new behaviour resolves to the old path when the feature flag is inactive.

No

### Cleanup Tasks

Yes, Feature Flag: `App\Features\PipelineArchivingFeature` — `.cleanup-tasks/2026_08_07_pipeline_archiving_cleanup.md`

---

### DIFF
--- a/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
+++ b/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
@@ -42,6 +42,7 @@ use App\Features\PipelineArchivingFeature;
 use Filament\Actions\Action;
 use Filament\Forms\Components\TableSelect;
 use Filament\Notifications\Notification;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Support\Facades\DB;
 
 trait HasPipelineSwitcherAction
@@ -69,6 +70,10 @@ trait HasPipelineSwitcherAction
                     blank($pipelineId) ||
                     ! Pipeline::query()
                         ->where('project_id', $this->getPipelineSwitcherProjectId())
+                        ->when(
+                            PipelineArchivingFeature::active(),
+                            fn (Builder $query): Builder => $query->withoutArchived(),
+                        )
                         ->whereKey($pipelineId)
                         ->exists()
                 ) {
--- a/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
+++ b/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
@@ -164,6 +164,7 @@ trait HasPipelineSwitcherAction
             ->modalHeading('Archive Pipeline')
             ->modalDescription('Are you sure you want to archive this pipeline? All related milestones and tasks will also be archived.')
             ->modalSubmitActionLabel('Archive')
+            ->cancelParentActions(fn (): bool => ! $this->switcherHasSelectablePipelines())
             ->action(function (): void {
                 $pipelineId = data_get($this->mountedActions, '0.data.pipeline_id');
 
@@ -181,6 +182,14 @@ trait HasPipelineSwitcherAction
             ->first();
 
         return $pipeline !== null && (bool) auth()->user()?->can('delete', $pipeline);
+    }
+
+    protected function switcherHasSelectablePipelines(): bool
+    {
+        return Pipeline::query()
+            ->where('project_id', $this->getPipelineSwitcherProjectId())
+            ->withoutArchived()
+            ->exists();
     }
 
     abstract protected function getPipelineSwitcherProjectId(): ?string;
--- a/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
+++ b/app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
@@ -38,6 +38,7 @@ namespace AidingApp\Project\Filament\Concerns;
 
 use AidingApp\Project\Filament\Tables\ProjectPipelinesTable;
 use AidingApp\Project\Models\Pipeline;
+use App\Features\PipelineArchivingFeature;
 use Filament\Actions\Action;
 use Filament\Forms\Components\TableSelect;
 use Filament\Notifications\Notification;
@@ -46,7 +47,7 @@ trait HasPipelineSwitcherAction
 {
     public function selectPipelineAction(): Action
     {
-        return Action::make('selectPipeline')
+        $action = Action::make('selectPipeline')
             ->label('Select Pipeline')
             ->modalHeading('Select Pipeline')
             ->modalSubmitActionLabel('Select')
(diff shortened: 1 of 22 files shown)


app-modules/project/src/Filament/Concerns/HasPipelineSwitcherAction.php
```diff
@@ -80,11 +82,107 @@ public function selectPipelineAction(): Action
...
+            $action->getModalSubmitAction(),
+            $action->getModalCancelAction(),
+            $this->archivePipelineAction(),
+        ]);
+    }
+
+    public function archivePipelineFromSwitcher(string $pipelineId): void
+    {
+        $projectId = $this->getPipelineSwitcherProjectId();
+
```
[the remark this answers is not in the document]
mxm1070: Good catch — fixed in 3139de7f3. `archivePipelineFromSwitcher()` now early-returns when `PipelineArchivingFeature::active()` is false, so a direct call to the public method can't bypass the flag the way the hidden UI action already does. Added a widget test asserting a direct `->call()` with the flag inactive does not archive (verified it fails without the guard). Also audited the other archive entry points: the RelationManager `ArchiveAction` is already server-side-guarded by Filament's `->visible(flag)` + authorize, and the model `booted()` cascades are intentionally ungated (downstream cascade mechanics, not entry points) — this was the only unguarded path.

app-modules/project/tests/Tenant/PipelineArchivingFeatureCascadeTest.php
```diff
@@ -0,0 +1,200 @@
...
+    PipelineEntry::archiving(function (): void {
+        throw new RuntimeException('boom');
+    });
+
+    try {
+        livewire(ProjectWorkPipelineWidget::class, ['record' => $project])
+            ->call('archivePipelineFromSwitcher', $pipeline->getKey());
+    } catch (Throwable) {
+        // Expected: the forced failure propagates out of the transaction.
+    }
```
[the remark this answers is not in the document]
mxm1070: This isn't reachable in our test harness. Model event listeners registered via `PipelineEntry::archiving()` attach to the *current* event dispatcher, and the harness refreshes the application before every test (`setUp()` → `refreshApplication()`), with `DatabaseServiceProvider::boot()` swapping in a fresh `Model` event dispatcher each time — so the throwing listener is discarded along with that test's dispatcher and can't leak into a later test. Under `--parallel` it's even more isolated: paratest runs workers as separate OS processes with no shared static state, so a listener in one worker is invisible to the others.

Verified empirically: registering the throwing `archiving` listener in one test and then archiving a fresh `PipelineEntry` in the next test (same file, same process, run sequentially) archives cleanly with no `boom` — exactly the same-process leak vector this describes. Given that, leaving the test as-is.

app-modules/project/src/Models/Pipeline.php
```diff
@@ -95,4 +97,15 @@ public function entries(): HasManyThrough
...
+
+    protected static function booted(): void
+    {
+        static::archived(function (Pipeline $pipeline): void {
+            $pipeline->stages()->withoutArchived()->get()->each->archive();
+        });
+
+        static::unarchived(function (Pipeline $pipeline): void {
+            $pipeline->stages()->onlyArchived()->get()->each->unarchive();
+        });
```
Orrison: Let's remove this for now. There is currently no flow where a user is able to unarchive something. And when we are asked to allow for that I think we need to spend a little more time on this here.

For example, we should consider ONLY unarchiving the stages that were archived >= the time it's pipeline was archived. If we don't do that you run the risk of restoring stages that they archived independently of the pipeline archive.

app-modules/project/src/Models/Pipeline.php
```diff
@@ -95,4 +97,15 @@ public function entries(): HasManyThrough
             'id'
         );
     }
+
+    protected static function booted(): void
+    {
+        static::archived(function (Pipeline $pipeline): void {
+            $pipeline->stages()->withoutArchived()->get()->each->archive();
```
Orrison: `->get()->each` will load these all directly into memory won't it? Can we just do an `eachById` so that it properly chunks in case there are a lot of stages and/or underlying models?

app-modules/project/src/Models/PipelineEntry.php
```diff
@@ -86,6 +88,13 @@ class PipelineEntry extends Model implements Auditable
         'is_visible_to_guests' => 'boolean',
     ];
 
+    public function reevaluateLinkedMilestones(): void
+    {
+        $this->milestones()->get()->each(
```
Orrison: Same as above, I am pretty sure `->get()->each` is not very memory safe. Anywhere we are doing that, let's not.

app-modules/project/src/Models/PipelineEntry.php
```diff
@@ -142,4 +151,10 @@ public function serviceRequests(): BelongsToMany
             ->using(PipelineEntryServiceRequest::class)
             ->withTimestamps();
     }
+
+    protected static function booted(): void
+    {
+        static::archived(fn (PipelineEntry $entry) => $entry->reevaluateLinkedMilestones());
+        static::unarchived(fn (PipelineEntry $entry) => $entry->reevaluateLinkedMilestones());
```
Orrison: Same as above, I think we need to either just NOT do unarchiving now, or if it is easy enough to do we should scope these all to only unarchive items that were archived at >= it's parent.

Orrison (changes requested):

app-modules/project/src/Models/Pipeline.php
mxm1070: There's no user-facing unarchive flow anywhere in the app (no UnarchiveAction, no restore UI), so a blind cascade restore would resurrect stages and tasks that were archived independently before the pipeline was. Removed the `unarchived` hook here in dec4d9a29. If we add an unarchive flow later we can scope the restore to children where `archived_at >= parent.archived_at`.

app-modules/project/src/Models/Pipeline.php
mxm1070: You're right, `->get()->each` hydrates the whole collection into memory at once. Switched to `eachById()` in dec4d9a29 so it chunks through the stages instead.

app-modules/project/src/Models/PipelineEntry.php
mxm1070: Same concern applies, `reevaluateLinkedMilestones` now uses `eachById()` in dec4d9a29 (passing the qualified `project_milestones.id` since it's a belongsToMany over a pivot that has its own `id` column).

app-modules/project/src/Models/PipelineEntry.php
mxm1070: Same as the Pipeline hook, there's no unarchive flow to exercise this yet, so rather than scope it I removed the `unarchived` cascade here in dec4d9a29. We can revisit with the `archived_at >= parent` approach if/when we add a restore flow.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ManageMilestonesTest.php
```diff
@@ -118,6 +118,22 @@
         ->assertCanSeeTableRecords($project->milestones);
 });
 
+it('hides archived milestones when the archiving feature is active', function () {
```
Orrison: Any test you write, here and others I am seeing should not be testing "when the archiving feature is active".

In 9/10 cases we activate the feature flag within a migration, so you would assume that it is active, without needing to make it active or even mention the FF here in this test.

This is important as if we did things this way we would then need to update our cleanup tasks to come in and say remove the "when the archiving feature is active" from the test names, etc.

The only time you really need to do things with FFs in tests is if you are testing the backwards compatibility of it being off.

Orrison (changes requested):

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ManageMilestonesTest.php
mxm1070: You're right — the migration commits `PipelineArchivingFeature::activate()`, so the flag is already active by default in the test DB (RefreshDatabase runs the migration, and per-test `deactivate()` rolls back). Dropped the "when the archiving feature is active" wording from the titles and the redundant explicit `::activate()` calls across the PR's active-path tests in b26335906, keeping FF handling only in the backwards-compat (`deactivate()`) cases. That also keeps the cleanup task from having to strip flag references out of test names later.

Orrison (approved):

PR #1360, opened 2026-08-12
[AIDAPP-1370]: Update milestones in a project to managed within a pipeline itself

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1370

### Technical Description

> Update milestones in a project to managed within a pipeline itself

> No

### Cleanup Tasks

> Yes, Feature Flag: `PipelineEntryMilestoneFeature`

---


Orrison:
@mxm1070 just a note that this was requested to be hotfixed. So I will make a release branch and reach out to you and @amit-canyon soon on how to handle this here. Still review it, in fact prioritize reviewing it, but don't merge it yet till was have it target the new branch.

Orrison (dismissed):
@amit-canyon please rebase on and change the base of this PR to `release/1.43.1` so we can hotfix release it. Thank you!

app-modules/project/src/Filament/Resources/Pipelines/Forms/PipelineEntryForm.php
```diff
@@ -143,9 +145,13 @@ public static function components(?Pipeline $pipeline = null, bool $isStageVisib
                 ->inline()
                 ->live()
                 ->afterStateHydrated(function (Set $set, ?PipelineEntry $record) {
-                    $state = $record?->milestones()->exists();
+                    $state = PipelineEntryMilestoneFeature::active()
+                        ? $record?->milestone()->exists()
+                        : $record?->milestones()->exists();
 
-                    if (filled($state)) {
+                    Log::info('PipelineEntryForm::components: afterStateHydrated: $state=' . var_export($state, true));
```
mxm1070: Looks like a debug leftover:
```suggestion
```
(removing the `Log::info('PipelineEntryForm::components: afterStateHydrated: $state=' . var_export($state, true));` line entirely, plus the now-unused `use Illuminate\Support\Facades\Log;` import)

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectMilestonesWidget.php
```diff
@@ -36,17 +36,16 @@
 
 namespace AidingApp\Project\Filament\Resources\Projects\Widgets;
 
+use AidingApp\Project\Filament\Actions\CreateProjectMilestoneAction;
```
mxm1070: `ProjectMilestonesWidget.php` doesn't look referenced anywhere else now. Please delete it as dead code (class file, blade view, and its permission test).

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -242,6 +357,54 @@ protected function getTableHeadingView(?Pipeline $pipeline): View
         ]);
     }
 
+    protected function milestoneProgressDescription(PipelineEntry $record, ?Pipeline $pipeline): string
```
mxm1070, line 413: `milestoneProgressDescription()` was running 2 queries per unique milestone on the page (a total-count query plus a cloned query with an extra `whereHas` filter for the completed count) — e.g. 20 distinct milestones meant ~40 queries.

Since the pipeline is known before the table renders, I replaced this with a single grouped/aggregate query (`GROUP BY project_milestone_id` with a conditional `COUNT` for completed vs total, joined through `pipelineStage`) that precomputes progress for every milestone in one query. This matches the `withCount`/aggregate pattern already used elsewhere in this module (`PipelineEntryKanban.php`) and across the codebase.

Pushed this directly as commit `f778c8dce` rather than just suggesting it, since it was straightforward to verify (confirmed identical output against real data, and confirmed via query log: 1 query total regardless of milestone count).

app-modules/project/src/Filament/Resources/Pipelines/Pages/EditPipelineEntry.php
```diff
@@ -160,14 +161,23 @@ public function save(): void
 
         $data = $this->form->getState();
 
-        $milestones = $data['milestones'] ?? [];
+        //TODO: PipelineEntryMilestoneFeature clean up: Please remove the entire if block below.
+        if (! PipelineEntryMilestoneFeature::active()) {
+            $milestones = $data['milestones'] ?? [];
+            unset($data['milestones']);
+        }
```
mxm1070: When `PipelineEntryMilestoneFeature` is active, `$data['milestones']` is never unset before `$pipelineEntry->update($data)`. Harmless today only because `milestones` isn't in `$fillable`, but worth tidying:
```suggestion
        if (! PipelineEntryMilestoneFeature::active()) {
            $milestones = $data['milestones'] ?? [];
            unset($data['milestones']);
        } else {
            unset($data['milestones']);
        }
```

app-modules/project/database/migrations/2026_08_11_090933_add_project_milestone_id_to_pipeline_entries_table.php
```diff
@@ -0,0 +1,56 @@
...
+
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('pipeline_entries', function (Blueprint $table) {
+            $table->foreignUuid('project_milestone_id')->nullable()->references('id')->on('project_milestones')->nullOnDelete();
```
mxm1070: Every other FK in this module (18 of 19) uses `->constrained()` rather than spelling out `->references()->on()`:
```suggestion
            $table->foreignUuid('project_milestone_id')->nullable()->constrained('project_milestones')->nullOnDelete();
```

.cleanup-tasks/2026_08_11_pipeline_entry_milestone_feature.md
```diff
@@ -0,0 +1,16 @@
...
+
+- App\Features\PipelineEntryMilestoneFeature
+
+## Temporary Migrations
+
+- app-modules/project/database/migrations/2026_08_11_091532_tmp_backfill_pipeline_entry_milestones_data.php
+
+## Additional Cleanup
+
+- Search for `TODO: PipelineEntryMilestoneFeature clean up:` and follow the instructions at each site.
```
mxm1070: The cleanup task and inline `TODO` comments use `TODO: PipelineEntryMilestoneFeature clean up: ...` instead of the house convention `TODO: Cleanup Task (pipeline-entry-milestone): ...`. Worth aligning with the convention (this touches the cleanup-task file plus several inline comments across the codebase, not just one spot).

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -137,10 +157,30 @@ public function table(Table $table): Table
...
-                Group::make('pipelineStage.name')
-                    ->label('Stage')
-                    ->collapsible(),
+                function () use ($pipeline) {
+                    if (! PipelineEntryMilestoneFeature::active()) {
+                        return Group::make('pipelineStage.name')
+                            ->label('Stage')
+                            ->collapsible();
+                    }
+
```
mxm1070, line 188: "No Associated Milestone" is correctly grouped last, but only incidentally because Postgres sorts NULL group keys last by default — there's no explicit ordering logic and no test pins this down. A regression test would guard against this silently breaking later.

mxm1070 (changes requested):
Just a few things to be addressed before merge.

One note on process: while reviewing, I found and fixed an N+1 query in `milestoneProgressDescription()` (see inline comment below) and pushed that fix directly to this branch rather than just describing it, since it was a quick, low-risk win.

app-modules/project/src/Filament/Resources/Pipelines/Pages/EditPipelineEntry.php
amit-canyon: @mxm1070  As per the existing code, we store `milestones` in a separate variable and unset it from `$data` because milestones are saved using `sync()` and are not directly related to the `pipeline_entries` table.

Please correct me if I’m wrong, and let me know your thoughts on this. Thanks!

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -137,10 +157,45 @@ public function table(Table $table): Table
...
+                                        ->select('title')
+                                        ->whereColumn(
+                                            (new ProjectMilestone())->qualifyColumn('id'),
+                                            $model->qualifyColumn('project_milestone_id'),
+                                        ),
+                                    $direction,
+                                );
+                        })
+                        ->getTitleFromRecordUsing(
+                            fn (PipelineEntry $record): string => $record->milestone->title ?? 'No Associated Milestone'
```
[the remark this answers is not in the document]
amit-canyon: @Orrison  If I add ? before title, the linter reports an error and asks me to remove it, while Copilot suggests adding it. IMO, the code works correctly without ?, and I’ve tested both approaches without encountering any errors.

app-modules/project/src/Filament/Resources/Pipelines/Pages/EditPipelineEntry.php
mxm1070: Actually, tracing this further — the `ModalTableSelect::make('milestones')` field itself is already tagged for removal in `PipelineEntryForm.php`, which is exactly what the cleanup task tracks. Once that's deleted, `$data['milestones']` won't exist at all, so this whole concern goes away on its own. Adding a fix for it now would just be another untagged line cleanup wouldn't catch. Leaving this as-is — good catch flagging it though.

mxm1070 (dismissed):
Looks good!

Orrison (commented):
@amit-canyon and @mxm1070 Product has changed their mind on this being hotfixed. So I have targeted this back at `main`. @amit-canyon please resolve the merge conflicts and then once that is done please re-request @mxm1070's review. @mxm1070 you should be good to merge this into `main` once you approve

amit-canyon:
@mxm1070 

I’ve made several changes related to the milestone relationship because some other PRs were merged before this one. Could you please re-review the changes? Thanks!

cc: @Orrison

app-modules/project/src/Models/PipelineEntry.php
```diff
@@ -90,12 +91,9 @@ class PipelineEntry extends Model implements Auditable
...
-        $this->milestones()->eachById(
-            function (ProjectMilestone $milestone): void {
-                $milestone->reevaluateArchivedState();
-            },
-            column: 'project_milestones.id',
-        );
+        if ($this->milestone) {
+            $this->milestone->reevaluateArchivedState();
+        }
     }
```
mxm1070, line 105: `reevaluateLinkedMilestones()` isn't gated by `PipelineEntryMilestoneFeature::active()`, unlike every other call site touching this relationship in this PR. During the zero-downtime rollout window, existing associations still live only in the legacy `pipeline_entry_milestones` pivot, so `project_milestone_id` (and `$this->milestone`) is null — the `if ($this->milestone)` check is simply false, so `reevaluateArchivedState()` never runs for those entries, versus the old pivot-walk which worked regardless of flag state. (`ProjectMilestone::pipelineEntries()` itself doesn't need to change — its only other call site is safe since the milestone-delete UI only renders once the flag is active.)
```suggestion
    public function reevaluateLinkedMilestones(): void
    {
        if (PipelineEntryMilestoneFeature::active()) {
            $this->milestone?->reevaluateArchivedState();

            return;
        }

        $this->milestones()->eachById(
            fn (ProjectMilestone $milestone) => $milestone->reevaluateArchivedState(),
            column: 'project_milestones.id',
        );
    }
```

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -280,6 +407,59 @@ protected function getTableHeadingView(?Pipeline $pipeline): View
...
+    protected function loadMilestoneProgressDescriptions(Pipeline $pipeline): void
+    {
+        if ($this->milestoneProgressDescriptionsLoadedForPipelineId === $pipeline->getKey()) {
+            return;
+        }
+
+        $counts = PipelineEntry::query()
+            ->whereNotNull('project_milestone_id')
+            ->whereHas('pipelineStage', fn (Builder $query) => $query->where('pipeline_id', $pipeline->getKey()))
+            ->join('pipeline_stages', 'pipeline_stages.id', '=', 'pipeline_entries.pipeline_stage_id')
```
mxm1070, line 447: This aggregate query has no `withoutArchived()` filter, unlike the main table query in this same widget, which conditionally excludes archived entries when `PipelineArchivingFeature::active()`. Reproduced with a test: an archived, complete entry is still counted toward a milestone's progress, showing 50% when the table (correctly hiding it) would suggest 0%.
```suggestion
            ->when(
                PipelineArchivingFeature::active(),
                fn (Builder $query): Builder => $query->withoutArchived(),
            )
            ->join('pipeline_stages', 'pipeline_stages.id', '=', 'pipeline_entries.pipeline_stage_id')
```

app-modules/project/src/Models/PipelineEntryMilestone.php
```diff
@@ -41,6 +41,8 @@
 use Illuminate\Database\Eloquent\Relations\Pivot;
 
 /**
+ * TODO: Remove this model once the milestone feature is fully implemented and the pivot table is no longer needed.
```
mxm1070: This introduces a bare TODO that doesn't match the `TODO: Cleanup Task (pipeline-entry-milestone)` tag used at every other TODO site in this PR. Since the cleanup-task file's "Additional Cleanup" section instructs searching for that exact tag, this one would be silently missed, leaving the pivot model as permanent dead code.
```suggestion
 * TODO: Cleanup Task (pipeline-entry-milestone): Remove this model once the milestone feature is fully implemented and the pivot table is no longer needed.
```

mxm1070 (changes requested):
Re-reviewed after the recent rebase (large refactor consolidating the milestone relationship to a direct FK). The refactor itself looks solid, but found 4 issues worth addressing before merge.

One isn't anchorable as an inline comment since the touched line predates this PR (only the new progress-cache mechanism it interacts with is new):

**`ProjectWorkPipelineWidget::editPipelineEntryAction()`** (line ~300):
```php
    public function editPipelineEntryAction(): Action
    {
        return EditPipelineEntryAction::make(
            $this->getSelectedPipeline(),
            'editPipelineEntry',
            after: fn () => $this->dispatch('projectPipelineUpdated'),
        )
```
This only dispatches `projectPipelineUpdated` — unlike every other mutation path in this widget (create, delete, drag-and-drop), it never calls `$this->resetMilestoneProgressDescriptions()`. The dispatch only notifies *other* listening components (`ProjectDashboardHeaderWidget`, `ProjectStatsWidget`), not this one, so the memoized progress cache is never invalidated here. Reproduced: after moving an entry to a "Complete" stage via this action, the milestone's progress badge keeps showing the stale pre-edit percentage until the page is refreshed. Suggested fix:
```suggestion
            after: fn () => [
                $this->resetMilestoneProgressDescriptions(),
                $this->dispatch('projectPipelineUpdated'),
            ],
```

mxm1070 (approved):

PR #1362, opened 2026-08-13
[AIDAPP-1384]: Update SLA management to exclude time in the waiting classification

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1384

### Technical Description

- Added a `service_request_status_periods` ledger (model, factory, migrations) and observers/actions to record periods on create, status changes, and status reclassification.
- Updated `ServiceRequest` SLA timing calculations (`getResolutionSeconds`(), `getLatestResponseSeconds`(), and stored time_to_resolution on close) to subtract excluded classifications when `SlaWaitingExclusionFeature` is active.
- Added/updated tests covering waiting/closed exclusion behavior and period-recording behaviors, plus backfill + feature activation via a temporary migration.

 NO

### Cleanup Tasks

- Feature Flag Cleanup: .cleanup-tasks/2026_08_12_sla_waiting_exclusion_feature.md

---

### DIFF
--- a/app-modules/report/src/Filament/Widgets/ServiceRequestFeedbackTable.php
+++ b/app-modules/report/src/Filament/Widgets/ServiceRequestFeedbackTable.php
@@ -92,6 +92,7 @@ class ServiceRequestFeedbackTable extends BaseWidget
                         'serviceRequest.priority.sla',
                         'serviceRequest.respondent',
                         'serviceRequest.assignedTo.user',
+                        'serviceRequest.statusPeriods',
                     ])
                     ->when(
                         $startDate && $endDate,
--- a/app-modules/service-management/database/factories/ServiceRequestStatusPeriodFactory.php
+++ b/app-modules/service-management/database/factories/ServiceRequestStatusPeriodFactory.php
@@ -34,29 +34,33 @@
 </COPYRIGHT>
 */
 
-namespace AidingApp\ServiceManagement\Observers;
+namespace AidingApp\ServiceManagement\Database\Factories;
 
-use AidingApp\ServiceManagement\Jobs\RecordReclassifiedServiceRequestStatusPeriods;
+use AidingApp\ServiceManagement\Enums\SystemServiceRequestClassification;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
 use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
-use Carbon\CarbonImmutable;
-use Illuminate\Support\Facades\DB;
+use AidingApp\ServiceManagement\Models\ServiceRequestStatusPeriod;
+use Illuminate\Database\Eloquent\Factories\Factory;
 
-class ServiceRequestStatusObserver
+/**
+ * @extends Factory<ServiceRequestStatusPeriod>
+ */
+class ServiceRequestStatusPeriodFactory extends Factory
 {
-    public function creating(ServiceRequestStatus $serviceRequestStatus): void
+    /**
+     * Define the model's default state.
+     *
+     * @return array<string, mixed>
+     */
+    public function definition(): array
     {
-        if (! isset($serviceRequestStatus->sort)) {
-            $serviceRequestStatus->setAttribute(
-                'sort',
-                DB::raw('(SELECT COALESCE(MAX(service_request_statuses.sort), 0) + 1 FROM service_request_statuses)')
-            );
-        }
-    }
-
-    public function updated(ServiceRequestStatus $serviceRequestStatus): void
-    {
-        if ($serviceRequestStatus->wasChanged('classification')) {
-            RecordReclassifiedServiceRequestStatusPeriods::dispatch($serviceRequestStatus, CarbonImmutable::now());
-        }
+        return [
+            'service_request_id' => ServiceRequest::factory(),
+            'classification' => fn () => $this->faker->randomElement(SystemServiceRequestClassification::cases()),
+            'service_request_status_id' => fn (array $attributes) => ServiceRequestStatus::factory()->state([
+                'classification' => $attributes['classification'],
+            ]),
+            'started_at' => now(),
+        ];
     }
 }
--- a/app-modules/service-management/database/factories/ServiceRequestStatusPeriodFactory.php
+++ b/app-modules/service-management/database/factories/ServiceRequestStatusPeriodFactory.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Database\Factories;
+
+use AidingApp\ServiceManagement\Enums\SystemServiceRequestClassification;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
+use AidingApp\ServiceManagement\Models\ServiceRequestStatusPeriod;
+use Illuminate\Database\Eloquent\Factories\Factory;
+
+/**
+ * @extends Factory<ServiceRequestStatusPeriod>
+ */
+class ServiceRequestStatusPeriodFactory extends Factory
+{
+    /**
+     * Define the model's default state.
+     *
+     * @return array<string, mixed>
+     */
+    public function definition(): array
+    {
+        $status = ServiceRequestStatus::factory();
+
+        return [
+            'service_request_id' => ServiceRequest::factory(),
+            'service_request_status_id' => $status,
+            'classification' => $this->faker->randomElement(SystemServiceRequestClassification::cases()),
+            'started_at' => now(),
+        ];
+    }
+}
--- a/app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
+++ b/app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
@@ -1,0 +1,177 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use App\Features\SlaWaitingExclusionFeature;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Query\Builder;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Str;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function (): void {
+            $classifications = DB::table('service_request_statuses')
+                ->pluck('classification', 'id')
+                ->all();
+
+            DB::table('service_requests')
+                ->select(['id', 'status_id', 'created_at'])
+                ->whereNotExists(function (Builder $query): void {
+                    $query->select(DB::raw(1))
+                        ->from('service_request_status_periods')
+                        ->whereColumn('service_request_status_periods.service_request_id', 'service_requests.id');
+                })
+                ->orderBy('id')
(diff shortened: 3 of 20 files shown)


app-modules/service-management/database/factories/ServiceRequestStatusPeriodFactory.php
```diff
@@ -0,0 +1,69 @@
...
+        $status = ServiceRequestStatus::factory()->state([
+            'classification' => $classification,
+        ]);
+
+        return [
+            'service_request_id' => ServiceRequest::factory(),
+            'service_request_status_id' => $status,
+            'classification' => $classification,
+            'started_at' => now(),
+        ];
```
Orrison: We should almost never have code outside of the return array in a factory, you should work with the values between themselves by using the attribute accessor function like below.

So you would define classification first, then, define the `service_request_status_id` as a function based on it. return the factory with the proper classification set as the state.

Image

app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
```diff
@@ -0,0 +1,182 @@
...
+use App\Features\SlaWaitingExclusionFeature;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Query\Builder;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Str;
+
+return new class () extends Migration {
+    // @phpstan-ignore Common.multipleMigrationChangesNotWrappedInTransaction
```
Orrison: We shouldn't be ignoring this and this should be all wrapped in a transaction, that way if any part of it fails it all rolls back.

app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
```diff
@@ -0,0 +1,182 @@
...
+    public function up(): void
+    {
+        if (! Schema::hasTable('service_request_status_periods')) {
+            return;
+        }
```
Orrison: No migration should ever have a conditional like this. These migrations are sources of truth and they NEED to run properly and in the right order, if this table didn't exist when this migration ran then we would want it to fail so we could figure out what we did wrong.

app-modules/service-management/src/Actions/RecordReclassifiedServiceRequestStatusPeriods.php
```diff
@@ -0,0 +1,75 @@
...
+        public CarbonInterface $reclassifiedAt,
+    ) {
+        $this->afterCommit = true;
+    }
+
+    public function handle(RecordServiceRequestStatusPeriod $recordServiceRequestStatusPeriod): void
+    {
+        ServiceRequest::query()
+            ->where('status_id', $this->status->getKey())
+            ->chunkById(200, function (Collection $serviceRequests) use ($recordServiceRequestStatusPeriod): void {
```
Orrison: I would prefer here and elsewhere in this PR, to do `eachById` it defaults to a good level of chunking underneath the hood.

app-modules/service-management/src/Jobs/RecordReclassifiedServiceRequestStatusPeriods.php
Orrison, line 1: This should just be a job, not a queue-able action. We prefer to keep queue things as jobs in the jobs folder.

app-modules/service-management/src/Actions/RecordReclassifiedServiceRequestStatusPeriods.php
```diff
@@ -0,0 +1,75 @@
...
+                foreach ($serviceRequests as $serviceRequest) {
+                    $serviceRequest->setRelation('status', $this->status);
```
Orrison: What is the purpose of this `setRelation`

app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -352,37 +373,83 @@ public function latestOutboundServiceRequestUpdate(): HasOne
...
+    }
+
+    /**
+     * Sum the seconds this request spent in any of the given classifications within [$start, $end].
+     *
+     * @param array<int, SystemServiceRequestClassification> $classifications
+     */
+    public function getExcludedSecondsBetween(CarbonInterface $start, CarbonInterface $end, array $classifications): int
+    {
+        $periods = $this->statusPeriods()
```
Orrison: This reverts the eager-loading fix from b193a24ad: calling `$this->statusPeriods()->orderBy(...)->get()` always fires a fresh query against the relation, ignoring any eager-loaded `statusPeriods` collection. That makes the `serviceRequest.statusPeriods` eager load added in `ServiceRequestFeedbackTable` a no-op, and `ListServiceRequestTableWidgets` never eager-loads `statusPeriods` at all - so every SLA column render (state + tooltip, x2 methods) fires a fresh `statusPeriods` query per row once the flag is active. Suggest reading the loaded relation (`$this->statusPeriods`) and sorting the collection in memory instead of querying the relation builder here, like b193a24ad originally did.

app-modules/report/src/Filament/Widgets/ServiceRequestFeedbackTable.php
```diff
@@ -92,6 +92,7 @@ public function table(Table $table): Table
                         'serviceRequest.priority.sla',
                         'serviceRequest.respondent',
                         'serviceRequest.assignedTo.user',
+                        'serviceRequest.statusPeriods',
```
Orrison, line 95: This eager load doesn't actually help - `getExcludedSecondsBetween()` queries `$this->statusPeriods()` as a fresh relation query rather than reading this loaded collection, so it's still N+1 per row for `statusPeriods` (see comment on `ServiceRequest::getExcludedSecondsBetween`).

app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
```diff
@@ -0,0 +1,182 @@
...
+                        $previousStatusId = null;
+
+                        foreach ($periods as $period) {
+                            if ($period['status_id'] === $previousStatusId) {
+                                continue;
+                            }
+
+                            $previousStatusId = $period['status_id'];
+
+                            if (! array_key_exists($period['status_id'], $classifications)) {
```
Orrison: When a status is hard-deleted (not present in `$classifications`), this `continue` skips inserting the period entirely rather than inserting it with `classification = null` as described in the tech plan (hard-deleted status -> unknown classification, treated as active/non-excluded). Skipping the row instead of inserting a null-classification row causes the *previous* period in the ledger to silently extend over this span in `getExcludedSecondsBetween()`, misattributing that time to the previous period's classification instead of leaving it unclassified.

app-modules/service-management/database/migrations/2026_08_12_163559_tmp_backfill_service_request_status_periods.php
```diff
@@ -0,0 +1,182 @@
...
+
+use App\Features\SlaWaitingExclusionFeature;
+use Illuminate\Database\Migrations\Migration;
+use Illuminate\Database\Query\Builder;
+use Illuminate\Support\Collection;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+use Illuminate\Support\Str;
+
+return new class () extends Migration {
```
Orrison, line 44: I don't see any test coverage for this backfill migration - no `describe()` block in `tests/TenantMigrationTests.php`, which is the established pattern for testing data migrations like this (e.g. the contacts citext migration). This is the most history-dependent piece of new logic (multiple `resolveInitialStatusId` fallback paths, hard-deleted-status handling, drafts) and currently has zero automated coverage.

app-modules/service-management/src/Observers/ServiceRequestObserver.php
```diff
@@ -103,22 +107,39 @@ public function saving(ServiceRequest $serviceRequest): void
...
             $serviceRequest->status_updated_at = CarbonImmutable::now();
 
-            if (
-                $serviceRequest->status->classification === SystemServiceRequestClassification::Closed &&
-                is_null($serviceRequest->time_to_resolution)
-            ) {
-                $createdTime = $serviceRequest->created_at;
-                $currentTime = Carbon::now();
+            if ($serviceRequest->status->classification === SystemServiceRequestClassification::Closed) {
+                if (SlaWaitingExclusionFeature::active() && $serviceRequest->exists) {
```
Orrison, line 114: Minor: this ledger-aware recompute only runs when `$serviceRequest->exists`, so a service request created directly with a Closed status still falls through to the legacy (non-excluding) `time_to_resolution` calculation below, even with the flag active, since no status periods exist yet at `saving()` time for a brand-new record. Probably fine/necessary given the ordering of hooks, just flagging since it's a slightly different behavior than "recompute from the ledger on every transition to Closed."

app-modules/service-management/tests/Tenant/Models/ServiceRequestTest.php
```diff
@@ -108,3 +115,163 @@ function recentUpdateCreatorInfo(ServiceRequest $serviceRequest): string
...
+        ServiceRequest::query()->findOrFail($serviceRequest->getKey())->update(['status_id' => $inProgress->getKey()]);
+
+        $this->travelTo($start->addSeconds(180));
+        ServiceRequest::query()->findOrFail($serviceRequest->getKey())->update(['status_id' => $closed->getKey()]);
+
+        // The raw 180s exceeds the 150s SLA, but excluding the 50s waiting span brings it to 130s.
+        expect($serviceRequest->fresh()->getResolutionSlaComplianceStatus())->toBe(SlaComplianceStatus::Compliant);
+    });
+
+    it('excludes time spent closed before a reopen from the resolution seconds when the feature is active', function () {
```
Orrison, line 253: Minor: this covers `getResolutionSeconds()` for the reopen case, but I don't see a test asserting the persisted `time_to_resolution` column is correct after a reopen + re-close cycle (only the initial close is covered above).

Orrison (changes requested):
Some feedback, and one final one as well:

`app/Filament/Widgets/ListServiceRequestTableWidgets.php` isn't touched by this PR, so I couldn't attach this inline, but it's worth calling out: this widget never eager-loads `statusPeriods`, unlike `ServiceRequestFeedbackTable`. Once `SlaWaitingExclusionFeature` is active, `getResponseSlaComplianceStatus()`/`getResolutionSlaComplianceStatus()` (each called twice per row, for state + tooltip) will hit `getExcludedSecondsBetween()`, which queries `statusPeriods` fresh per call — an N+1 on this list for every row's SLA columns, on top of the same issue already flagged inline on `ServiceRequest::getExcludedSecondsBetween()`.

app-modules/service-management/src/Actions/RecordReclassifiedServiceRequestStatusPeriods.php
mittal-canyon: Every service request handled in this loop was just matched by the reclassified status, so they all share $this->status. Calling $serviceRequest->setRelation('status', $this->status) pre-hydrates the relation so the downstream period recording doesn't lazy-load the identical status row once per record — it's a deliberate N+1 avoidance, not a correctness requirement.

Orrison (commented):
One thing still outstanding on the N+1 feedback: `ServiceRequest::getExcludedSecondsBetween()` now reads the loaded `$this->statusPeriods` collection and sorts it in memory, and `ServiceRequestFeedbackTable` eager-loads `serviceRequest.statusPeriods`, so those paths are covered. But `ListServiceRequestTableWidgets` still does not eager-load `statusPeriods` in its `ServiceRequest::with([...])` query, so once `SlaWaitingExclusionFeature` is active the SLA Response and SLA Resolution columns (state + tooltip) will lazy-load `statusPeriods` per row. I am intentionally leaving this as-is for now and will revisit if it shows up in profiling. Separately, I am fine with `chunkById` staying in the tmp backfill migration rather than switching to `eachById`.

app-modules/service-management/src/Jobs/RecordReclassifiedServiceRequestStatusPeriods.php
```diff
@@ -0,0 +1,75 @@
...
+    public function __construct(
+        public ServiceRequestStatus $status,
+        public CarbonInterface $reclassifiedAt,
+    ) {
+        $this->afterCommit = true;
+    }
+
+    public function handle(RecordServiceRequestStatusPeriod $recordServiceRequestStatusPeriod): void
+    {
+        ServiceRequest::query()
```
Orrison, line 65: This uses `ServiceRequest::query()`, which applies the `excludeDrafts` global scope, so draft requests currently in the reclassified status are skipped here. That is inconsistent with the backfill migration (which includes drafts via the `DB` query builder) and with the `created()`/`saved()` write path (which records periods for drafts). If such a draft is later published and closed, its pre-reclassification span keeps the old classification and its SLA can be misattributed. Suggest `ServiceRequest::query()->withoutGlobalScope('excludeDrafts')` (or `withoutGlobalScopes()`).

Orrison, line 65: Because this runs async (afterCommit) and matches on the current `status_id`, the reclassification split is best-effort and can silently mis-record in a few edge cases: (1) a request that leaves the status between dispatch and job execution never gets a split, so its time in the status stays on the old classification; (2) only the `ServiceRequestStatus` is serialized, so on deserialize it reloads the *current* classification — two rapid reclassifications record the newest one, not the one at `reclassifiedAt`; (3) if a request re-enters the status via the normal write path before this job runs, the duplicate guard in `RecordServiceRequestStatusPeriod` (same status_id + classification) skips the split entirely. These may be acceptable given the async design, but they are currently untested.

app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -267,6 +268,26 @@ public function histories(): HasMany
...
+     */
+    public function statusPeriods(): HasMany
+    {
+        return $this->hasMany(ServiceRequestStatusPeriod::class);
+    }
+
+    /**
+     * @return HasOne<ServiceRequestStatusPeriod, $this>
+     */
+    public function currentStatusPeriod(): HasOne
```
Orrison: `currentStatusPeriod()` is defined here but I could not find any usage of it anywhere in the codebase. Either it is dead code that should be dropped, or an intended wire-up (e.g. for display/eager-loading) that was missed.

app-modules/service-management/src/Models/ServiceRequest.php
```diff
@@ -352,37 +373,84 @@ public function latestOutboundServiceRequestUpdate(): HasOne
...
+
+            $periodStart = $period->started_at;
+            $nextPeriod = $periods->get($index + 1);
+            $periodEnd = $nextPeriod ? $nextPeriod->started_at : $ongoingEnd;
+
+            $overlapStart = $periodStart->greaterThan($start) ? $periodStart : $start;
+            $overlapEnd = $periodEnd->lessThan($end) ? $periodEnd : $end;
+
+            if ($overlapEnd->greaterThan($overlapStart)) {
+                $excludedSeconds += (int) round($overlapStart->diffInSeconds($overlapEnd));
```
Orrison: Minor: `getResolutionSeconds()` rounds the whole window once, while `getExcludedSecondsBetween()` rounds *each* excluded span independently before summing here. With sub-second timestamps this can produce an off-by-one versus rounding once. Negligible at second granularity, just flagging.

app-modules/service-management/src/Observers/ServiceRequestObserver.php
Orrison, line 114: This ledger-aware recompute only runs when `$serviceRequest->exists`, so a request created directly in a Closed status falls through to the legacy (non-excluding) calculation even with the flag active, since no status periods exist yet at `saving()` time. Impact is low (a brand-new record has ~0 elapsed and no prior waiting time), but the behavior differs from "recompute from the ledger on every transition to Closed" and is untested.

app-modules/service-management/tests/Tenant/Jobs/RecordReclassifiedServiceRequestStatusPeriodsTest.php
```diff
@@ -0,0 +1,74 @@
...
+
+use AidingApp\ServiceManagement\Actions\RecordServiceRequestStatusPeriod;
+use AidingApp\ServiceManagement\Enums\SystemServiceRequestClassification;
+use AidingApp\ServiceManagement\Jobs\RecordReclassifiedServiceRequestStatusPeriods;
+use AidingApp\ServiceManagement\Models\ServiceRequest;
+use AidingApp\ServiceManagement\Models\ServiceRequestStatus;
+use Carbon\CarbonImmutable;
+use Illuminate\Support\Facades\Notification;
+
+it('appends a status period for every request currently in the reclassified status', function () {
```
Orrison, line 45: This test asserts a period row is appended, but there is no coverage that the reclassification actually *splits* the SLA math — i.e. that `getExcludedSecondsBetween()` attributes the span before the reclassification to the old classification and the span after to the new one. Worth adding, since the split is the whole point of the reclassification path.

Orrison (changes requested):
@mittal-canyon can you please fix the merge conflict here

Orrison (approved):

PR #1365, opened 2026-08-14
[AIDAPP-1373]: Improve the user experience when creating or editing service monitors

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1373

### Technical Description

- Refactor service monitoring forms to use structured sections for better organization and layout

No

### Cleanup Tasks

No

---


payal-canyon (changes requested):
I understand that the ticket doesn’t specifically mention the View page, but I think we should make the same changes there as we did for the Create and Edit pages. So, let’s include the View page as well. Thanks!

payal-canyon (approved):

PR #1369, opened 2026-08-18
[AIDAPP-1336]: Introduce support for organization defined landing page tab in the portal for the knowledge base

mxm1070:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1336

### Technical Description

Adds a `KnowledgeBaseCategoryTabOrder` enum defining the 6 possible display orders of the self-service knowledge base portal's category-browsing tabs (All Articles, Featured, Most Viewed). A new `KnowledgeBasePortalSettings` (spatie/laravel-settings) class stores the organization's chosen order, managed via a new `ManageKnowledgeBasePortalSettings` Filament settings page in the Knowledge Management cluster (added last in nav order). The portal's bootstrap config and `KnowledgeManagementPortalController` now expose `category_tab_order`, and the Vue `ViewCategory` page renders/orders the category tabs accordingly, defaulting to the first tab in the configured order as the active tab. Gated behind the `KnowledgeBaseCategoryTabOrderFeature` flag for zero-downtime rollout.

No.

### Cleanup Tasks

Cleanup task added: `.cleanup-tasks/2026_08_18_knowledge_base_category_tab_order_feature.md` (tracks removal of `App\Features\KnowledgeBaseCategoryTabOrderFeature` post-deploy).

---

### DIFF
--- a/app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
+++ b/app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
@@ -53,7 +53,7 @@ class ManageKnowledgeBasePortalSettings extends SettingsPage
 {
     protected static ?string $navigationLabel = 'Portal';
 
-    protected static ?int $navigationSort = 5;
+    protected static ?int $navigationSort = 50;
 
     protected static string $settings = KnowledgeBasePortalSettings::class;
 
@@ -81,7 +81,7 @@ class ManageKnowledgeBasePortalSettings extends SettingsPage
                     ->schema([
                         Select::make('category_tab_order')
                             ->label('Category Tabs')
-                            ->helperText('Choose the order in which the All Articles, Featured, and Most Viewed tabs are displayed on the knowledge base category pages.')
+                            ->aboveContent('When customers browse a category or subcategory in the self-service portal, tabs organize the available articles. Use the options below to choose the order in which these tabs appear. The first tab in the selected order will be displayed by default when the page loads.')
                             ->options(KnowledgeBaseCategoryTabOrder::class)
                             ->enum(KnowledgeBaseCategoryTabOrder::class)
                             ->required(),
--- a/app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
+++ b/app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
@@ -1,0 +1,118 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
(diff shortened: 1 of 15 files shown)


app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
```diff
@@ -0,0 +1,118 @@
...
+    public function form(Schema $schema): Schema
+    {
+        return $schema
+            ->components([
+                Section::make('Configuration Options')
+                    ->schema([
+                        Select::make('category_tab_order')
+                            ->label('Category Tabs')
+                            ->aboveContent('When customers browse a category or subcategory in the self-service portal, tabs organize the available articles. Use the options below to choose the order in which these tabs appear. The first tab in the selected order will be displayed by default when the page loads.')
+                            ->options(KnowledgeBaseCategoryTabOrder::class)
```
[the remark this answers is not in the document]
mxm1070: `Select` extends `Field` (`vendor/filament/forms/src/Components/Field.php`), which defines `aboveContent()` at line 173, so this method does exist and resolves correctly for this field — it's just not yet used elsewhere in the repo, which is likely why the search didn't surface it. The settings page renders correctly; verified live via a browser screenshot and the passing `ManageKnowledgeBasePortalSettingsTest`. Leaving the code as-is.

app-modules/knowledge-base/src/Filament/Pages/ManageKnowledgeBasePortalSettings.php
```diff
@@ -0,0 +1,118 @@
...
+    public static function canAccess(): bool
+    {
+        if (! Gate::check(Feature::KnowledgeManagement->getGateName())) {
+            return false;
+        }
+
+        /** @var User $user */
+        $user = auth()->user();
+
+        return $user->can(['settings.view-any']);
```
payal-canyon: Could we gate `canAccess()` with the feature flag so the page isn’t accessible before the migration runs? Otherwise, saving the form before the migration could result in errors.

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/KnowledgeManagementPortalControllerTest.php
```diff
@@ -0,0 +1,76 @@
...
+
+    $knowledgeBasePortalSettings = app(KnowledgeBasePortalSettings::class);
+    $knowledgeBasePortalSettings->category_tab_order = KnowledgeBaseCategoryTabOrder::MostViewedFeaturedAllArticles;
+    $knowledgeBasePortalSettings->save();
+
+    $response = getJson(route('api.portal.define'));
+
+    $response->assertOk();
+
+    expect($response->json('category_tab_order'))->toBe(['most-viewed', 'featured', 'all-articles']);
```
payal-canyon: Could we add one more test to verify the default `category_tab_order` is `All Articles, Featured, Most Viewed`?

payal-canyon (commented):
Hey @mxm1070 ,
Looks good to me! Could we just confirm with @Orrison or Product whether the search tabs should also follow the configured order?

Currently, the category tabs use the configured order, but the search results switch to their own `All Articles, Featured, Most Viewed` order. Since both appear on the same screen, the change in order may look inconsistent to users.

If search is intentionally independent, then I think it’s fine. WDYT?

Orrison:
> Hey @mxm1070 , Looks good to me! Could we just confirm with @Orrison or Product whether the search tabs should also follow the configured order?
> 
> Currently, the category tabs use the configured order, but the search results switch to their own `All Articles, Featured, Most Viewed` order. Since both appear on the same screen, the change in order may look inconsistent to users.
> 
> If search is intentionally independent, then I think it’s fine. WDYT?

@payal-canyon excellent call out. I think this was just missed / not thought about. @mxm1070 if it isn't too much more I think we should handle the search tabs as well here please.

mxm1070:
Good catch — the search-results tabs (both on the Home page and the category page) were still driven by a separate hardcoded `All Articles, Featured, Most Viewed` array, independent of the new `category_tab_order` setting. Pushed a fix (`b805f2e7f`) that derives the search filter tabs from the same config-ordered source already used for category browsing, so both surfaces now stay consistent. Verified via a live browser check with the setting changed to a non-default order — search tabs, category-browsing tabs, and the default active tab all now follow it on both pages.

payal-canyon (approved):

PR #1373, opened 2026-08-19
[AIDAPP-1394]: Enhance the user interface for projects by adding a tabbed display for access, pipelines, and files

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1394

### Technical Description

> Enhance the user interface for projects by adding a tabbed display for access, pipelines, and files

> No

### Cleanup Tasks

> No

---


payal-canyon (changes requested):
Could you please resolve the merge conflicts when you get a chance? Thanks!

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +58,62 @@
         @livewire(ProjectStatsWidget::class, ['record' => $record])
     @endif
 
-    @if (ProjectAccessWidget::canView())
-        @livewire(ProjectAccessWidget::class, ['record' => $record])
-    @endif
+    @if ($canViewAccess || $canViewPipelines || $canViewFiles)
+        <div x-data="{ tab: '{{ $defaultTab }}' }" class="space-y-6">
```
payal-canyon, line 62: Could we persist the selected `tab` across page navigation and refreshes?

Currently, tab only lives in Alpine state, so it resets to the default tab whenever the page reloads. For example, a user can open the Pipelines tab, navigate to Kanban, and when they return, they are taken back to Access.

Could we follow the existing pattern from `ViewServiceRequest` and store the selected tab in a `#[Url]` Livewire property? That would preserve the selection on navigation/refresh and also allow linking directly to a specific tab.

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +58,62 @@
...
+                            :alpine-active="'tab === \'files\''"
+                            icon="heroicon-m-paper-clip"
+                        >
+                            Files
+                        </x-filament::tabs.item>
+                    @endif
+                </x-filament::tabs>
+            </div>
+
+            @if ($canViewAccess)
```
payal-canyon, line 100: Could we avoid rendering all three widgets on every page load?

Currently, `x-show` only hides the inactive tabs visually, so all three widgets are still mounted and their queries run even when the user only views one tab.

Since we already need to move the selected tab to a Livewire property, could we follow the `ViewServiceRequest` pattern here as well? That would allow inactive tabs to avoid rendering, so this could address both the tab persistence and unnecessary rendering concerns together.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -95,6 +95,21 @@ function loginAsUserWithProjectViewPermissions(): User
...
+it('renders the project dashboard widgets in access, pipeline, and file tabs', function () {
+    loginAsUserWithProjectViewPermissions();
+
+    $project = Project::factory()->create();
+
+    get(ViewProject::getUrl([
+        'record' => $project->getRouteKey(),
+    ]))
+        ->assertSuccessful()
+        ->assertSee('x-data="{ tab: \'access\' }"', false)
```
payal-canyon, line 107: Could we replace this with a behavioural assertion?

Currently, the test asserts the exact Blade/Alpine output, which makes it dependent on template formatting rather than actual behaviour. A small formatting change could break the test without changing anything functionally.

If we expose the selected tab as a Livewire property, as suggested above, we could instead assert the default behaviour directly:
```php
livewire(ViewProject::class, ['record' => $project])
    ->assertSet('tab', 'access');
```

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -95,6 +95,21 @@ function loginAsUserWithProjectViewPermissions(): User
...
+        ->assertSee('Access')
+        ->assertSee('Pipelines')
```
payal-canyon, line 109: Could we make these assertions more specific to the actual tab markup?

`assertSee('Access')` and `assertSee('Files')` can pass even if the tabs are removed, since those labels are rendered elsewhere on the page. Only Pipelines is unique to the tab markup.

It might be better to assert the tab behaviour directly, especially if we move to the Livewire `tab` property suggested above.

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -39,9 +39,16 @@
...
 
     $record = $this->getRecord();
+    $canViewAccess = ProjectAccessWidget::canView();
+    $canViewPipelines = auth()
+        ->user()
+        ?->can('viewAny', [Pipeline::class, $record]);
+    $canViewFiles = auth()
+        ->user()
+        ?->can('viewAny', [ProjectFile::class, $record]);
+    $defaultTab = $canViewAccess ? 'access' : ($canViewPipelines ? 'pipelines' : 'files');
```
payal-canyon, line 51: Could we add tests for the new per-tab visibility and `$defaultTab` fallback logic?

The current test only covers the all-permissions case, so it doesn't exercise the branches where a tab is unavailable. For example:

- A user with `project.*.view` but without `project.view-any` should default to Pipelines because the Access tab isn't available.
- When the `Feature::ProjectManagement` gate is inactive, the Files tab should not be available.

These would cover the new conditional logic introduced by this PR.

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +58,62 @@
...
-        @livewire(ProjectAccessWidget::class, ['record' => $record])
-    @endif
+    @if ($canViewAccess || $canViewPipelines || $canViewFiles)
+        <div x-data="{ tab: '{{ $defaultTab }}' }" class="space-y-6">
+            <div class="flex justify-center">
+                <x-filament::tabs>
+                    @if ($canViewAccess)
+                        <x-filament::tabs.item
+                            tag="button"
+                            x-on:click="tab = 'access'"
```
payal-canyon, line 68: Could we consider introducing a `ProjectDashboardTab` backed enum for `access`, `pipelines`, and `files`?

A typo in any of the current string literals would silently prevent the corresponding content from rendering, and neither PHPStan nor the current tests would catch it. We already have a similar pattern with `ServiceRequestTab`, including a `default()` method, so this could also give `$defaultTab` a single typed source of truth.

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +58,62 @@
...
+                </div>
+            @endif
 
-    @if (auth()->user()?->can('viewAny', [ProjectFile::class, $record]))
-        @livewire(ProjectFilesWidget::class, ['record' => $record])
+            @if ($canViewFiles)
+                <div x-show="tab === 'files'" x-cloak>
+                    @livewire(ProjectFilesWidget::class, ['record' => $record], key('project-files-' . $record->getKey()))
+                </div>
+            @endif
```
payal-canyon, line 116: Could we move the tab permission logic out of `view-project.blade.php` and into `ViewProject.php`?

With three gate checks and the nested `$defaultTab` logic, it would be easier to keep this logic in the page class as methods or a `#[Computed]` property. It would also make the behaviour directly testable from the Livewire tests instead of relying on rendered HTML, which would help cover the cases mentioned above. This also fits the existing `ViewServiceRequest` pattern.

payal-canyon (changes requested):

amit-canyon:
@payal-canyon  I’m going to close this PR and create a new one with the same changes. This is because the current PR contains some unnecessary commits from stacked branches that might cause issues. I’m not completely sure whether they will, so to be on the safe side, I’ve created a new PR with only the required changes.

Thanks!

cc: @Orrison

PR #1382, opened 2026-08-21
[AIDAPP-1394]: Enhance the user interface for projects by adding a tabbed display for access, pipelines, and files

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1394

### Technical Description

> Enhance the user interface for projects by adding a tabbed display for access, pipelines, and files

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
+++ b/app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
@@ -32,6 +32,7 @@
     </COPYRIGHT>
 --}}
 @php
+    use AidingApp\Project\Enums\ProjectTab;
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectFilesWidget;
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectStatsWidget;
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectAccessWidget;
@@ -51,15 +52,30 @@
         @livewire(ProjectStatsWidget::class, ['record' => $record])
     @endif
 
-    @if (ProjectAccessWidget::canView())
+    <div class="flex justify-center">
+        <x-filament::tabs>
+            @foreach (ProjectTab::cases() as $projectTab)
+                @if ($projectTab->canView($record))
+                    <x-filament::tabs.item
+                        :active="$tab === $projectTab->value"
+                        wire:click="$set('tab', '{{ $projectTab->value }}')"
+                    >
+                        {{ $projectTab->getLabel() }}
+                    </x-filament::tabs.item>
+                @endif
+            @endforeach
+        </x-filament::tabs>
+    </div>
+
+    @if ($tab === ProjectTab::Access->value && ProjectAccessWidget::canView())
         @livewire(ProjectAccessWidget::class, ['record' => $record])
     @endif
 
-    @if (auth()->user()?->can('viewAny', [Pipeline::class, $record]))
+    @if ($tab === ProjectTab::Pipelines->value && auth()->user()?->can('viewAny', [Pipeline::class, $record]))
         @livewire(ProjectWorkPipelineWidget::class, ['record' => $record])
     @endif
 
-    @if (auth()->user()?->can('viewAny', [ProjectFile::class, $record]))
+    @if ($tab === ProjectTab::Files->value && auth()->user()?->can('viewAny', [ProjectFile::class, $record]))
         @livewire(ProjectFilesWidget::class, ['record' => $record])
     @endif
 </x-filament-panels::page>
--- a/app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
+++ b/app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
@@ -38,8 +38,6 @@
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectAccessWidget;
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectDashboardHeaderWidget;
     use AidingApp\Project\Filament\Resources\Projects\Widgets\ProjectWorkPipelineWidget;
-    use AidingApp\Project\Models\Pipeline;
-    use AidingApp\Project\Models\ProjectFile;
     use AidingApp\Project\Models\ProjectMilestone;
 
     $record = $this->getRecord();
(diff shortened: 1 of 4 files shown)


app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +52,30 @@
...
+            @endforeach
+        </x-filament::tabs>
+    </div>
+
+    @if ($tab === ProjectTab::Access->value && ProjectAccessWidget::canView())
         @livewire(ProjectAccessWidget::class, ['record' => $record])
     @endif
 
-    @if (auth()->user()?->can('viewAny', [Pipeline::class, $record]))
+    @if ($tab === ProjectTab::Pipelines->value &&auth()->user()?->can('viewAny', [Pipeline::class, $record]))
```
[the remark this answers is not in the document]
amit-canyon: @Orrison  Formatter removed the space after &&.

app-modules/project/src/Filament/Resources/Projects/Pages/ViewProject.php
```diff
@@ -51,6 +55,38 @@ class ViewProject extends ViewRecord
 
     protected static ?string $navigationLabel = 'View';
 
+    #[Url()]
+    public ?string $tab = null;
+
+    public function mountCanAuthorizeResourceAccess(): void
```
payal-canyon: Could we remove this `mountCanAuthorizeResourceAccess()` override?

Filament authorizes resource access on both `mount` and `hydrate`. Since only the `mount` hook is overridden here, the page uses different authorization rules on the initial load and subsequent Livewire requests: a user with `project.*.view` but without `project.view-any` can load the page, but will hit a 403 on their first interaction.

The new `it('defaults to pipelines when the user cannot view the access tab')` test only mounts the page, so it doesn't catch this.

If the override was added only for that test, could we remove it and grant `project.view-any` instead, matching the existing project view permission helper? If users without `project.view-any` should genuinely access this page, it would be better to update the resource-level authorization so both mount and subsequent requests follow the same rule.

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +52,30 @@
...
         @livewire(ProjectWorkPipelineWidget::class, ['record' => $record])
     @endif
 
-    @if (auth()->user()?->can('viewAny', [ProjectFile::class, $record]))
+    @if ($tab === ProjectTab::Files->value &&auth()->user()?->can('viewAny', [ProjectFile::class, $record]))
```
payal-canyon: Could we use `ProjectTab::canView()` for the panel visibility checks as well?

These three `@if` conditions duplicate the logic already defined in `ProjectTab::canView()` and used when rendering the tabs. If one is updated without the other, we could end up with a visible tab whose panel does not render, or a panel with no corresponding tab.

For example:
`@if ($tab === ProjectTab::Access->value && ProjectTab::Access->canView($record))`

Alternatively, we could render the tabs and panels from the same `@foreach`, so the visibility is determined in one place. We should still keep a server-side visibility check for each panel; this would just avoid duplicating the `authorization` logic.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -207,6 +209,103 @@ function loginAsUserWithProjectViewPermissions(): User
...
+        $settings->save();
+
+        asSuperAdmin();
+
+        $project = Project::factory()->create();
+
+        expect(ProjectTab::Files->canView($project))->toBeFalse();
+    });
+
+    it('selects the access tab', function () {
```
payal-canyon: Could we also assert that only the active tab's widget is rendered?

The current `assertSet('tab', …)` and `assertSeeHtml("wire:click=…")` assertions only confirm the selected tab and tab bar. They don't verify the main behaviour: that inactive tab widgets are not rendered.

For example, after selecting a tab, we could assert its widget is present and the others are not:
```php
->set('tab', ProjectTab::Pipelines->value)
->assertSeeLivewire(ProjectWorkPipelineWidget::class)
->assertDontSeeLivewire(ProjectFilesWidget::class);
```
The three similar `it('selects the … tab')` cases could also be combined using a `ProjectTab::cases()` dataset.

payal-canyon (changes requested):

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
```diff
@@ -51,15 +50,36 @@
...
+                    >
+                        {{ $projectTab->getLabel() }}
+                    </x-filament::tabs.item>
+                @endif
+            @endforeach
+        </x-filament::tabs>
+    </div>
+
+    @if ($tab === ProjectTab::Access->value && ProjectTab::Access->canView($record))
+        <div data-project-tab-panel="{{ ProjectTab::Access->value }}">
```
payal-canyon, line 69: Could we remove the `data-project-tab-panel` wrapper `<div>`s and assert against the Livewire widget classes instead?

These wrappers don't appear to serve any runtime purpose; they were added so the tests can use `assertSeeHtml()` / `assertDontSeeHtml()`. Livewire already provides `assertSeeLivewire()` and `assertDontSeeLivewire()` for exactly this, and `ViewServiceRequestTest` uses that pattern for a very similar tabbed `ViewRecord` page.

For example:
```php
->assertSeeLivewire(ProjectWorkPipelineWidget::class)
->assertDontSeeLivewire(ProjectFilesWidget::class)
```
Could we drop the three wrapper divs and update the tests to assert directly against the widgets? It would also be worth adding `->assertSuccessful()` to these cases, as the sibling tests do.

app-modules/project/src/Filament/Resources/Projects/Pages/ViewProject.php
```diff
@@ -51,6 +53,31 @@ class ViewProject extends ViewRecord
...
+        $availableTab = collect(ProjectTab::cases())
+            ->first(fn (ProjectTab $tab): bool => $tab->canView($project));
+
+        if ($availableTab === null) {
+            $this->tab = null;
+
+            return;
+        }
+
+        $requestedTab = $requestedTab = is_string($this->tab) ? ProjectTab::tryFrom($this->tab) : null;
```
payal-canyon, line 75: Could we remove the duplicate assignment here?

payal-canyon (changes requested):

app-modules/project/resources/views/filament/resources/projects/view-project.blade.php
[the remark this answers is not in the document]
payal-canyon: This is resolved now - both conditions were rewritten to use `ProjectTab::…->canView($record)`, which fixed the spacing and removed the duplicated permission check.

payal-canyon (approved):

PR #1383, opened 2026-08-21
[AIDAPP-1415]: Enhance the information presented in the timeline for customer updates when viewing a service request

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1415

### Technical Description

- Refactor service request timeline components for improved readability and consistency

No

### Cleanup Tasks

No

---


app-modules/service-management/resources/views/components/service-request-update-timeline-item.blade.php
```diff
@@ -32,24 +32,38 @@
...
+        $createdBy instanceof User => 'Agent',
+        default => 'Customer',
+    };
+
+    $actorName = match (true) {
+        $createdBy instanceof ServiceRequest => 'AI',
+        $createdBy instanceof User => $createdBy->name,
+        $createdBy instanceof Contact => $createdBy->full_name,
+        default => 'Unknown',
+    };
```
rebecca-canyon, line 55: I think that these should match in behavior. If the `createdBy` isn't a `ServiceRequest` or a `User`, then it should presumably be a `Contact`; but if something went wrong and it's not, the `actorRole` probably shouldn't still say `Customer`. So I think checking that `createdBy` is an instance of `Contact` in the `actorRole` match statement and adding a default of `Unknown` makes sense here.

app-modules/service-management/resources/views/components/service-request-assignment-timeline-item.blade.php
```diff
@@ -71,17 +70,14 @@
 
 <div>
     <div class="flex flex-row justify-between">
-        <h3 class="mb-1 flex items-center text-lg font-semibold text-gray-500 dark:text-gray-100">
-            <a
-                class="ml-2 flex space-x-2 font-medium underline"
-                href="{{ ServiceRequestResource::getUrl('view', ['record' => $serviceRequest]) }}"
-            >
+        <h3 class="mb-1 flex items-center text-base font-semibold text-gray-950 dark:text-white">
```
rebecca-canyon: Are there other Filament classes we can be using here and in other locations? Similar to the `fi-ta-text` class you use later in this file?

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceRequests/Pages/ViewServiceRequestTest.php
```diff
@@ -457,6 +460,78 @@ function serviceRequestWithFeedbackCollection(
...
+        $customerName = $customer->full_name;
+        $agentName = $agent->name;
+
+        livewire(ViewServiceRequest::class, ['record' => $serviceRequest->getRouteKey()])
+            ->set('tab', ServiceRequestTab::Timeline->value)
+            ->assertSuccessful()
+            ->assertSeeLivewire(TimelineList::class)
+            ->assertSee("{$customerName} (Customer) posted the following update on {$customerPostedAt}:")
+            ->assertSee("{$agentName} (Agent) posted the following update on {$agentPostedAt}:")
+            ->assertSee("AI (AI) posted the following update on {$aiPostedAt}:")
```
rebecca-canyon: @Orrison It looks like the ticket does request this structure of "AI (AI)" for AI responses in a ticket, but looking at it now here it feels a little awkward. Just wanted to get your thoughts on if we should proceed with this as is, or if we should format the "name (role)" differently for AI responses?

rebecca-canyon (changes requested):

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceRequests/Pages/ViewServiceRequestTest.php
Orrison: @rebecca-canyon and @monali4-canyon ah, hmm. Let's do "Assistant (AI)"

monali4-canyon: @Orrison @rebecca-canyon I have changed it

rebecca-canyon (approved):

PR #1384, opened 2026-08-24
[AIDAPP-1432]: Optimize based on prior release: .cleanup-tasks/2026_08_11_pipeline_entry_milestone_feature.md

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1432

### Technical Description

- Optimize based on prior release: .cleanup-tasks/2026_08_11_pipeline_entry_milestone_feature.md

No

### Cleanup Tasks

No

---


app-modules/project/src/Models/PipelineEntryMilestone.php
```diff
@@ -1,67 +0,0 @@
-<?php
```
mxm1070, line 1: Now that `PipelineEntryMilestone.php` is being removed, nothing in the codebase reads or writes the `pipeline_entry_milestones` table anymore (checked model/relation references, raw queries, and pivot access via `ProjectMilestone`'s inverse relation — all clear). Since the ticket's cleanup doc already covers the flag and the tmp migration, it'd be worth rolling the table drop into this same PR rather than leaving it orphaned in the schema.

Would you mind adding a migration like this?

```php
<?php

use Illuminate\Database\Migrations\Migration;
use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;

return new class () extends Migration {
    public function up(): void
    {
        Schema::dropIfExists('pipeline_entry_milestones');
    }

    public function down(): void
    {
        Schema::create('pipeline_entry_milestones', function (Blueprint $table) {
            $table->uuid('id')->primary();
            $table->foreignUuid('pipeline_entry_id')->constrained()->cascadeOnDelete();
            $table->foreignUuid('project_milestone_id')->constrained()->cascadeOnDelete();
            $table->timestamps();
        });
    }
};
```

mxm1070 (changes requested):
Solid cleanup — verified all three items from the ticket's cleanup doc are handled (feature flag, tmp migration, TODO markers), no dangling references to the removed `milestones()` relation/model, and the flagged infolist issue is already fixed in this branch. One thing I'd like to see added before merging: a migration to drop the now-orphaned `pipeline_entry_milestones` table.

app-modules/project/src/Models/PipelineEntryMilestone.php
monali4-canyon: I think for schema hygiene is required to add it. so I have added this migration.

mxm1070 (approved):

PR #1394, opened 2026-08-24
[AIDAPP-1410]: Enhance project pipeline tasks kanban view by enhancing card navigation

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1410

### Technical Description

> Remove three dots icon from kanban cards and put trash icon instead to delete tasks.
> On the click on title of the card, it should go into Edit or View mode depending on whether you are a manager or auditor respectively.

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/project/resources/views/components/entry-card.blade.php
+++ b/app-modules/project/resources/views/components/entry-card.blade.php
@@ -76,6 +76,8 @@
     }
 
     $dueTooltip = $entry->due?->format('M j, Y g:i A');
+
+    $canUpdatePipeline = auth()->user()->can('update', $pipeline);
 @endphp
 
 <div
@@ -86,46 +88,24 @@
     wire:key="pipeline-entry-{{ $entry->getKey() }}"
 >
     <div class="flex items-start justify-between gap-2">
-        <div class="min-w-0 break-words text-base font-semibold text-gray-900 dark:text-white">
+        <button
+            type="button"
+            class="min-w-0 wrap-break-word text-left text-base font-semibold text-gray-900 hover:underline dark:text-white"
+            wire:click="mountAction('{{ $canUpdatePipeline ? 'editPipelineEntry' : 'viewPipelineEntry' }}', { entry: '{{ $entry->getKey() }}' })"
+        >
             {{ $entry->name }}
-        </div>
+        </button>
 
-        <x-filament::dropdown placement="bottom-end">
-            <x-slot name="trigger">
-                <x-filament::icon-button
-                    class="shrink-0"
-                    icon="heroicon-m-ellipsis-horizontal"
-                    label="Actions"
-                    size="xs"
-                />
-            </x-slot>
-
-            <x-filament::dropdown.list>
-                <x-filament::dropdown.list.item
-                    icon="heroicon-m-eye"
-                    wire:click="mountAction('viewPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
-                >
-                    View
-                </x-filament::dropdown.list.item>
-
-                @can('update', $pipeline)
-                    <x-filament::dropdown.list.item
-                        icon="heroicon-m-pencil"
-                        wire:click="mountAction('editPipelineEntry', { entry: '{{ $entry->getKey() }}' })"
-                    >
-                        Edit
-                    </x-filament::dropdown.list.item>
-
-                    <x-filament::dropdown.list.item
-                        icon="heroicon-m-trash"
-                        color="danger"
-                        wire:click="mountAction('removePipelineEntry', { entry: '{{ $entry->getKey() }}' })"
-                    >
-                        Remove
-                    </x-filament::dropdown.list.item>
-                @endcan
-            </x-filament::dropdown.list>
-        </x-filament::dropdown>
+        @can('update', $pipeline)
+            <x-filament::icon-button
+                class="shrink-0"
+                icon="heroicon-m-trash"
+                color="danger"
+                label="Remove"
+                size="xs"
+                wire:click="mountAction('removePipelineEntry', { entry: '{{ $entry->getKey() }}' })"
+            />
+        @endcan
     </div>
 
     <hr class="my-3 border-gray-200 dark:border-gray-700" />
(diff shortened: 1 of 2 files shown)


app-modules/project/resources/views/components/entry-card.blade.php
```diff
@@ -77,6 +77,10 @@
     }
 
     $dueTooltip = $entry->due?->format('M j, Y g:i A');
+
+    $canUpdatePipeline = auth()
+        ->user()
+        ->can('update', $pipeline);
```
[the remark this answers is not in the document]
jayushi-canyon: Not a concern, since this component only renders  inside authenticated Filament panel pages

app-modules/project/resources/views/components/entry-card.blade.php
```diff
@@ -77,6 +77,10 @@
     }
 
     $dueTooltip = $entry->due?->format('M j, Y g:i A');
+
+    $canUpdatePipeline = auth()
```
payal-canyon, line 80: Could we reuse `$canUpdatePipeline` for the trash button as well?

The update policy is currently evaluated twice per card: once when `$canUpdatePipeline` is computed and again through `@can('update', $pipeline)`. Since both checks use the same user and pipeline, the result will be the same.

This policy is relatively expensive for non-super-admin users, as it delegates through `ProjectPolicy::update`, which performs additional relationship queries. On a board with many cards, evaluating the same authorization check twice per card adds unnecessary database work.

Could we replace the `@can('update', $pipeline)` with:
```php
@if ($canUpdatePipeline)
```
That reuses the value already computed, avoids the duplicate authorization check, and keeps the card consistent by using a single approach for the same permission.

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -732,13 +732,16 @@
         ->assertActionHidden('removePipelineEntry');
 });
 
-it('hides the edit action without pipeline update permission', function () {
+it('mounts the edit pipeline entry action from the card title with pipeline update permission', function () {
```
payal-canyon, line 735: Could we add `->assertActionHidden('editPipelineEntry')` back to the no-update-permission test?

The previous test explicitly covered the `->authorize()` guard on `editPipelineEntry`, but the rewritten tests now only verify that the edit button isn't rendered. That doesn't verify that a user without update permission is prevented from invoking the Livewire action directly.

The other actions (`addEntry`, `viewPipelineEntry`, and `removePipelineEntry`) still have tests covering their authorization guards, so `editPipelineEntry` is currently the only one without equivalent coverage.

Could we keep the new markup assertions and also add:
```php
->assertActionHidden('editPipelineEntry')
```
This preserves the existing authorization coverage while testing the new UI behaviour as well.

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -748,44 +751,40 @@
...
-    $user->givePermissionTo('pipeline.*.view');
     $user->refresh();
 
-    expect($user->can('update', $pipeline))->toBeFalse();
+    expect($user->can('update', $pipeline))->toBeTrue();
 
     livewire(PipelineEntryKanban::class, ['pipeline' => $pipeline])
-        ->assertActionVisible('viewPipelineEntry')
-        ->assertActionHidden('editPipelineEntry');
+        ->assertSee("mountAction('editPipelineEntry', { entry: '{$entry->getKey()}' })", false)
```
payal-canyon: Could we avoid asserting the exact `wire:click` attribute here?

`->assertSee("mountAction('editPipelineEntry', { entry: '{$entry->getKey()}' })", false)`

This is coupled to the exact whitespace/formatting of the Blade attribute, so a formatter-only change could break the test without changing any behaviour. It also doesn't really verify the action itself.

The same test file already uses the more idiomatic pattern:
```php
->assertActionHidden('addEntry')
->assertDontSeeHtml('addEntry');
```
Could we follow that approach here - use `assertActionVisible` / `assertActionHidden` for the `authorization` behaviour and `assertSeeHtml` / `assertDontSeeHtml` only to verify the relevant action is rendered?

If we specifically want to verify that the title's action targets the correct entry, `mountAction(...)->assertActionMounted(...)` would be an even stronger behavioural test.

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -748,44 +751,40 @@
...
+        ->assertDontSee("mountAction('viewPipelineEntry', { entry: '{$entry->getKey()}' })", false);
 });
 
-it('shows the view and edit actions with update permission on the pipeline\'s project', function () {
+it('mounts the view pipeline entry action from the card title without pipeline update permission', function () {
```
payal-canyon, line 768: Could we update the test names to reflect what they actually assert?

Both tests are currently named `it('mounts the … pipeline entry action from the card title …')`, but they don't actually mount an action - they only assert the rendered markup. This could be misleading when someone later uses the test name to understand or debug a failure.

could we rename them to something like:
```php
it('shows the `editPipelineEntry` action on the card title with pipeline update permission')
it('shows the `viewPipelineEntry` action on the card title without pipeline update permission')
```
Alternatively, if we switch to `mountAction(...)->assertActionMounted(...)` as suggested above, the existing `mounts the...` naming would accurately describe the behaviour being tested.

app-modules/project/resources/views/components/entry-card.blade.php
```diff
@@ -87,46 +91,24 @@ class="z-10 flex w-full transform cursor-move flex-col rounded-lg bg-white p-5 s
...
-            </x-filament::dropdown.list>
-        </x-filament::dropdown>
+        </button>
+
+        @can('update', $pipeline)
+            <x-filament::icon-button
+                class="shrink-0"
+                icon="heroicon-m-trash"
+                color="danger"
+                label="Remove"
```
payal-canyon, line 106: Could we add `tooltip="Remove"` to the trash button?

The current label is sufficient for accessibility - Filament uses it for the aria-label and native title - so there isn't an accessibility issue here.

However, the sibling `icon-only` button in `project-work-pipeline-heading.blade.php` uses both `label` and `tooltip`, which gives the user Filament's styled tooltip rather than the browser's default tooltip.

payal-canyon (changes requested):

payal-canyon (approved):

PR #1395, opened 2026-08-24
[AIDAPP-1439]: Introduce the ability to import and export organizations

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1439

### Technical Description

Adds case-insensitive uniqueness to Organization (and it's meta-data models) name's. Then adds an Organization exporter and importer.

No

### Cleanup Tasks

`.cleanup-tasks/2026_08_24_organization_name_uniqueness_feature.md`

---

### DIFF
--- a/app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
+++ b/app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
@@ -36,12 +36,17 @@
 
 namespace AidingApp\Contact\Filament\Resources\OrganizationResource\Pages;
 
+use AidingApp\Contact\Filament\Exports\OrganizationExporter;
 use AidingApp\Contact\Filament\Resources\OrganizationResource;
+use AidingApp\Contact\Imports\OrganizationImporter;
+use AidingApp\Contact\Models\Organization;
 use App\Filament\Tables\Columns\IdColumn;
 use Filament\Actions\BulkActionGroup;
 use Filament\Actions\CreateAction;
 use Filament\Actions\DeleteBulkAction;
 use Filament\Actions\EditAction;
+use Filament\Actions\ExportAction;
+use Filament\Actions\ImportAction;
 use Filament\Actions\ViewAction;
 use Filament\Resources\Pages\ListRecords;
 use Filament\Tables\Columns\TextColumn;
@@ -89,6 +94,13 @@ class ListOrganizations extends ListRecords
     protected function getHeaderActions(): array
     {
         return [
+            ImportAction::make()
+                ->importer(OrganizationImporter::class)
+                ->authorize('import', Organization::class),
+            ExportAction::make()
+                ->label('Export')
+                ->exporter(OrganizationExporter::class)
+                ->authorize('import', Organization::class),
             CreateAction::make(),
         ];
     }
--- a/app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
+++ b/app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
@@ -40,6 +40,7 @@ use AidingApp\Contact\Filament\Exports\OrganizationExporter;
 use AidingApp\Contact\Filament\Resources\OrganizationResource;
 use AidingApp\Contact\Imports\OrganizationImporter;
 use AidingApp\Contact\Models\Organization;
+use App\Features\OrganizationNameUniquenessFeature;
 use App\Filament\Tables\Columns\IdColumn;
 use Filament\Actions\BulkActionGroup;
 use Filament\Actions\CreateAction;
@@ -96,11 +97,13 @@ class ListOrganizations extends ListRecords
         return [
             ImportAction::make()
                 ->importer(OrganizationImporter::class)
-                ->authorize('import', Organization::class),
+                ->authorize('import', Organization::class)
+                ->visible(fn (): bool => OrganizationNameUniquenessFeature::active()),
             ExportAction::make()
                 ->label('Export')
                 ->exporter(OrganizationExporter::class)
-                ->authorize('import', Organization::class),
+                ->authorize('import', Organization::class)
+                ->visible(fn (): bool => OrganizationNameUniquenessFeature::active()),
             CreateAction::make(),
         ];
     }
--- a/app-modules/contact/src/Imports/OrganizationImporter.php
+++ b/app-modules/contact/src/Imports/OrganizationImporter.php
@@ -39,15 +39,16 @@ namespace AidingApp\Contact\Imports;
 use AidingApp\Contact\Models\Organization;
 use AidingApp\Contact\Models\OrganizationIndustry;
 use AidingApp\Contact\Models\OrganizationType;
-use AidingApp\Contact\Rules\UniqueOrganizationDomain;
 use Closure;
 use Filament\Actions\Imports\Exceptions\RowImportFailedException;
 use Filament\Actions\Imports\ImportColumn;
 use Filament\Actions\Imports\Importer;
 use Filament\Actions\Imports\Models\Import;
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 
 class OrganizationImporter extends Importer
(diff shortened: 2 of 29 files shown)


app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
```diff
@@ -89,6 +94,13 @@ public function table(Table $table): Table
     protected function getHeaderActions(): array
     {
         return [
+            ImportAction::make()
+                ->importer(OrganizationImporter::class)
+                ->authorize('import', Organization::class),
```
payal-canyon: Could we chain `->visible()` here to ensure these buttons are only shown when the feature flag is active?

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+                    }
+
+                    $record->domains = $domains;
+                })
+                ->example('collegecc.edu|college.edu'),
+        ];
+    }
+
+    public function resolveRecord(): ?Model
+    {
```
payal-canyon, line 212: Could we add the advisory lock here? Since chunked imports can run concurrently, two workers may both fail to find the same organization and then attempt to insert it. With the new unique constraint, that race results in a constraint violation and aborts the transaction.

The Olympus importers already handle this pattern with `pg_advisory_xact_lock()`. Could we apply the same approach here, before resolving the record?

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+            return;
+        }
+
+        /** @var array<int, array{domain: string}> $domains */
+        $domains = $this->record->domains ?? [];
+
+        $ignoreId = $this->record->exists ? $this->record->getKey() : null;
+
+        foreach ($domains as $domain) {
+            (new UniqueOrganizationDomain($ignoreId))->validate(
```
payal-canyon: The domain claims are written to Redis from inside the chunk's `DB::transaction()`, but Redis isn't part of that transaction, so the claims survive a rollback. If a chunk fails and retries, its rows can then fail with "already imported on an earlier row in this file" even though the import never actually happened. Since the claim TTL (`now()->addDay()`) matches the job's retry window (`getJobRetryUntil()`), the stale claims won't expire before the retries give up.

I don't think we should remove the cache, though - it covers a case a query-based check can't. Rows within the same chunk share a transaction, but Filament processes 100-row chunks independently, potentially on different workers. A domain repeated more than ~100 rows apart could therefore be invisible to both transactions, allowing both organizations to save the same domain.

Could we instead store the claiming organization's `name` as the cache value and only reject when the stored value belongs to a different organization? A retry would then re-claim its own domains and pass, while a genuinely different organization would still be rejected.

I'd key the claim by the `organization name` rather than the `record ID`, since a new organization doesn't have a `UUID` until it's saved.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+                ->rules(['nullable', 'email', 'max:255'])
+                ->example('admissions@collegecc.edu'),
+            ImportColumn::make('phone_number')
+                ->rules(['nullable', 'string'])
+                ->example('+1 (555) 555-5555'),
+            ImportColumn::make('website')
+                ->rules(['nullable', 'url', 'max:255'])
+                ->example('https://www.collegecc.edu'),
+            ImportColumn::make('industry')
+                ->relationship(
```
payal-canyon: Could we handle `blank` values explicitly for `industry` and `type`?

On an update, an `empty` mapped cell is still passed to `resolveUsing()`. That resolves to the default record and overwrites the organization's existing value. This means an export -> edit an unrelated field -> re-import can unintentionally assign the `default industry` or `type` to `organizations` that previously had no value.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+                    resolveUsing: function (mixed $state): ?OrganizationIndustry {
```
payal-canyon: Could we replace these `resolveUsing()` closures with `->relationship(resolveUsing: 'name')`?

Filament automatically validates relationship columns and rejects unknown values. However, because these closures fall back to the `is_default` record, an unknown value never resolves to `null`, so Filament's built-in exists validation never fails. As a result, a typo such as `Techology` could silently be imported as the default value.

Using the built-in relationship resolver would restore that validation and simplify the implementation:
`->relationship(resolveUsing: 'name')`

This should also work nicely with the `blank-value` handling discussed above: `blank` values remain allowed during validation and can be handled explicitly when filling the existing record.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+            ImportColumn::make('is_contact_generation_enabled')
+                ->label('Automatically generate contact record on login')
+                ->boolean()
+                ->rules(['boolean'])
+                ->example('false'),
+            ImportColumn::make('domains')
+                ->label('Domains')
+                ->rules(['nullable', 'string', function (string $attribute, mixed $value, Closure $fail): void {
+                    $domains = collect(preg_split('/[|,]/', (string) $value) ?: [])
+                        ->map(fn (string $domain): string => trim($domain))
```
payal-canyon, line 180: Could we simplify the `domains` handling using Filament's `->array()` and `->nestedRecursiveRules()` pattern, similar to `ClientImporter` in Olympus?

Right now the parsing is duplicated - both the validation closure and `parseDomains()` split on `|` / `,` , trim, and filter. The `->unique()` in `parseDomains()` is also redundant, since duplicates are already rejected during validation.

Using Filament's built-in array handling would give us a single parsing path and let us normalize domains with `Str::lower()` before both validation and storage. That would also fix the case-sensitivity gap where `Shared.edu` can pass when another organization already has `shared.edu`.

It would also let `UniqueOrganizationDomain` run per domain during validation, so errors are reported per domain rather than for the whole cell.

Two things to confirm:
- Should comma-separated input remain supported? `->array()` takes a single separator, but `castStateUsing()` also receives the raw unsplit value, so we can split on `/[|,]/` there if needed. Otherwise we can keep it pipe-only like Olympus.
- `nestedRecursiveRules()` changes the default error format (`The domains.1 field format is invalid`), so we'd want to override the domain-specific messages to preserve the current wording.

On the tests, `it('rejects a domain already used by another organization')` would now expect a `ValidationException` rather than `RowImportFailedException`.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+            ImportColumn::make('description')
+                ->example('A public community college serving the local community.'),
+            ImportColumn::make('number_of_employees')
+                ->integer()
+                ->rules(['nullable', 'integer', 'min:0'])
+                ->example('250'),
+            ImportColumn::make('address')
+                ->example('123 College Ave.'),
+            ImportColumn::make('city')
+                ->example('Riverside'),
```
payal-canyon, line 138: Could we add validation for `address`, `city`, `state`, `postalcode`, and `country`?

All five database columns are `varchar(255)`, and the forms and reference importers already enforce the same limit. Without validation, an over-long value reaches PostgreSQL and causes a database error, aborting the entire chunk transaction instead of failing the individual row cleanly.
Could we add:
```php
->rules(['nullable', 'string', 'max:255'])
```
to all five fields, and:
```php
->rules(['nullable', 'string'])
```
to description?

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+
+class OrganizationImporter extends Importer
+{
+    protected static ?string $model = Organization::class;
+
+    public static function getColumns(): array
+    {
+        return [
+            ImportColumn::make('name')
+                ->rules(['required', 'max:255'])
```
payal-canyon, line 64: Could we add both `->label()` and `->exampleHeader()` to all 19 columns?

Currently, only 5 columns define a `label`, and without `->exampleHeader()`, Filament falls back to the column name. This means the generated example CSV uses headers like `phone_number` and `postalcode`, while the exported CSV uses user-friendly headers like `Phone Number` and `Postal Code`.

app-modules/contact/tests/Tenant/Imports/OrganizationImporterTest.php
```diff
@@ -0,0 +1,279 @@
...
+    ]);
+
+    $organization = Organization::query()->where('name', 'Umbrella')->firstOrFail();
+
+    expect($organization->industry->is($defaultIndustry))->toBeTrue()
+        ->and($organization->type->is($defaultType))->toBeTrue();
+});
+
+it('fails the row when the name is missing', function () {
+    runOrganizationImport([
```
payal-canyon, line 177: Could we add tests for these behaviours?

- Blank `domains` should leave the organization's existing domains unchanged.
- Unknown `industry` / `type` should fail validation rather than silently falling back to the `default`.
- Blank `industry` / `type` on `update` should preserve the existing value rather than assigning the `default`.

app-modules/contact/tests/Tenant/Filament/Exports/OrganizationExporterTest.php
```diff
@@ -0,0 +1,119 @@
...
+        ->and($row[3])->toBe('https://www.acme.com')
+        ->and($row[4])->toBe('Technology')
+        ->and($row[5])->toBe('Partner')
+        ->and($row[9])->toBe('Springfield')
+        ->and($row[16])->toBe('acme.com|acme.org')
+        ->and($row[17])->toBe('true');
+});
+
+it('exports a blank domains value when the organization has no domains', function () {
+    $organization = Organization::factory()->create(['domains' => []]);
```
payal-canyon, line 110: Could we also add a test covering an organization with no `industry` or `type`?

This is a valid and common state, so it would be good to verify that the exporter produces a `blank` value rather than failing or attempting to access a missing relationship.

app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
```diff
@@ -89,6 +94,13 @@ public function table(Table $table): Table
...
+            ExportAction::make()
+                ->label('Export')
+                ->exporter(OrganizationExporter::class)
+                ->authorize('import', Organization::class),
```
payal-canyon: Could we add `->formats([ExportFormat::Csv])`?
Filament defaults to both `CSV` and `XLSX`, but importer only accepts `CSV`. That means users can export an `XLSX` file that they cannot re-import.

payal-canyon (dismissed):

app-modules/contact/src/Filament/Resources/OrganizationResource/Pages/ListOrganizations.php
```diff
@@ -89,6 +94,13 @@ public function table(Table $table): Table
...
             CreateAction::make(),
```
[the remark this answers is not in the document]
Orrison: This is expected in our systems.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+                        if (blank($state)) {
+                            return OrganizationIndustry::query()->where('is_default', true)->first();
+                        }
+
+                        return OrganizationIndustry::query()
+                            ->where('name', $state)
+                            ->first()
+                            ?? OrganizationIndustry::query()->where('is_default', true)->first();
+                    },
+                )
```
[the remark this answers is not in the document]
Orrison: This is an acceptable gap, we will leave it how it is for now.

app-modules/contact/src/Imports/OrganizationImporter.php
```diff
@@ -0,0 +1,251 @@
...
+        assert($this->record instanceof Organization);
+
+        // Domain uniqueness only needs enforcing when this row actually sets domains.
+        if (! $this->record->isDirty('domains')) {
+            return;
+        }
+
+        /** @var array<int, array{domain: string}> $domains */
+        $domains = $this->record->domains ?? [];
+
```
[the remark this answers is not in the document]
Orrison: Var narrowing is acceptable for arrays to define their structure like this.

app-modules/contact/src/Imports/OrganizationImporter.php
Orrison: Good call on the case-sensitivity gap — I've addressed that part directly: domains are now normalized to lowercase on import (before both validation and storage), and the cross-org check is a case-insensitive `jsonb_array_elements` query, so `Shared.edu` is now correctly rejected when another org already has `shared.edu`. The redundant `->unique()` in the parse step is gone too.

I opted **not** to do the full `->array()` + `nestedRecursiveRules()` restructure in this PR, though, for one main reason: it conflicts with the cache-based cross-row "first wins" design. Uniqueness here has two layers:

1. **Cross-org** (against already-persisted orgs) — this *can* live in validation, and effectively does now.
2. **Within-import, across rows** (first row wins, reject the rest) — this *cannot* be done purely in validation. Filament processes the CSV in independent 100-row chunks, potentially on different workers, so a validation rule only ever sees its own row. The atomic Redis claim (in `beforeSave`) is what actually enforces "first wins" across chunks, and that has to reject at save time, not validation time.

Because of that, moving everything to per-domain validation rules would give per-domain error messages for the cross-org case but would still leave the within-import enforcement (and its `RowImportFailedException`) exactly where it is — so we'd end up with a split model and a bunch of test churn (RowImportFailedException → ValidationException) for no real functional gain.

Comma-separated input is also something we want to keep (Olympus is pipe-only), which `->array()` doesn't handle without a custom `castStateUsing` split anyway.

Net: I kept the single-cell parsing path but fixed the substantive bug you flagged (case-insensitivity). If you feel strongly about the per-domain validation error UX, I'm happy to pick up the `->array()` restructure for the cross-org layer as a follow-up — just didn't want to fragment the uniqueness logic in this PR.

mxm1070 (approved):

PR #1402, opened 2026-08-25
[AIDAPP-1441]: System forward messages are not being properly processed by our system to create tickets

Orrison:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1441

### Technical Description

Fixes an issue where system forwarded emails to Aiding App are not correctly routed to the right Tenant and paths. Also adds a command to be able to retry the parsing of a failed inbound email.

No

### Cleanup Tasks

N/A

---

### DIFF
--- a/app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
@@ -41,13 +41,12 @@ use Illuminate\Console\Command;
 use Illuminate\Support\Facades\Config;
 use Illuminate\Support\Facades\Storage;
 use Illuminate\Support\Str;
-use Throwable;
 
 class RetryFailedSesS3InboundEmail extends Command
 {
     protected $signature = 'engagement:retry-failed-inbound-email {path : Path to the failed inbound email on the s3-inbound-email disk (e.g. failed/<id>)}';
 
-    protected $description = 'Move a failed inbound email back out of the failed folder and reprocess it';
+    protected $description = 'Move a failed inbound email back out of the failed folder and queue it for reprocessing';
 
     public function handle(): int
     {
@@ -61,23 +60,19 @@ class RetryFailedSesS3InboundEmail extends Command
             return static::FAILURE;
         }
 
-        // Reprocess from the original (non-failed) location so a re-failure moves it back to
-        // `failed/<id>` rather than nesting it under `failed/failed/<id>`.
+        // Move the file back to the disk root (its original pending location) so that a re-failure
+        // lands at `failed/<id>` rather than nesting under `failed/failed/<id>`, and so the queued
+        // job shares the same unique id the scheduled gatherer would use — letting the job's
+        // ShouldBeUnique lock dedupe the two and prevent double-processing.
         $originalPath = Str::startsWith($path, 'failed/') ? Str::after($path, 'failed/') : $path;
 
         if ($originalPath !== $path) {
             $disk->move($path, $originalPath);
         }
 
-        try {
-            ProcessSesS3InboundEmail::dispatchSync($originalPath);
-        } catch (Throwable $exception) {
-            $this->error("Reprocessing failed: {$exception->getMessage()}");
+        ProcessSesS3InboundEmail::dispatch($originalPath);
 
-            return static::FAILURE;
-        }
-
-        $this->info("Reprocessed inbound email: {$originalPath}");
+        $this->info("Queued inbound email for reprocessing: {$originalPath}");
 
         return static::SUCCESS;
     }
--- a/app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
+++ b/app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
@@ -1,0 +1,105 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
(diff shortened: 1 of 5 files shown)


app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
```diff
@@ -0,0 +1,105 @@
...
+        // Reprocess from the original (non-failed) location so a re-failure moves it back to
+        // `failed/<id>` rather than nesting it under `failed/failed/<id>`.
+        $originalPath = Str::startsWith($path, 'failed/') ? Str::after($path, 'failed/') : $path;
+
+        if ($originalPath !== $path) {
+            $disk->move($path, $originalPath);
+        }
+
+        try {
+            ProcessSesS3InboundEmail::dispatchSync($originalPath);
```
mxm1070: `ProcessSesS3InboundEmail::handle()` catches its four "instant-fail" exceptions (spam/virus detected, unable to detect tenant, unable to retrieve content, unknown service request type) and calls `$this->fail($exception)` without rethrowing, so `handle()` returns normally. Because `dispatchSync()` bypasses the queue worker's uncaught-exception handling, no exception propagates back here for this whole class of failures — so this command prints "Reprocessed inbound email: ..." and returns `SUCCESS`, even though the job immediately failed again and the file was moved right back to `failed/` (or `spam-or-virus-detected/`).

Since this is the retry tool's primary use case, a false-success report could lead an operator to believe an issue was fixed when it wasn't. Confirmed via an isolated test that `dispatchSync()` doesn't throw here even though `failed()` still fires.

Suggested fix: after `dispatchSync()`, check whether the file ended up back in a `failed/`-or-similar path before reporting success.

app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
```diff
@@ -0,0 +1,105 @@
...
+
+            return static::FAILURE;
+        }
+
+        // Reprocess from the original (non-failed) location so a re-failure moves it back to
+        // `failed/<id>` rather than nesting it under `failed/failed/<id>`.
+        $originalPath = Str::startsWith($path, 'failed/') ? Str::after($path, 'failed/') : $path;
+
+        if ($originalPath !== $path) {
+            $disk->move($path, $originalPath);
```
mxm1070, line 70: Moving the file into the disk root before the synchronous retry makes it visible to `GatherAndDispatchSesS3InboundEmails`, which scans that root every minute (`app/Console/Kernel.php`, `->everyMinute()->onOneServer()`) and dispatches a normal async `ProcessSesS3InboundEmail` job for anything it finds. `ProcessSesS3InboundEmail` implements `ShouldBeUnique`, but the unique-job lock is only acquired via `PendingDispatch` (the normal `dispatch()` path) — `dispatchSync()` bypasses that entirely. So if the gatherer's scheduled scan lands in the window between the move and this command's `dispatchSync()`/cleanup, it can dispatch a genuine duplicate job for the same file, risking double-processing (e.g. a duplicate service request/engagement). Reproduced empirically: `dispatchSync()` followed by a real `dispatch()` for the same path queues 2 jobs instead of being blocked.

Suggested fix: keep the file outside the gatherer's scanned root while retrying (e.g. dispatch directly against the `failed/...` path, or use a distinct in-progress prefix), or coordinate with the gatherer via a shared lock/claim.

mxm1070 (changes requested):
Found two correctness issues in the new `RetryFailedSesS3InboundEmail` command that I think should be addressed before merging (both independently reproduced with tests, not just theorized):

1. **False success reporting** — the command can print "Reprocessed inbound email" and exit successfully even when the retried job instantly fails again.
2. **Race condition with the gatherer job** — moving the file to the disk root before the synchronous retry creates a window where the every-minute `GatherAndDispatchSesS3InboundEmails` job can dispatch a genuine duplicate `ProcessSesS3InboundEmail` job for the same file, since `ShouldBeUnique`'s lock isn't acquired on the `dispatchSync()` path.

Details inline below.

app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
```diff
@@ -0,0 +1,100 @@
...
+        // lands at `failed/<id>` rather than nesting under `failed/failed/<id>`, and so the queued
+        // job shares the same unique id the scheduled gatherer would use — letting the job's
+        // ShouldBeUnique lock dedupe the two and prevent double-processing.
+        $originalPath = Str::startsWith($path, 'failed/') ? Str::after($path, 'failed/') : $path;
+
+        if ($originalPath !== $path) {
+            $disk->move($path, $originalPath);
+        }
+
+        ProcessSesS3InboundEmail::dispatch($originalPath);
```
mxm1070, line 73: Switching to `dispatch()` correctly fixes the race condition, but it introduces a new silent-success case: `ProcessSesS3InboundEmail` implements `ShouldBeUnique` with a 30-minute window. If the unique lock is already held when this runs (e.g. the file is already being processed, or the retry command is run twice quickly for the same path), `PendingDispatch::shouldDispatch()` returns `false` and the job is silently never queued, no exception, no signal. This command still prints "Queued inbound email for reprocessing" and returns `SUCCESS` in that case, even though nothing was dispatched.

Worth checking the lock result (or catching `UniqueJobSkipped`) and reporting a distinct outcome to the operator instead of a blanket success message.

mxm1070 (commented):
The fix for the race condition (switching to `dispatch()`) introduces a new, narrower issue below: it can silently skip the job entirely without telling the operator.

app-modules/engagement/src/Console/Commands/RetryFailedSesS3InboundEmail.php
Orrison: Not an issue, the only other system that can pick it up is the scheduler. And if it gets picked up there then it is fine, it still gets processed either way.

mxm1070 (approved):

PR #1408, opened 2026-08-26
[AIDAPP-1417]: Introduce support for keyword match monitors in service monitoring

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1417

### Technical Description

Add `MonitorType` enum, column to `ServiceMonitoringTarget`. Add `should_contain` and `should_not_contain` columns (nullable arrays)  to `ServiceMonitoringTarget`. Add `keyword_match_failures` (nullable array) to `HistoricalServiceMonitoring`. Add ability to populate said fields through UI. Modify `ServiceMonitoringCheckJob` to conditionally check for keywords instead of availability. Update email notification to display details of keyword match failures when appropriate.

No

### Cleanup Tasks

.cleanup-tasks/2026_08_24_monitor_type_feature.md

---


app-modules/service-management/database/migrations/2026_08_24_223522_add_keyword_match_columns_to_service_monitoring_targets_table.php
```diff
@@ -0,0 +1,60 @@
...
+use AidingApp\ServiceManagement\Enums\MonitorType;
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('service_monitoring_targets', function (Blueprint $table) {
+            $table->string('monitor_type')->initial(MonitorType::Availability->value);
```
Orrison: I think what copilot was trying to signal here is that we typically do not put references to classes that could change in a permanent migration. It would be better, and we do so in other places, to just put the actual string representation of the case value here.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/CreateServiceMonitoring.php
```diff
@@ -84,6 +90,30 @@ public function form(Schema $schema): Schema
...
+                            ->options(MonitorType::class)
+                            ->enum(MonitorType::class)
+                            ->default(MonitorType::Availability)
+                            ->live()
+                            ->inline()
+                            ->visible(MonitorTypeFeature::active())
+                            ->columnSpanFull(),
+                        TextEntry::make('helperText')
+                            ->hiddenLabel()
+                            ->state('Spaces may be used within a string. Use quotes when a string contains a comma.')
```
Orrison: Looking at this logic further I think we should also guide them that if they want a string that actually has quotes in it themselves, they ALSO need to surround that with quotes. Like ""something"" would save as and check for `"something"`

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/EditServiceMonitoring.php
```diff
@@ -89,6 +95,38 @@ public function form(Schema $schema): Schema
...
+                            ->options(MonitorType::class)
+                            ->enum(MonitorType::class)
+                            ->default(MonitorType::Availability)
+                            ->live()
+                            ->inline()
+                            ->visible(MonitorTypeFeature::active())
+                            ->columnSpanFull(),
+                        TextEntry::make('helperText')
+                            ->hiddenLabel()
+                            ->state('Spaces may be used within a string. Use quotes when a string contains a comma.')
```
Orrison: Same note as the create page.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/EditServiceMonitoring.php
```diff
@@ -89,6 +95,38 @@ public function form(Schema $schema): Schema
...
+                            ->visible(fn (Get $get): bool => $get('monitor_type') === MonitorType::KeywordMatch && MonitorTypeFeature::active())
+                            ->columnSpanFull(),
+                        TextInput::make('should_contain')
+                            ->label('Should Contain')
+                            ->formatStateUsing(fn (?array $state): ?string => filled($state) ? implode(', ', $state) : null)
```
Orrison: Hmm, for this (and below) will it properly rehydrate back out if one of the entries has commas in it?

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/ViewServiceMonitoring.php
```diff
@@ -74,6 +76,18 @@ public function infolist(Schema $schema): Schema
                                 TextEntry::make('frequency')
                                     ->label('Frequency')
                                     ->columnSpan(1),
+                                TextEntry::make('monitor_type')
+                                    ->label('Monitor Type')
+                                    ->visible(MonitorTypeFeature::active())
+                                    ->columnSpanFull(),
+                                TextEntry::make('should_contain')
+                                    ->label('Should Contain')
+                                    ->separator(', ')
```
Orrison: Is this going to properly render for entries that contain commas?

app-modules/service-management/src/Rules/ValidServiceMonitoringKeywordValues.php
```diff
@@ -0,0 +1,100 @@
...
+*/
+
+namespace AidingApp\ServiceManagement\Rules;
+
+use Closure;
+use Illuminate\Contracts\Validation\ValidationRule;
+
+class ValidServiceMonitoringKeywordValues implements ValidationRule
+{
+    public function __construct(protected mixed $shouldContain, protected mixed $shouldNotContain) {}
```
Orrison: Rather than having to pass in the values of both the fields with `$get` on construct here, I wonder if it makes more sense to make this a `DataAwareRule` (under "Accessing Additional Data"
 in https://laravel.com/framework/docs/validation)

app-modules/service-management/src/Rules/ValidServiceMonitoringKeywordValues.php
```diff
@@ -0,0 +1,100 @@
...
+            $fail('Double quotes must be paired.');
+
+            return;
+        }
+
+        if (is_string($value) && blank(array_filter(array_map('trim', str_getcsv($value)), filled(...)))) {
+            $fail('Keyword values cannot be empty.');
+
+            return;
+        }
```
Orrison: Does this make both fields required? If so, I think what we really need to do is ensure that at least one of them has valid data in it.

It should fail only if BOTH this and the other input are empty

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
Orrison, line 1: Just to be really safe I think a `KeywordMatch` monitor should fail if both should contain and not contain are empty.

I know validation on the frontend prevents that from happening, but this job doesn't know that and in the future something could allow for that.

app-modules/service-management/src/Rules/ValidServiceMonitoringKeywordValues.php
```diff
@@ -0,0 +1,100 @@
...
+
+        if (array_intersect($this->parse($this->shouldContain), $this->parse($this->shouldNotContain)) !== []) {
+            $fail('A keyword value cannot appear in both lists.');
+
+            return;
+        }
```
Orrison: I think this needs to have some case-insensitive handling. `Error` and `error` would pass through here without being caught as being the same

app-modules/service-management/src/Jobs/ServiceMonitoringCheckJob.php
```diff
@@ -131,4 +139,61 @@ public function handleResponses(int $status, float $responseTime, bool $success)
...
+        }
+    }
+
+    protected function handleKeywordMatch(): void
+    {
+        try {
+            $response = Http::maxRedirects(15)
+                ->get($this->serviceMonitoringTarget->domain);
+
+            $responseBody = $this->readableResponseBody($response->body());
```
Orrison: We may need to check the `Content-Type` of the response and consider it a failure if it is an unreadable type like binary or non-text content

Orrison: Suggesting we rework the text extraction here. The current `readableResponseBody()` approach has a few issues that will produce incorrect match results on real-world pages:

1. **`strip_tags()` joins adjacent text with no separator.** `<p>Sign</p><p>in</p>` becomes `Signin`. Source whitespace between tags usually masks this, but minified HTML (most production pages) has none, so phrases split across block elements will fail to match.
2. **The whitespace collapse misses NBSP.** `html_entity_decode()` turns `&nbsp;` into U+00A0, and `/\s+/` without the `u` modifier doesn't match it. `Sign&nbsp;in` will never match a configured `sign in`. This is a live false-positive generator today.
3. **The script/style regex is fragile.** It breaks on unclosed tags, `</script` inside JS strings, and `>` inside attributes. `noscript` and `template` content also survives into the searchable text even though it never renders.

Since we're on PHP 8.4, we should use `Dom\HTMLDocument`. It's the spec-compliant HTML5 parser, recovers from malformed markup the way a browser does, and decodes entities for us, so `html_entity_decode()` goes away entirely.

It would be best as well to pull extraction out of the job into a dedicated class that walks the DOM and adds separators only at *visible* boundaries — block elements and `<br>` get a space, inline elements (`span`, `strong`, `em`, `a`) do not. That preserves the intended matching semantics: `<span>New</span><span>York</span>` stays `NewYork`, while `<p>Sign</p><p>in</p>` becomes `Sign in`.

This should be roughly what would work, but may need some refinement.

```php
final class HtmlTextExtractor
{
    private const NON_VISIBLE = ['script', 'style', 'noscript', 'template', 'head'];

    private const BLOCK = [
        'address', 'article', 'aside', 'blockquote', 'br', 'dd', 'details',
        'div', 'dl', 'dt', 'fieldset', 'figcaption', 'figure', 'footer',
        'form', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'header', 'hr', 'li',
        'main', 'nav', 'ol', 'option', 'p', 'pre', 'section', 'summary',
        'table', 'td', 'th', 'tr', 'ul',
    ];

    public function extract(string $html, ?string $encoding = null): string
    {
        $document = \Dom\HTMLDocument::createFromString(
            $html,
            LIBXML_NOERROR,
            $encoding,
        );

        $buffer = '';
        $this->walk($document->body ?? $document, $buffer);

        return trim((string) preg_replace('/\s+/u', ' ', $buffer));
    }

    private function walk(\Dom\Node $node, string &$buffer): void
    {
        foreach ($node->childNodes as $child) {
            if ($child instanceof \Dom\Text) {
                $buffer .= $child->textContent;

                continue;
            }

            if (! $child instanceof \Dom\Element) {
                continue;
            }

            $tag = strtolower($child->localName);

            if (in_array($tag, self::NON_VISIBLE, true)) {
                continue;
            }

            $isBlock = in_array($tag, self::BLOCK, true);

            if ($isBlock) {
                $buffer .= ' ';
            }

            $this->walk($child, $buffer);

            if ($isBlock) {
                $buffer .= ' ';
            }
        }
    }
}
```

A few implementation details to include:

1. `LIBXML_NOERROR` so malformed customer HTML doesn't spam warnings on every scheduled check.
2. Pass the charset from the response's `Content-Type` header as `$encoding` when present; otherwise the parser sniffs via BOM/meta per the HTML5 encoding algorithm.
3. Normalize the needles the same way as the haystack — apply the same `/\s+/u` collapse to the configured `should_contain` / `should_not_contain` strings before comparing, so a double space or pasted NBSP in a customer's configured string doesn't reintroduce the mismatch.
4. Cap the response body size before parsing so a huge response can't OOM a queue worker.

Tests should cover the boundary contract explicitly:

- `<span>Sign</span><span>in</span>` does **not** match `sign in`
- `<p>Sign</p><p>in</p>` matches `sign in`
- `Sign<br>in` matches `sign in`
- `<strong>sign</strong> <em>in</em>` matches `sign in`
- `Sign&nbsp;in` matches `sign in`
- Minified HTML (no source whitespace) still gets block boundaries
- `script`, `style`, and `noscript` content is excluded

One thing I would ask you to add a comment to the ticket about: Neither the current implementation OR what I am suggesting will work for a JS based SPA website. Since we don't render the JS.
I think that is acceptable for this increment and there are less and less non-server-side rendered SPAs these days. But I think we should document that and make sure Product knows this is a limitation of this current iteration and if they want that covered we will need to do some more advanced browser rendering.

Orrison (changes requested):
We changed the content of the email to contain the specific failures of `KeywordMatch`. But I think we should add that to the body of the in-app notification as well, as people who only have it doing an in-app notification wouldn't get that data in the notice.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/ViewServiceMonitoring.php
rebecca-canyon: It did, but stripped out any external double quotes, which looked confusing; so I've moved each item to be on its own line to be more clear.

Orrison (approved):

PR #1412, opened 2026-08-26
[AIDAPP-1413]: Improve the presentation of the list view in projects for milestones

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1413

### Technical Description

> Improve the presentation of the list view in projects for milestones

> No

### Cleanup Tasks

> N/A

---


app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -348,34 +361,106 @@ public function editMilestoneAction(): Action
...
+            ->whereNull('project_milestones.archived_at')
+            ->whereNotExists(function (QueryBuilder $query) use ($pipeline): void {
+                $query->select(new Expression('1'))
+                    ->from('pipeline_entries')
+                    ->join('pipeline_stages', 'pipeline_stages.id', '=', 'pipeline_entries.pipeline_stage_id')
+                    ->whereColumn('pipeline_entries.project_milestone_id', 'project_milestones.id')
+                    ->where('pipeline_stages.pipeline_id', $pipeline->getKey())
+                    ->whereNull('pipeline_entries.archived_at');
+            })
+            ->select($placeholderSelects);
```
payal-canyon, line 456: Could we include the soft-delete scope in the placeholder milestone query?

The query currently uses `getConnection()->table('project_milestones')`, which bypasses Eloquent scopes. Since `ProjectMilestone` uses `SoftDeletes`, it filters `archived_at` but not `deleted_at`.

This becomes an issue with the delete flow introduced here: the milestone's entries are removed and the milestone is soft-deleted, but `resetTable()` then re-renders the table. Because the eager-loaded milestone excludes trashed records, the soft-deleted milestone can appear as a phantom "No tasks yet" row under "No Associated Milestone".

The minimal fix would be:
`->whereNull('project_milestones.deleted_at')`

Alternatively, we could build this from `ProjectMilestone::query()` and call `->toBase()`, which would apply the model scopes consistently with `getActionMilestone()`.

The existing delete test only checks `trashed()` and doesn't re-render the table, so it doesn't catch this case.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -209,15 +213,17 @@ function () use ($pipeline) {
                                 );
                         })
                         ->getTitleFromRecordUsing(
-                            fn (PipelineEntry $record): string => $record->milestone->title ?? 'No Associated Milestone'
+                            fn (PipelineEntry $record): Htmlable => new HtmlString(
+                                view('project::filament.tables.groups.milestone-title', [
+                                    'milestone' => $record->milestone,
+                                ])->render()
```
payal-canyon: `Group::getTitle()` calls `resolveTitle()` for every record when `getTitleFromRecordUsing` is set. The new title view then runs `@can('update', $milestone)` per row. `ProjectMilestonePolicy::update()` ultimately performs two `exists()` queries for non-super-admins, so with 50 rows this adds roughly 100 extra queries per table render.

Since the policy resolves to `can('update', $milestone->project)`, and this widget already uses the same check at [line no: 245, 278, and 319] could we compute it once in table() and pass the boolean to the view instead?

This would avoid the per-row policy/query overhead.

app-modules/project/resources/views/filament/tables/groups/milestone.blade.php
```diff
@@ -31,31 +31,6 @@
...
-                    :wire:click="'mountAction(\'deleteMilestone\', { milestone:\'' . $milestone->getKey() . '\' })'"
-                >
-                    Delete
-                </x-filament::link>
-            @endcan
-        </span>
-    @endif
-</span>
+<div class="flex w-full justify-end">
+    <x-project::progress-circle :progress="$percentage" />
```
payal-canyon: Could we avoid rendering the progress ring for tasks without a milestone?

The "No Associated Milestone" group now shows a "Progress: 0%" ring because the description view renders it unconditionally, while `milestoneProgressPercentage()` returns 0 when `project_milestone_id` is blank.

Previously this returned ` `, allowing the existing CSS to hide the empty description. Since that no longer applies, could we guard the view and render the progress ring only when a milestone exists? Alternatively, `getDescriptionFromRecordUsing()` could return `null` for records without a milestone.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -209,15 +213,17 @@ function () use ($pipeline) {
...
+                            )
                         )
                         ->getDescriptionFromRecordUsing(
                             fn (PipelineEntry $record): View => view('project::filament.tables.groups.milestone', [
-                                'milestone' => $record->milestone,
-                                'progress' => $this->milestoneProgressDescription($record, $pipeline),
+                                'percentage' => $this->milestoneProgressPercentage($record, $pipeline),
```
payal-canyon: Could we restore `->collapsible()` on the group definition?

It looks like it was removed even though the ticket doesn't require it, which means users can no longer collapse milestone groups. The new title link likely shares the header click area, but the previous markup handled this with `x-on:click.stop`.

Could we restore `->collapsible()` and add `x-on:click.stop` to the `<x-filament::link>` in `milestone-title.blade.php`? This should preserve both the collapse and title-link behaviours.

app-modules/project/resources/views/components/progress-circle.blade.php
```diff
@@ -61,5 +61,5 @@ class="text-gray-200 dark:text-gray-700"
             class="text-primary-600 dark:text-primary-500"
         />
     </svg>
-    <span>Progress: {{ $progress }}%</span>
+    <span class="text-sm text-primary-600 dark:text-primary-500">Progress: {{ $progress }}%</span>
```
payal-canyon: Could we scope the new styling to the pipeline widget instead?

Adding `text-sm text-primary-600 dark:text-primary-500` to the shared `x-project::progress-circle` component affects two other call sites outside this ticket's scope: the project dashboard header and the List Projects table. In both cases, the progress text is now styled differently from its surrounding content.

Could we apply these classes through the existing `.project-work-pipeline-widget` styles, or make the component styling optional via a class/size prop? That would keep the existing call sites unchanged while applying the new styling only where needed.

app-modules/project/resources/views/filament/tables/groups/milestone.blade.php
```diff
@@ -31,31 +31,6 @@
...
-                    color="danger"
-                    :wire:click="'mountAction(\'deleteMilestone\', { milestone:\'' . $milestone->getKey() . '\' })'"
-                >
-                    Delete
-                </x-filament::link>
-            @endcan
-        </span>
-    @endif
-</span>
+<div class="flex w-full justify-end">
```
payal-canyon: Could we keep the progress markup as phrasing content?

Filament wraps the group description in a `<p class="fi-ta-group-description">` in both render paths, but the new view outputs a `<div>`. Since `<div>` isn't valid inside `<p>`, the browser reparents it, which also makes the second `@assets` rule ineffective and prevents Filament's description styling from applying.

The previous implementation already handled this correctly with `<span>` elements.

Could we switch the wrapper to `<span class="flex w-full justify-end">` and use a span-rooted variant of progress-circle? This would keep the HTML valid, allow the intended CSS to work, and avoid needing the global styling change to the shared component.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -150,6 +146,7 @@ public function table(Table $table): Table
                     ->view('project::filament.tables.columns.pipeline-entry.tickets'),
                 IconColumn::make('is_visible_to_guests')
                     ->label('Customer Visible')
+                    ->state(fn (PipelineEntry $record): ?bool => $this->isPlaceholderRecord($record) ? null : $record->is_visible_to_guests)
```
payal-canyon: Could we remove the `is_visible_to_guests` state override?

The placeholder query already selects `null as is_visible_to_guests`, and Laravel's boolean cast preserves `null`. `IconColumn::getIcon()` also returns `null` for a null state, so the cell is already blank for both placeholder and real rows without this closure.

Since the closure adds no behavioural value and introduces an unnecessary untested callback, could we drop it?

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -348,34 +361,106 @@ public function editMilestoneAction(): Action
...
     {
-        return Action::make('deleteMilestone')
-            ->label('Delete')
-            ->icon('heroicon-m-trash')
-            ->color('danger')
-            ->requiresConfirmation()
-            ->authorize(fn (Action $action): bool => auth()->user()->can('delete', $this->getActionMilestone($action)))
-            ->action(function (Action $action): void {
-                $milestone = $this->getActionMilestone($action);
+        $columns = ['id', 'name', 'pipeline_stage_id', 'project_milestone_id', 'is_visible_to_guests', 'start_date', 'due', 'created_by'];
```
payal-canyon, line 426: Could we test the partial-model behaviour here?

Because the outer query uses `fromSub(...)`, `PipelineEntry` is hydrated with only the attributes listed in `$columns`. Any omitted attributes will silently resolve to `null`, which could cause subtle issues if a table column or closure later accesses them.

Could we add a short comment noting that `$columns` must include every attribute used by the table, and add a test verifying that a real entry's `created_by` and `is_visible_to_guests` still render correctly through the union path? The existing `displays a pipeline entry start date…` test already covers `start_date`, so this would follow the same pattern.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -655,13 +656,59 @@ function loginAsUserWithProjectViewPermissions(): User
...
+            TestAction::make('deleteMilestone'),
+        ])
         ->assertHasNoActionErrors();
 
     expect($milestone->fresh()->trashed())->toBeTrue()
         ->and($entry->fresh()->project_milestone_id)->toBeNull();
 });
 
+it('shows milestones that have no pipeline tasks as an empty group', function () {
+    asSuperAdmin();
```
payal-canyon, line 885: Could we add coverage for the new closure-driven behaviour?

There are a few paths added here that currently aren't tested:

- Archived/deleted milestones: Add a case ensuring a soft-deleted milestone isn't rendered as an empty group. This would catch the current placeholder-query regression.
- Classification filter: Add coverage confirming placeholder entries still appear even when the normal classification filter would exclude them.
- `@can('update')` title: Verify that view-only users see the milestone title as plain text and cannot access manageMilestone. The existing project-permission test provides a similar pattern.
- deleteMilestone visibility: Add assertActionVisible/assertActionHidden coverage for the permission-based visible() closure, rather than only covering it for a super admin.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -112,29 +116,21 @@ public function table(Table $table): Table
...
-                        'pipelineStage.pipeline.project',
-                        'milestone',
-                    ]);
-            })
+            ->query(fn (): Builder => $this->getPipelineEntriesQuery($pipeline))
             ->heading(fn (): View => $this->getTableHeadingView($pipeline))
             ->columns([
                 TextColumn::make('name')
                     ->label('Task Name')
+                    ->state(fn (PipelineEntry $record): string => $this->isPlaceholderRecord($record) ? 'No tasks yet' : $record->name)
```
payal-canyon, line 124: Could we disable the column click for placeholder records?

`->action()` is currently registered for every record, so "No tasks yet" still renders as a clickable `<button>` with a pointer cursor. Clicking it also triggers an unnecessary Livewire request before the action returns early.

Could we add:
`->disabledClick(fn (PipelineEntry $record): bool => $this->isPlaceholderRecord($record))`

This would render placeholder rows as plain text, avoid the unnecessary request, and allow the early-return guard inside `->action()` to be removed.

payal-canyon (changes requested):

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -209,15 +210,21 @@ function () use ($pipeline) {
...
+                            )
                         )
                         ->getDescriptionFromRecordUsing(
-                            fn (PipelineEntry $record): View => view('project::filament.tables.groups.milestone', [
-                                'milestone' => $record->milestone,
-                                'progress' => $this->milestoneProgressDescription($record, $pipeline),
-                            ])
-                        )
-                        ->collapsible();
+                            fn (PipelineEntry $record): ?View => filled($record->project_milestone_id)
```
payal-canyon: Could we gate this on `$record->milestone` instead of `filled($record->project_milestone_id)`? The FK can still be populated when the milestone is soft-deleted, causing the relation to return `null` and the view to error when it dereferences `$milestone`.

This is also consistent with the existing relation check at :215. Could we add a test covering a task whose milestone is soft-deleted directly, and verify the pipeline renders it as “No Associated Milestone”?

app-modules/project/resources/views/filament/resources/projects/widgets/project-work-pipeline-widget.blade.php
```diff
@@ -32,7 +32,47 @@
...
+        {{ $this->table }}
+    </div>
+
+    @assets
+        {{-- Allow the milestone group heading block to fill the row and keep the title and progress vertically aligned. --}}
+        <style>
+            .project-work-pipeline-widget .fi-ta-group-header > div {
+                display: flex;
+                flex: 1 1 auto;
+                align-items: center;
```
payal-canyon: Could we move the four rules into `theme.css`, keeping the `.project-work-pipeline-widget` scope, and remove the now-dead `.project-milestone-group-actions` block?

The branch removed that wrapper class, so those rules no longer match, and the new `> .fi-icon-btn { order: -1 }` also duplicates the existing button ordering rule. This would keep the styling in the appropriate place and remove the duplication.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -655,13 +819,184 @@ function loginAsUserWithProjectViewPermissions(): User
...
+        ->for($project)
+        ->has(PipelineStage::factory()->count(1), 'stages')
+        ->create();
+    $milestone = ProjectMilestone::factory()->for($project)->create(['title' => 'Empty Milestone']);
+
+    livewire(ProjectWorkPipelineWidget::class, [
+        'record' => $project,
+    ])
+        ->assertSee($milestone->title)
+        ->assertSee('No tasks yet');
```
payal-canyon: Since the name column now uses a custom `->state()`, could we update the relevant tests to assert the resolved column state directly instead of only checking that the text is visible?

For example, `assertTableColumnStateSet('name', 'No tasks yet', record: $milestone->getKey())` would verify that the custom state is resolved correctly.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -655,13 +819,184 @@ function loginAsUserWithProjectViewPermissions(): User
...
+
+    $emptyMilestone = ProjectMilestone::factory()->for($project)->create(['title' => 'Empty Milestone']);
+
+    livewire(ProjectWorkPipelineWidget::class, [
+        'record' => $project,
+    ])
+        ->filterTable('classification', PipelineStageClassification::Planning)
+        ->assertCanNotSeeTableRecords([$entry])
+        ->assertSee($emptyMilestone->title)
+        ->assertSee('No tasks yet');
```
payal-canyon: same as above.

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -655,13 +819,184 @@ function loginAsUserWithProjectViewPermissions(): User
...
+    $entry = PipelineEntry::factory()->create([
+        'pipeline_stage_id' => $pipeline->stages->sole()->getKey(),
+    ]);
+    $milestone = ProjectMilestone::factory()->for($project)->create(['title' => 'Empty Milestone']);
+
+    livewire(ProjectWorkPipelineWidget::class, [
+        'record' => $project,
+    ])
+        ->assertSeeHtml("wire:click.prevent.stop=\"callTableColumnAction(&#039;name&#039;, &#039;{$entry->getKey()}&#039;)\"")
+        ->assertDontSeeHtml("wire:click.prevent.stop=\"callTableColumnAction(&#039;name&#039;, &#039;{$milestone->getKey()}&#039;)\"");
```
payal-canyon: Could we avoid asserting Filament’s internal HTML here? The current assertions depend on the generated `wire:click` markup and internal `callTableColumnAction` method, which makes the test brittle.

Could we instead assert the column’s `extraAttributes` for the real entry vs. the placeholder milestone? This would verify the intended behaviour without coupling the test to Filament’s internal markup, and would also cover the currently untested `->extraAttributes()` logic.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -209,15 +210,21 @@ function () use ($pipeline) {
                                 );
                         })
                         ->getTitleFromRecordUsing(
-                            fn (PipelineEntry $record): string => $record->milestone->title ?? 'No Associated Milestone'
+                            fn (PipelineEntry $record): Htmlable => new HtmlString(
+                                trim(view('project::filament.tables.groups.milestone-title', [
```
payal-canyon: Since the title is now plain text, could we return `$record->milestone?->title ?? 'No Associated Milestone'` directly instead of rendering a separate Blade view?

This avoids a Blade render per row, keeps the value as a plain string for escaping and `@js()`, and removes a view that’s only used in this one place. The warning comment could also be moved above the closure, with the unused view removed.

app-modules/project/resources/views/components/progress-circle.blade.php
```diff
@@ -32,11 +32,20 @@
     </COPYRIGHT>
 --}}
 
+{{--
+    The `tag` prop lets callers embedding this inside a Filament table Group description
+    render it as a `<span>` instead of the default `<div>`, since Group descriptions only
+    accept phrasing content (see the "Group descriptions must be phrasing content only"
+    rule). Text styling (size/color) is intentionally left off the label below and must be
+    passed in via `class`, so it only affects the call sites that opt into it.
+--}}
```
payal-canyon: The comment refers to a “Group descriptions must be phrasing content only” rule, but I couldn’t find that rule elsewhere in the repo. Could we update the comment to reference the actual guideline or remove the reference if there isn’t one?

app-modules/project/tests/Tenant/Filament/Resources/ProjectResource/Pages/ViewProjectTest.php
```diff
@@ -638,6 +661,147 @@ function loginAsUserWithProjectViewPermissions(): User
...
+    $milestone = ProjectMilestone::factory()->for($project)->create(['title' => 'Alpha']);
+    PipelineEntry::factory()->create([
+        'pipeline_stage_id' => $pipeline->stages->first()->getKey(),
+        'project_milestone_id' => $milestone->getKey(),
+    ]);
+
+    // Filament reuses the Group title raw inside the collapsible toggle's `aria-label`
+    // attribute. If the title view ever renders interactive markup (e.g. a wire:click
+    // button) again, its own double-quoted attribute breaks out of this aria-label and
+    // corrupts the surrounding row markup instead of rendering just the milestone title.
```
payal-canyon: The test names already make the intent clear. Could we remove these `// comments` and rely on the test names instead?

app-modules/project/resources/views/filament/tables/groups/milestone.blade.php
```diff
@@ -31,31 +31,35 @@
...
+--}}
+<span class="flex w-full items-center justify-between gap-x-3">
+    @if ($canManageMilestone)
+        <x-filament::link
+            tag="button"
+            x-on:click.stop
+            wire:click="mountAction('manageMilestone', { milestone: '{{ $milestone->getKey() }}' })"
+        >
+            {{ $milestone->title }}
+        </x-filament::link>
```
payal-canyon, line 55: Since `titlePrefixedWithLabel(false)` makes the collapse button’s `aria-label` the milestone title, the manage link in the same group header has the same accessible name but a different action - opening the manage slide-over. This makes the two buttons indistinguishable to screen reader users.

Could we add a visually hidden `edit milestone` qualifier to the manage link? `fi-sr-only` is already a Filament utility class used for text alternatives.

payal-canyon (changes requested):

resources/css/filament/admin/theme.css
```diff
@@ -95,20 +95,26 @@ li[data-group-label=''] > ul > .fi-sidebar-item > .fi-sidebar-item-button > .fi-
...
+.project-work-pipeline-widget .fi-ta-group-header > div {
+    display: flex;
+    flex: 1 1 auto;
+    align-items: center;
+}
+
+.project-work-pipeline-widget .fi-ta-group-header > div > .fi-ta-group-description {
+    flex: 1 1 auto;
+}
+
```
payal-canyon, line 107: One thing I noticed is that the `collapse-toggle` ordering rule didn’t make it over when the styles were moved to `theme.css`. Without it, the collapse button appears on the right of the milestone title/progress instead of the left.

Could we re-add the `order: -1` rule? Also, could we quickly check the group header at a narrow viewport with a long milestone title to make sure the previous `shrink-0 self-start` and `flex-nowrap items-start` styles aren’t needed to prevent wrapping issues?

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
```diff
@@ -212,12 +211,14 @@ function () use ($pipeline) {
...
                         ->getDescriptionFromRecordUsing(
-                            fn (PipelineEntry $record): View => view('project::filament.tables.groups.milestone', [
-                                'milestone' => $record->milestone,
-                                'progress' => $this->milestoneProgressDescription($record, $pipeline),
-                            ])
-                        )
-                        ->collapsible();
+                            fn (PipelineEntry $record): ?View => $record->milestone
+                                ? view('project::filament.tables.groups.milestone', [
+                                    'milestone' => $record->milestone,
```
payal-canyon, line 216: The crash fix looks correct, but there’s currently no test covering it. The existing soft-delete test uses a milestone with no entries, so it doesn’t exercise the description callback when the milestone relation is `null`.

Could we add a test that creates an entry for a milestone, soft-deletes the milestone directly, renders the widget, and asserts `assertSee('No Associated Milestone')`? This would pin the fix and prevent the regression from going unnoticed.

payal-canyon (changes requested):
Looks good overall! Just a couple of things to address, and then I think it’ll be good to merge.

app-modules/project/src/Filament/Resources/Projects/Widgets/ProjectWorkPipelineWidget.php
jayushi-canyon: This test case is already taken care of in `ProjectWorkPipelineWidgetTest` file test `it('renders a task as having no associated milestone when its milestone has been soft-deleted)`.

payal-canyon (approved):

PR #1413, opened 2026-08-26
[AIDAPP-830]: Address accessibility conformance issue for 1.1.1 Non-text Content

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-830

### Technical Description

> Address accessibility conformance issue for 1.1.1 Non-text Content

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/form/resources/views/blocks/submissions/signature.blade.php
+++ b/app-modules/form/resources/views/blocks/submissions/signature.blade.php
@@ -34,7 +34,7 @@
 <x-form::blocks.field-wrapper class="py-3" :$label :$isRequired>
     <div class="flex h-16 max-w-xs items-center dark:bg-white">
         @if (filled($response ?? null))
-            <img class="h-16 max-w-xs" src="{{ $response }}" alt="Signature" />
+            <img class="h-16 max-w-xs" src="{{ $response }}" alt="{{ $label }}" />
         @else
             <div class="px-3 text-gray-500">No response</div>
         @endif
--- a/app-modules/form/resources/views/blocks/submissions/signature.blade.php
+++ b/app-modules/form/resources/views/blocks/submissions/signature.blade.php
@@ -34,7 +34,7 @@
 <x-form::blocks.field-wrapper class="py-3" :$label :$isRequired>
     <div class="flex h-16 max-w-xs items-center dark:bg-white">
         @if (filled($response ?? null))
-            <img class="h-16 max-w-xs" src="{{ $response }}" />
+            <img class="h-16 max-w-xs" src="{{ $response }}" alt="Signature" />
         @else
             <div class="px-3 text-gray-500">No response</div>
         @endif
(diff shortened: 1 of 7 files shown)


app-modules/in-app-communication/resources/js/components/UserSearchSelect.vue
```diff
@@ -145,9 +146,17 @@
                         :display-value="() => searchQuery"
                         @change="searchQuery = $event.target.value"
                     />
-                    <ComboboxButton class="absolute inset-y-0 right-0 flex items-center pr-3">
-                        <ArrowPathIcon v-if="isSearching" class="h-5 w-5 animate-spin text-gray-400" />
-                        <ChevronUpDownIcon v-else class="h-5 w-5 text-gray-400" />
+                    <ComboboxButton
+                        class="absolute inset-y-0 right-0 flex items-center pr-3"
+                        aria-label="Toggle user options"
```
payal-canyon, line 153: Could we add a visually hidden label for the combobox input?

The new `aria-label="Toggle user options"` is also used as the accessible name for the `ComboboxInput` because there is no `ComboboxLabel`. As a result, the search field is announced as `"Toggle user options"`, which describes the button rather than the input.

Could we follow the existing Headless UI pattern and add a visually hidden label?
```javascript
import { ..., ComboboxLabel, ... } from '@headlessui/vue';

<div class="relative w-full">
    <ComboboxLabel class="sr-only">Search users</ComboboxLabel>
    <ComboboxInput ... />
```
This gives the input the correct accessible name while keeping the button's existing label intact.

app-modules/in-app-communication/resources/js/components/ServiceRequestConversationQueue.vue
```diff
@@ -108,7 +108,7 @@
 
         <template v-else-if="activeItems.length === 0 && !inline">
             <div class="flex flex-col items-center justify-center p-8 text-center">
-                <ClockIcon class="w-10 h-10 text-gray-300 dark:text-gray-600 mb-2" />
+                <ClockIcon class="w-10 h-10 text-gray-300 dark:text-gray-600 mb-2" aria-hidden="true" />
```
payal-canyon: Could we remove the redundant `aria-hidden="true"` attributes from the `Heroicons`?

`@heroicons/vue` already renders `aria-hidden="true"` on the root `<svg>`, so the five attributes added here don't change the rendered output. They're harmless, but they also make the diff noisier and imply a convention that isn't applied consistently to the other Heroicons in these components.

I'd suggest removing those redundant attributes so the diff contains only changes with an actual effect.

The `aria-hidden="true"` on the hand-written SVG in `progress-circle.blade.php` is different and should remain, since the adjacent `Progress: {{ $progress }}%` text already provides the accessible meaning.

app-modules/form/resources/views/blocks/submissions/signature.blade.php
```diff
@@ -34,7 +34,7 @@
 <x-form::blocks.field-wrapper class="py-3" :$label :$isRequired>
     <div class="flex h-16 max-w-xs items-center dark:bg-white">
         @if (filled($response ?? null))
-            <img class="h-16 max-w-xs" src="{{ $response }}" />
+            <img class="h-16 max-w-xs" src="{{ $response }}" alt="Signature" />
```
payal-canyon: Could we use the field `label` for the image alt text instead of the literal `Signature`?
` alt="{{ $label }}" `

payal-canyon (changes requested):

payal-canyon (approved):

PR #1420, opened 2026-08-27
[AIDAPP-1369]: Introduce support for desktop notifications

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1369

### Technical Description

> Introduce support for desktop notifications

> NO

### Cleanup Tasks

> N/A

---


database/migrations/2026_08_25_130000_create_push_subscriptions_table.php
```diff
@@ -0,0 +1,69 @@
...
+use Illuminate\Database\Schema\Blueprint;
+use Illuminate\Support\Facades\DB;
+use Illuminate\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        DB::transaction(function () {
+            Schema::create(config('webpush.table_name', 'push_subscriptions'), function (Blueprint $table) {
+                $table->bigIncrements('id');
```
danharrin: Can we ensure that UUIDs are used as the primary key of this table, and any other model setup work is done to support this?

danharrin (changes requested):

app/Filament/Plugins/HeadOnlyBrowserNotificationsPlugin.php
```diff
@@ -0,0 +1,71 @@
...
+        // resolves the panel via Panel::make() first), which already queued its
+        // own head meta tag and the prompt banner onto this same panel. Render
+        // hooks can only be appended, never removed, so both are cleared here
+        // before this plugin adds its own head meta tag hook.
+        Closure::bind(function () {
+            unset(
+                $this->renderHooks[PanelsRenderHook::HEAD_END][''],
+                $this->renderHooks[PanelsRenderHook::BODY_END][''],
+            );
+        }, $panel, Panel::class)();
```
danharrin: Can you please explain what the render hooks did that we are removing and why they don't fit the ticket goals?

danharrin (changes requested):

database/migrations/2026_08_25_130000_create_push_subscriptions_table.php
mittal-canyon: Good catch — we were relying on the vendor model's default (auto-incrementing int) since it doesn't ship with UUID support out of the box. We can extend it: create our own model subclass with HasUuids, point config('webpush.model') at it, and switch the migration to a UUID primary key — matching how the rest of the app's models work. will update accordingly and fix.

app/Filament/Plugins/HeadOnlyBrowserNotificationsPlugin.php
mittal-canyon: @danharrin BrowserNotificationsServiceProvider::boot() calls Panel::configureUsing() to unconditionally add its own plugin to every panel (guarded only by hasPlugin('browser-notifications')). That callback runs when Panel::make() is constructed — before our own panel() method runs — so by the time we register our plugin, the vendor's plugin has already called its register() and queued its render hooks (the VAPID meta tag and a floating "Enable notifications" popup) onto the panel. I checked Filament\Support\View\ViewManager and Panel's HasRenderHooks trait — there's no forget/remove API, hooks are append-only. So this plugin uses Closure::bind to reach into the panel's own queued hooks and clear the popup one before re-adding just the meta tag hook we still need.

The reason we want the popup gone: this ticket calls for a dedicated Notifications settings page for subscribing, so the floating popup would be a second, redundant entry point for the same action.

danharrin:
Desktop browser notifications require three environment variables:

```dotenv
VAPID_SUBJECT=mailto:notifications@aiding.app
VAPID_PUBLIC_KEY=<generated-public-key>
VAPID_PRIVATE_KEY=<generated-private-key>
```

Generate a new key pair once from an Aiding App container:

```bash
php artisan webpush:vapid --show
```

The command prints `VAPID_PUBLIC_KEY` and `VAPID_PRIVATE_KEY`. Store the private key in the DevOps secret store; the public key is not sensitive. Supply the same pair to every web and queue-worker instance in that environment.

Use a separate persistent pair for each deployed environment. Do not regenerate or rotate the keys during ordinary deployments because existing browser subscriptions are tied to the public key and users may need to subscribe again after rotation.

danharrin (dismissed):

Orrison (approved):

PR #1424, opened 2026-09-01
[AIDAPP-1449]: Create a new Request Communications report under Service Desk

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1449

### Technical Description

> Create a new Request Communications report under Service Desk

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/report/src/Filament/Pages/RequestCommunications.php
+++ b/app-modules/report/src/Filament/Pages/RequestCommunications.php
@@ -40,8 +40,9 @@ use AidingApp\Report\Enums\ReportAccessKey;
 use AidingApp\ServiceManagement\Enums\ServiceRequestEmailTemplateType;
 use AidingApp\ServiceManagement\Enums\ServiceRequestNotificationChannel;
 use AidingApp\ServiceManagement\Enums\ServiceRequestTypeEmailTemplateRole;
-use AidingApp\ServiceManagement\Filament\Blocks\ServiceRequestTypeEmailTemplateButtonBlock;
-use AidingApp\ServiceManagement\Filament\Blocks\SurveyResponseEmailTemplateTakeSurveyButtonBlock;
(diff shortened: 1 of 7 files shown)


app-modules/report/src/Filament/Pages/RequestCommunications.php
```diff
@@ -0,0 +1,373 @@
...
+                ->mergeTags(ServiceRequestTypeEmailTemplate::getMergeTags())
+                ->customBlocks([
+                    ServiceRequestTypeEmailTemplateButtonBlock::class,
+                    SurveyResponseEmailTemplateTakeSurveyButtonBlock::class,
+                ])
+                ->fileAttachmentsDisk('s3-public')
+                ->resizableImages()
+                ->disabled()
+                ->json()
+                ->columnSpanFull(),
```
Orrison: Would it be possible to centralize the form fields being used in the actual forms, but then call these extra methods to set / override it to make it work here?

app-modules/report/src/Filament/Pages/RequestCommunications.php
```diff
@@ -0,0 +1,373 @@
...
+                            ->orderBy('name')
+                            ->pluck('name', 'id')
+                            ->all())
+                        ->searchable()
+                        ->live()
+                        ->afterStateUpdated(function (): void {
+                            $this->refreshEmailTemplates();
+                            $this->refreshNotifications();
+                        })
+                        ->placeholder('Select a service request type'),
```
Orrison: I think there exists a service request type select component that displays them in the order of and under the "areas" they are in in the service request type edit area. Are we able to use that here instead?

I think we use it on like the Service Request creation flow for example

Orrison (changes requested):

Orrison (approved):

PR #1426, opened 2026-09-01
[AIDAPP-1440]: Systems have reported a performance issue for the milestone relation in pipeline entry Kanban board

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1440

### Technical Description

>Systems have reported a performance issue for the milestone relation in pipeline entry Kanban board

> NO

### Cleanup Tasks

> N/A
---

### DIFF
--- a/app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
+++ b/app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
@@ -767,10 +767,12 @@ it('eager loads the milestone relation instead of lazy loading it per card', fun
 
     $stageId = $pipeline->stages->first()->getKey();
 
-    PipelineEntry::factory()->count(3)->create([
-        'pipeline_stage_id' => $stageId,
-        'project_milestone_id' => ProjectMilestone::factory()->create(['project_id' => $project->getKey()]),
-    ]);
+$milestone = ProjectMilestone::factory()->create(['project_id' => $project->getKey()]);
+
+PipelineEntry::factory()->count(3)->create([
+    'pipeline_stage_id' => $stageId,
+    'project_milestone_id' => $milestone->getKey(),
+]);
 
     DB::enableQueryLog();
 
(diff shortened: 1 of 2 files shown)


app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -753,3 +754,38 @@
...
+
+    $milestone = ProjectMilestone::factory()->create(['project_id' => $project->getKey()]);
+
+    PipelineEntry::factory()->count(3)->create([
+        'pipeline_stage_id' => $stageId,
+        'project_milestone_id' => $milestone->getKey(),
+    ]);
+
+    DB::flushQueryLog();
+    DB::enableQueryLog();
```
payal-canyon: The current test checks the query log and matches raw SQL for `project_milestones`, which feels a little brittle. It also introduces a query-log pattern that isn’t used elsewhere in the codebase.

Could we assert this through Eloquent instead? `relationLoaded('milestone')` directly verifies that the relation was eager loaded without depending on SQL text or the active database connection. It also keeps the test consistent with the existing Eloquent-based assertions in this file.

A couple of small improvements in the proposed approach:

- `assertSee($milestone->title)` verifies the actual milestone is rendered, rather than the unconditional `Milestone:` label.
- `toHaveCount(3)` ensures the `every()` assertion doesn’t pass vacuously on an empty collection.
- This also makes the `DB` facade import unused, so it can be removed.

payal-canyon (changes requested):

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
```diff
@@ -753,3 +754,38 @@
...
+
+    try {
+        livewire(PipelineEntryKanban::class, ['pipeline' => $pipeline])
+            ->assertSee('Milestone:');
+
+        $milestoneQueries = collect(DB::getQueryLog())
+            ->filter(fn (array $query): bool => str_contains($query['query'], 'project_milestones'));
+    } finally {
+        DB::disableQueryLog();
```
payal-canyon: It looks like you haven’t pushed any changes yet. Maybe it was missed? Could you please check from your side? Thanks!

payal-canyon (changes requested):

app-modules/project/tests/Tenant/Livewire/PipelineEntryKanbanTest.php
mittal-canyon: not sure what happened,i think because of rebase issue push was not successful. Fixed it now

payal-canyon (approved):

PR #1427, opened 2026-09-02
[AIDAPP-1436]: Deprecate milestone status

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1436

### Technical Description

> Deprecate milestone status

> NO

### Cleanup Tasks

> N/A

---


app-modules/project/src/Filament/Resources/Projects/Pages/ManageMilestones.php
```diff
@@ -80,11 +79,6 @@ public function form(Schema $schema): Schema
                 Textarea::make('description')
```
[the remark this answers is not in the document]
mittal-canyon: Fixed!

app-modules/project/src/Filament/Actions/CreateProjectMilestoneAction.php
```diff
@@ -83,14 +81,6 @@ public static function formSchema(): array
             Textarea::make('description')
```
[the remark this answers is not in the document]
mittal-canyon: Fixed!

.cleanup-tasks/2026_09_02_project_milestone_status_removed.md
```diff
@@ -0,0 +1,15 @@
...
+---
+
+## Feature Flags
+
+- App\Features\ProjectMilestoneStatusRemovedFeature
+
+## Additional Cleanup
+
+- Search for `TODO: Cleanup Task (project-milestone-status-removed)` and follow the instructions at each site.
+- Delete `AidingApp\Project\Models\ProjectMilestoneStatus`, `ProjectMilestoneStatusPolicy`, `ProjectMilestoneStatusFactory`, `ProjectMilestoneStatusSeeder`, and the `ProjectMilestoneStatuses` Filament resource.
```
rebecca-canyon: I believe that these things should be removed in this ticket

.cleanup-tasks/2026_09_02_project_milestone_status_removed.md
```diff
@@ -0,0 +1,15 @@
...
+- Remove `status_id` from `ProjectMilestone::$fillable`, delete its `status()` relation, the `project_milestone_status` morph map entry in `ProjectServiceProvider`, and the `ProjectMilestoneStatus::class` entry in `tests/Tenant/Unit/legacy-v4-uuid-models.php`.
+- The `status_id` column and `project_milestone_statuses` table were already dropped by the flag's own migration; no further schema change is needed.
```
rebecca-canyon: Because this is not a cleanup related note, this can be removed from here

app-modules/project/tests/Tenant/Filament/Actions/CreateProjectMilestoneActionTest.php
```diff
@@ -0,0 +1,56 @@
...
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use AidingApp\Project\Filament\Actions\CreateProjectMilestoneAction;
+use App\Features\ProjectMilestoneStatusRemovedFeature;
+
+// Checks the schema array directly (no Livewire mount/render) so the assertion does not
+// depend on the `project_milestone_statuses` table, which the flag's migration has dropped.
```
rebecca-canyon: Migrations are run before tests are, so this test file can be written with the statuses already having been dropped in mind

rebecca-canyon (changes requested):

.cleanup-tasks/2026_09_02_project_milestone_status_removed.md
mittal-canyon: We can't delete those yet — they're still being used. When the flag is off (before a tenant's migration has run), the milestone form still calls ProjectMilestoneStatus to show the status field. If we delete it now, that page will error out for any tenant that hasn't been migrated yet. Once the migration's rolled out everywhere and the flag is always on, that old code path is dead — that's what next week's cleanup task removes.

I have added some details been discussed over teams to get clarification.
Thanks!

app-modules/project/tests/Tenant/Filament/Actions/CreateProjectMilestoneActionTest.php
```diff
@@ -0,0 +1,55 @@
...
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+use AidingApp\Project\Filament\Actions\CreateProjectMilestoneAction;
+use App\Features\ProjectMilestoneStatusRemovedFeature;
+
+// Checked directly against the schema array since the status table is already dropped by the time tests run.
```
rebecca-canyon, line 40: Hm, I think I would add a note to delete these tests in the cleanup task, as well. While there probably should be other tests for this action, the ones that currently exist are all related to the status- and after the cleanup is finished, that won't make sense any more.

.cleanup-tasks/2026_09_02_project_milestone_status_removed.md
rebecca-canyon: You bring up a good point. However, I think we can still delete the policy and the resource itself. Existing tenants will already have statuses that they can use in that moment between deployment and migrations running, and if a new tenant were to be created in that time they would still have the seeded values that they can use.

So keeping the model, factory, and seeder makes sense. But we can remove the CRUD actions now.

rebecca-canyon (changes requested):

.github/workflows/pull_requests.yml
rebecca-canyon, line 1: It looks like you picked up some changes from another PR while rebasing

rebecca-canyon (changes requested):

.github/workflows/pull_requests.yml
mittal-canyon: NO. Kevin has added this as was facing some cache issue, pr checks wasn;t passing

rebecca-canyon (approved):

PR #1434, opened 2026-09-07
[AIDAPP-1491]: Present an initial display of projects to contacts

nency-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1491

### Technical Description

- Adds backend support for displaying accessible projects in the Knowledge Management Portal, including project listing, details, progress, task visibility, permissions, licensing, and archive handling.
- Added a new Projects list page: paginated table with Name / Start Date / Target Date / an action to view, no search or filter.
- Added a new project detail page ("Project Work"): shows an info banner explaining that progress reflects all work including internal tasks not visible in the portal, a pipeline dropdown (auto-selected and hidden entirely when there's only one pipeline), and the project's tasks grouped by milestone with a progress badge per milestone, plus an "unassigned" group for tasks with no milestone.

> No

### Cleanup Tasks

> N/A

---


app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/GetPortalProjectsController.php
```diff
@@ -0,0 +1,109 @@
...
+                'projects.description',
+                'projects.start_date',
+                'projects.target_completion_date',
+            ])
+            ->addSelect([
+                'total_pipeline_entries_count' => (clone $entries)->selectRaw('count(*)'),
+                'complete_pipeline_entries_count' => (clone $entries)
+                    ->where('pipeline_stages.classification', PipelineStageClassification::Complete->value)
+                    ->selectRaw('count(*)'),
+            ])
```
Orrison: I am worried about all the custom eloquent calls we are doing here instead of relying on eloquent directly through a `Project`'s relationships.

Could we do `with` and/or `withCount` calls here instead of fully mapping out relationships and archive/soft delete mechanisms manually here?

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/GetPortalProjectsController.php
```diff
@@ -0,0 +1,109 @@
...
+            ->orderBy('projects.name')
+            ->paginate(10)
```
Orrison, line 71: I can't see it in the screenshots, but does this all work properly with the pagination set up on the Frontend? Can you please add some screenshots showing the working pagination, I think the test is proving that this is working. But it would be good to see it in the screenshots.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/ShowPortalProjectController.php
```diff
@@ -0,0 +1,188 @@
...
+use AidingApp\Project\Models\Scopes\VisibleToPortalContact;
+use App\Settings\LicenseSettings;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Collection;
+use Illuminate\Http\JsonResponse;
+use Symfony\Component\HttpFoundation\Response;
+
+class ShowPortalProjectController
+{
+    public function __invoke(string $project): JsonResponse
```
Orrison: Is there a way to handle this with proper route model binding? There may be a way to setup route model binding where you can apply additional scoping like `withoutArchived` and the tap of `VisibleToPortalContact`

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/ShowPortalProjectController.php
```diff
@@ -0,0 +1,188 @@
...
+            ->oldest()
+            ->get([
+                'id',
+                'name',
+                'pipeline_stage_id',
+                'project_milestone_id',
+                'is_visible_to_guests',
+                'start_date',
+                'due',
+            ]);
```
Orrison: Is it possible to handle this with a single eloquent query getting all the proper data needed via `with` and sub-queries?

Additionally, I think we will end up needed pagination here that works WITH the milestone grouping and stuff like that, just like it does in Filament. It is possible that a project would have way too many tasks to properly show on one page and/or possible crash memory here.

Orrison (changes requested):

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/GetPortalProjectsController.php
amit-canyon: Added.

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/ShowPortalProjectController.php
```diff
@@ -0,0 +1,272 @@
...
+
+        $entries = $selectedPipeline->entries()
+            ->withoutArchived()
+            ->where('is_visible_to_guests', true)
+            ->whereIn(
+                'pipeline_stage_id',
+                PipelineStage::query()
+                    ->withoutArchived()
+                    ->select('id'),
+            )
```
Orrison: This can and should be replaced with `whereHas('pipelineStage', fn ($query) => $query->withoutArchived())`

Orrison (changes requested):

app-modules/portal/src/Http/Controllers/KnowledgeManagementPortal/ShowPortalProjectController.php
amit-canyon: I have changed because getting lint error. let me revert it.

app-modules/portal/tests/Tenant/Http/Controllers/KnowledgeManagementPortal/GetPortalProjectsControllerTest.php
Orrison, line 1: Let's add a direct test here proving that a Project that the Contact has NOT been given access to does not appear. There is some implicit testing of this here. But it would be important to have a test directly asserting that.

Orrison (changes requested):

Orrison (approved):

PR #1436, opened 2026-09-08
[AIDAPP-1480]: Introduce the ability to create groups and to associate users with groups

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1480

### Technical Description

> Introduce the ability to create groups and to associate users with groups

> No

### Cleanup Tasks

> Yes, Feature Flag: `GroupManagementFeature`

---


app-modules/group/database/migrations/2026_09_07_121815_create_groups_table.php
```diff
@@ -0,0 +1,59 @@
...
+    public function up(): void
+    {
+        Schema::create('groups', function (Blueprint $table) {
+            $table->uuid('id')->primary();
+            $table->caseInsensitiveText('name');
+            $table->text('description')->nullable();
+            $table->foreignUuid('created_by_id')->nullable()->constrained('users')->nullOnDelete();
+            $table->timestamps();
+
+            $table->uniqueIndex('name');
```
Orrison: I'll add a note to the Model as well, but let's make this Model soft deletable and archivable. And then the `uniqueIndex` here should be a partial index on whether or not `deleted_at` is `null`.

app-modules/group/src/Filament/Resources/Groups/Schemas/GroupForm.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return $schema
+            ->components([
+                Section::make('Properties')
+                    ->schema([
+                        TextInput::make('name')
+                            ->required()
+                            ->string()
+                            ->maxLength(255)
+                            ->unique(ignoreRecord: true)
```
Orrison: This will need to handle the soft delete unique change when that happens.

app-modules/group/src/Filament/Resources/Groups/RelationManagers/UsersRelationManager.php
```diff
@@ -0,0 +1,78 @@
...
+                TextColumn::make('email')
+                    ->searchable()
+                    ->sortable(),
+            ])
+            ->headerActions([
+                AttachAction::make()
+                    ->label('Associate users')
+                    ->multiple()
+                    ->slideOver()
+                    ->preloadRecordSelect(),
```
Orrison: We need to make sure, and have tests for, that this doesn't display or allow the selection of any User with any of the admin roles defined in `app/Models/Authenticatable.php` unless `app.filter_admins_from_selection` is `false` in `config/app.php`

app-modules/group/src/Filament/Resources/Groups/Pages/ListGroups.php
```diff
@@ -0,0 +1,94 @@
...
+class ListGroups extends ListRecords
+{
+    protected static string $resource = GroupResource::class;
+
+    public function table(Table $table): Table
+    {
+        return $table
+            ->columns([
+                TextColumn::make('name')
+                    ->description(fn (Group $record): ?string => $record->description)
```
Orrison: The `description` could be pretty long, so let's make sure we limit this to `40` characters then cut it off with `...`, I think Filament may already have something for that if not Laravel does.

Then make the tooltip of the description the full one so they can see it when they hover over it.

app-modules/group/src/Models/Group.php
```diff
@@ -0,0 +1,76 @@
...
+use App\Models\User;
+use Illuminate\Database\Eloquent\Attributes\ObservedBy;
+use Illuminate\Database\Eloquent\Concerns\HasUuids;
+use Illuminate\Database\Eloquent\Factories\HasFactory;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Database\Eloquent\Relations\BelongsTo;
+use Illuminate\Database\Eloquent\Relations\BelongsToMany;
+
+#[ObservedBy([GroupObserver::class])]
+class Group extends Model
```
Orrison: Let's make this model auditable / audited

Orrison (changes requested):

Orrison (approved):

PR #1437, opened 2026-09-08
[AIDAPP-1482]: Enhance the way status changes are editable in Manage Assignments and Assign to Me scenarios

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1482

### Technical Description

> Enhance the way status changes are editable in Manage Assignments and Assign to Me scenarios

> NO

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/service-management/src/Filament/Resources/ServiceRequests/Actions/ManageAssignmentAction.php
+++ b/app-modules/service-management/src/Filament/Resources/ServiceRequests/Actions/ManageAssignmentAction.php
@@ -37,7 +37,7 @@
 namespace AidingApp\ServiceManagement\Filament\Resources\ServiceRequests\Actions;
 
 use AidingApp\ServiceManagement\Enums\ServiceRequestAssignmentStatus;
-use AidingApp\ServiceManagement\Filament\Resources\ServiceRequests\Schemas\Components\ServiceRequestStatusSelect;
+use AidingApp\ServiceManagement\Filament\Resources\ServiceRequests\Schemas\Components\ServiceRequestStatusToggleButtons;
 use AidingApp\ServiceManagement\Filament\Tables\ManagersTable;
 use AidingApp\ServiceManagement\Models\ServiceRequest;
 use Filament\Actions\Action;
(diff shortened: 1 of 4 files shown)


app-modules/service-management/src/Filament/Resources/ServiceRequests/Actions/ManageAssignmentAction.php
```diff
@@ -57,7 +57,11 @@ public static function make(ServiceRequest $serviceRequest, string $actionName =
             ->slideOver()
             ->modalHeading('Manage Assignment')
             ->modalSubmitActionLabel('Submit')
-            ->fillForm(['userId' => $defaultUserId])
+            ->fillForm([
+                'userId' => $defaultUserId,
+                // fillForm() disables default-state hydration for the whole schema, so status_id must be filled explicitly to be pre-selected.
```
Orrison: Rather than add and keep a comment here I would rather us not have the comment and instead just make sure we have a test that ensures that the proper status is pre-selected on load.

I know we have a test making sure the component can handle having an external default set, but it would be good to have a test specifically making sure that the component loaded in this page loads correctly with the correct value.

Orrison (changes requested):

app-modules/service-management/src/Filament/Resources/ServiceRequests/Actions/ManageAssignmentAction.php
mittal-canyon: Sure, Let me check and will fix it!

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceRequests/Pages/AssignedToRelationManagerTest.php
```diff
@@ -412,6 +412,45 @@
...
+    livewire(AssignedToRelationManager::class, [
+        'ownerRecord' => $serviceRequest,
+        'pageClass' => ViewServiceRequest::class,
+    ])
+        ->mountTableAction('manageAssignment')
+        ->setTableActionData(['userId' => $manager->getKey()])
+        ->callMountedTableAction()
+        ->assertHasNoTableActionErrors();
+
+    expect($serviceRequest->refresh()->status_id)->toBe($status->getKey());
```
Orrison: This doesn't really specifically test that the current status is pre-filled in this actions form. I believe there are ways to test what is set in a form field for an action like this. We should be testing that instead of submitting the form then asserting on a model value that may or may not have been affected by it.

Orrison (changes requested):

Orrison (approved):

PR #1438, opened 2026-09-08
[AIDAPP-1272]: When navigating between portal and administrator, the session expires

payal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1272

### Technical Description

> The admin panel and the portal share one session, and therefore one CSRF token. Signing a contact in rotated that token, but Filament renders its token into the page rather than re-reading it - so any open admin panel page was left holding a token the session no longer recognised, surfacing as a 419 `page has expired`.
 
> All four contact sign-in paths (portal login, portal registration, Employee Self-Service, and the AI assistant widget) now go through a single `AuthenticatePortalContact` action that preserves the CSRF token across the sign-in. The session ID is still rotated, so session-fixation protection is unaffected.

> No

### Cleanup Tasks

> N/A

---


Orrison (changes requested):
@mxm1070 you were selected as the auto reviewer for this one, which is fine. But I would like to do a final review of this once you happy with it. So please re-request my review once you have reached approval on yours.

Orrison (approved):

Orrison (changes requested):
Holding off on merging this one as it seems it is technically most of the time no longer a problem in L13 with https://github.com/laravel/framework/pull/58400

But that it reveals a deeper issue we may want to fix a different way.

Orrison (approved):

PR #1441, opened 2026-09-09
[AIDAPP-1493]: Introduce an SLA report in analytics

monali4-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1493

### Technical Description

- Implemented `SlaComplianceOverTimeLineChart` for visualizing SLA compliance over the last 12 months.
- Created `SlaPerformanceByAgentTable` to display SLA performance metrics by agent, including export functionality.
- Developed SlaStats widget to summarize SLA metrics, including total requests and breach percentages.
- Added tests for SLA reporting features, ensuring access control and correct calculations based on filters.
- Introduced `chartTickAffix.js` to enhance chart label formatting with prefixes and suffixes.
- Updated `panel.js` to include the new `chartTickAffix` plugin for consistent chart rendering.

No

### Cleanup Tasks

No

---


app-modules/report/src/Filament/Widgets/SlaBreachesByServiceRequestTypeTable.php
```diff
@@ -0,0 +1,208 @@
...
+                'Response Compliance',
+                'Resolution Compliance',
+            ]);
+
+            foreach ($rows as $row) {
+                fputcsv($handle, [
+                    $row['type'],
+                    $row['sla_requests'],
+                    $row['response_breaches'],
+                    $row['resolution_breaches'],
```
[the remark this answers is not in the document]
monali4-canyon: Current inputs are expected not to include formula-prefixed values so not required

app-modules/report/src/Filament/Widgets/SlaPerformanceByAgentTable.php
```diff
@@ -0,0 +1,218 @@
...
+                'Avg Response Time',
+                'Avg Resolution Time',
+            ]);
+
+            foreach ($rows as $row) {
+                fputcsv($handle, [
+                    $row['agent'],
+                    $row['requests'],
+                    "{$row['responseSla']}%",
+                    "{$row['resolutionSla']}%",
```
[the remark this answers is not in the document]
monali4-canyon: Current inputs are expected not to include formula-prefixed values so not required

app-modules/report/src/Filament/Widgets/Concerns/InteractsWithSlaReportData.php
```diff
@@ -0,0 +1,231 @@
...
+            'sla_total' => $slaTotal,
+            'response_sla_total' => $responseSlaTotal,
+            'resolution_sla_total' => $resolutionSlaTotal,
+            'response_breaches' => $responseBreaches,
+            'resolution_breaches' => $resolutionBreaches,
+            'breaches' => $breaches,
+            'response_breach_percentage' => $this->slaPercentage($responseBreaches, $responseSlaTotal),
+            'resolution_breach_percentage' => $this->slaPercentage($resolutionBreaches, $resolutionSlaTotal),
+            'response_compliance_percentage' => $this->slaPercentage($responseSlaTotal - $responseBreaches, $responseSlaTotal),
+            'resolution_compliance_percentage' => $this->slaPercentage($resolutionSlaTotal - $resolutionBreaches, $resolutionSlaTotal),
```
Orrison, line 200: **Use all filtered requests as the percentage denominator (not just requests that have an SLA).**

All four SLA percentages currently divide by `responseSlaTotal` / `resolutionSlaTotal`, i.e. only requests whose type has an SLA configured. A request whose type has no SLA is dropped from the denominator, which inflates the breach and compliance percentages.

The core question this report answers is: *of the requests within these filters, what % have SLA issues?* A request with no SLA should count as "not breaching" while still being part of the whole. So 3 requests (1 breaching-with-SLA, 1 compliant-with-SLA, 1 with no SLA) should read **33.3%** breaches, not 50%.

Please divide all four by `$total`:

```php
'response_breach_percentage' => $this->slaPercentage($responseBreaches, $total),
'resolution_breach_percentage' => $this->slaPercentage($resolutionBreaches, $total),
'response_compliance_percentage' => $this->slaPercentage($total - $responseBreaches, $total),
'resolution_compliance_percentage' => $this->slaPercentage($total - $resolutionBreaches, $total),
```

This keeps breach% + compliance% = 100% and makes every widget (stats, line chart, both tables) tell one consistent story. Leave `sla_total` as-is — it's still a valid count for the type table's "SLA Requests" column.

Please also add test coverage for a filtered request whose type has **no** SLA. The current stats test uses 3 requests that all have SLAs, so it passes under both the old and new denominator and never exercises this path — add a no-SLA request and assert the percentage drops accordingly.

app-modules/report/src/Filament/Pages/Sla.php
```diff
@@ -0,0 +1,182 @@
...
+        return ReportAccessKey::fromPageClass(static::class)?->userCanAccess($user) ?? false;
+    }
+
+    public function filtersForm(Schema $schema): Schema
+    {
+        return $schema->components([
+            Section::make('Filters')
+                ->schema([
+                    ...$this->serviceRequestTypeFilterComponents(),
+                    DatePicker::make('startDate')
```
Orrison, line 107: **Perf: default the report to a bounded date window (e.g. last 12 months).**

With the date filters empty, `applySlaFilters()` applies no date constraint, so every widget loads **all** service requests for the tenant across all time and reduces them in PHP. Defaulting `startDate`/`endDate` in the filters form's default state bounds the dataset on first load — the biggest single win here.

Heads up: this interacts with the caching change requested on the tables below. Once default dates are always present, the existing `$shouldBypassCache = filled($startDate) || ...` logic would disable caching entirely, so pair this with filter-aware cache keys (see the two table comments) so the default view is still cached.

app-modules/report/src/Filament/Widgets/SlaPerformanceByAgentTable.php
```diff
@@ -0,0 +1,218 @@
...
+    /**
+        * @return Collection<int, mixed>
+     */
+    private function getAgentRows(): Collection
+    {
+        if ($this->agentRows !== null) {
+            return $this->agentRows;
+        }
+
+        $serviceRequests = $this->slaServiceRequestsQuery()
```
Orrison, line 186: **Perf: cache the computed rows here, keyed by the active filters.**

`getAgentRows()` only memoizes per request, so a full load + in-memory SLA reduction runs on every render (and again for the export). Cache the result like the stats/charts do — `Cache::tags(["{{$this->cacheTag}}"])->remember(...)` — but build the cache key from the active filter set (types, dates, classification, assignedAgents) so each filter combination (including the new default date window) is cached instead of recomputed. The `refresh-widgets` listener already clears the tag on manual refresh.

app-modules/report/src/Filament/Widgets/SlaBreachesByServiceRequestTypeTable.php
```diff
@@ -0,0 +1,208 @@
...
+    /**
+     * @return Collection<int, array{id: mixed, type: string, sla_requests: int, response_breaches: int, resolution_breaches: int, response_compliance: float, resolution_compliance: float}>
+     */
+    private function getTypeRows(): Collection
+    {
+        if ($this->typeRows !== null) {
+            return $this->typeRows;
+        }
+
+        $serviceRequests = $this->slaServiceRequestsQuery()
```
Orrison, line 181: **Perf: same as the agent table — cache these rows keyed by the active filters.**

`getTypeRows()` re-runs a full load + SLA reduction on every render and on export. Wrap it in `Cache::tags(["{{$this->cacheTag}}"])->remember(...)` with a key derived from the active filters (types, dates, classification, assignedAgents) so repeated/default filter combinations are served from cache.

app-modules/report/src/Filament/Widgets/Concerns/InteractsWithSlaReportData.php
```diff
@@ -0,0 +1,231 @@
...
+        return [
+            'status',
+            'priority.sla',
+            'priority.type',
+            'assignedTo.user',
+            'latestInboundServiceRequestUpdate.createdBy',
+            'latestOutboundServiceRequestUpdate.createdBy',
+            'statusPeriods',
+        ];
+    }
```
Orrison, line 71: **Perf: trim the loaded payload.**

Every SLA widget hydrates full `ServiceRequest` models plus these 7 relations for the entire filtered set. Consider narrowing the base query with `select()` to only the columns the SLA model methods actually read, and constraining the eager-loaded relations to the columns they need. Please verify the selected columns against what `getLatestResponseSeconds()` / `getResolutionSeconds()` / `getResolvedAt()` touch (e.g. `created_at`, `status_id`, `status_updated_at`, `updated_at`, `priority_id`, `category`) so nothing is silently dropped. Lighter rows meaningfully cut memory on large tenants.

Orrison (changes requested):

Orrison:
A few changes before this is ready. One correctness fix on how the SLA percentages are calculated, plus some performance wins (bounded default date window, filter-aware caching on the two tables, and trimming the loaded payload). Details inline.

Orrison (approved):

PR #1442, opened 2026-09-09
[AIDAPP-1481]: Enhance Settings > Service Request Types to be able to leverage Groups in Manager and Auditor assignements

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1481

### Technical Description

> Enhance Settings > Service Request Types to be able to leverage Groups in Manager and Auditor assignements

> No

### Cleanup Tasks

> Yes, Feature Flag: `ServiceRequestTypeGroupAssignmentsFeature`

---


app-modules/service-management/src/Models/Scopes/AccessibleServiceRequests.php
```diff
@@ -0,0 +1,57 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class AccessibleServiceRequests
+{
+    public function __construct(
+        protected User $user,
+    ) {}
+
+    /** @param Builder<covariant Model> $query */
```
Orrison: Instead of `covariant Model` can we please make this reference the Model it is actually used with, in this case I believe it is for `ServiceRequest`, right?

Let's make sure all the scopes added or modified in this branch do this please.

app-modules/service-management/src/Models/Scopes/AuditsServiceRequestType.php
```diff
@@ -0,0 +1,75 @@
...
+use AidingApp\ServiceManagement\Models\ServiceRequestTypeAuditorGroup;
+use App\Features\ServiceRequestTypeGroupAssignmentsFeature;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class AuditsServiceRequestType
+{
+    public function __construct(
+        protected string $serviceRequestTypeId,
+        protected ?ServiceRequest $serviceRequest = null,
```
Orrison: The `->when` usage of this `$serviceRequest` variable is not consistently applied to the groups `orWhereHas`.

But, after looking at this closer, I don't think this variable and the `$query->whereHas('serviceRequests',...`  that uses it are actually needed I don't think there is any logical or mathematic way this check does anything.
Do you agree? And if so, I think we should actually just remove it here and in `ManagesServiceRequestType`

Make sure you look into this and do agree with me though, I feel pretty confident that it doesn't actually do anything now given all the callers of these scopes, but we want to be sure.

Orrison (changes requested):
I have a couple of feedback comments and there is also a merge conflict to fix here atm.

app-modules/service-management/src/Models/Scopes/AuditsServiceRequestType.php
amit-canyon: Agreed. I've removed it; it was present in the old code but is no longer necessary.

app-modules/service-management/src/Filament/Resources/ServiceRequests/ServiceRequestResource.php
```diff
@@ -69,33 +70,17 @@ public static function getGloballySearchableAttributes(): array
...
-     * @return Builder<Model>
+     * @return Builder<ServiceRequest>
      */
     public static function getGlobalSearchEloquentQuery(): Builder
     {
         $user = auth()->user();
         assert($user instanceof User);
 
-        return parent::getGlobalSearchEloquentQuery()
+        return ServiceRequest::query()
```
Orrison: I think we should keep this a `parent::getGlobalSearchEloquentQuery()` if possible. It needs to make sure that any effects done to `parent::getGlobalSearchEloquentQuery()` are applied here as well.

Not technically an issue right now, but it could be in the future if we make an upstream change and don't notice this is not pulling that in.

app-modules/service-management/src/Models/Scopes/AccessibleServiceRequestTypes.php
```diff
@@ -0,0 +1,58 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class AccessibleServiceRequestTypes
+{
+    public function __construct(
+        protected User $user,
+    ) {}
+
+    /** @param Builder<covariant Model> $query */
```
Orrison: Please scope this to the specific model like the others.

app-modules/service-management/src/Models/Scopes/AuditedServiceRequestTypes.php
```diff
@@ -0,0 +1,60 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class AuditedServiceRequestTypes
+{
+    public function __construct(
+        protected User $user,
+    ) {}
+
+    /** @param Builder<covariant Model> $query */
```
Orrison: Please scope this to the specific model like the others.

app-modules/service-management/src/Models/Scopes/ManagedServiceRequestTypes.php
```diff
@@ -0,0 +1,60 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class ManagedServiceRequestTypes
+{
+    public function __construct(
+        protected User $user,
+    ) {}
+
+    /** @param Builder<covariant Model> $query */
```
Orrison: Please scope this to the specific model like the others.

app-modules/service-management/src/Models/Scopes/ManagedServiceRequests.php
```diff
@@ -0,0 +1,57 @@
...
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+
+class ManagedServiceRequests
+{
+    public function __construct(
+        protected User $user,
+    ) {}
+
+    /** @param Builder<covariant Model> $query */
```
Orrison: Please scope this to the specific model like the others.

Orrison (changes requested):

Orrison (approved):

PR #1447, opened 2026-09-10
[AIDAPP-1490]: Improve service monitoring reporting for more granularity regarding report delivery

mittal-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1490

### Technical Description

> Improve service monitoring reporting for more granularity regarding report delivery

> N/A

### Cleanup Tasks

> N/A
---

### DIFF
--- a/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
+++ b/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
@@ -13,7 +13,6 @@ created: 2026-09-08
 
 ## Additional Cleanup
 
-- Search for `TODO: Cleanup Task (service-monitoring-report-configurations)` and follow the instructions at each site.
 - Once every tenant has migrated, drop the legacy reporting columns (`is_reporting_active`, `report_frequency`, `is_reported_via_email`, `is_reported_via_database`) from `service_monitoring_targets`, and drop the legacy pivot tables `service_monitoring_target_report_user`, `service_monitoring_target_report_department`, `service_monitoring_target_report_contact` — along with the `ServiceMonitoringTarget` relations/casts/fillable entries and pivot models (`ServiceMonitoringTargetReportUser`, `ServiceMonitoringTargetReportDepartment`, `ServiceMonitoringTargetReportContact`) and their factories that reference them.
 - Delete the tests that cover the inactive (pre-migration) legacy-target branch, since that branch no longer exists once the flag is removed:
     - `app-modules/service-management/tests/Tenant/Jobs/ServiceMonitoringReportJobTest.php` — `it('still dispatches for confidential targets, since delivery suppression happens per recipient')` should keep asserting against `ServiceMonitoringReportConfiguration`, but delete `it('falls back to legacy targets when the feature is inactive')` and `it('does not dispatch for legacy targets with inactive reporting when the feature is inactive')`.
--- a/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
+++ b/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
@@ -13,8 +13,6 @@ created: 2026-09-08
 
 ## Additional Cleanup
 
-- Once every tenant has migrated, drop the legacy reporting columns (`is_reporting_active`, `report_frequency`, `is_reported_via_email`, `is_reported_via_database`) from `service_monitoring_targets`, and drop the legacy pivot tables `service_monitoring_target_report_user`, `service_monitoring_target_report_department`, `service_monitoring_target_report_contact` — along with the `ServiceMonitoringTarget` relations/casts/fillable entries and pivot models (`ServiceMonitoringTargetReportUser`, `ServiceMonitoringTargetReportDepartment`, `ServiceMonitoringTargetReportContact`) and their factories that reference them.
-- Delete the tests that cover the inactive (pre-migration) legacy-target branch, since that branch no longer exists once the flag is removed:
-    - `app-modules/service-management/tests/Tenant/Jobs/ServiceMonitoringReportJobTest.php` — `it('still dispatches for confidential targets, since delivery suppression happens per recipient')` should keep asserting against `ServiceMonitoringReportConfiguration`, but delete `it('falls back to legacy targets when the feature is inactive')` and `it('does not dispatch for legacy targets with inactive reporting when the feature is inactive')`.
-    - Each also needs its `use App\Features\ServiceMonitoringReportConfigurationsFeature;` import removed if no longer referenced.
+- Once every tenant has migrated, remove the legacy reporting columns, pivot tables, and model compatibility that are no longer needed.
+- Remove all inactive pre-migration compatibility tests after the legacy branch is removed.
 - Delete this file (`.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md`) once all the above cleanup is complete.
--- a/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
+++ b/.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
@@ -1,0 +1,21 @@
+---
+title: Service Monitoring Report Configurations Feature
+created: 2026-09-08
+---
+
+## Feature Flags
+
+- App\Features\ServiceMonitoringReportConfigurationsFeature
+
+## Temporary Migrations
+
+- app-modules/service-management/database/migrations/2026_09_08_090400_tmp_backfill_service_monitoring_report_configurations.php
+
+## Additional Cleanup
+
+- Search for `TODO: Cleanup Task (service-monitoring-report-configurations)` and follow the instructions at each site.
+- Once every tenant has migrated, drop the legacy reporting columns (`is_reporting_active`, `report_frequency`, `is_reported_via_email`, `is_reported_via_database`) from `service_monitoring_targets`, and drop the legacy pivot tables `service_monitoring_target_report_user`, `service_monitoring_target_report_department`, `service_monitoring_target_report_contact` — along with the `ServiceMonitoringTarget` relations/casts/fillable entries and pivot models (`ServiceMonitoringTargetReportUser`, `ServiceMonitoringTargetReportDepartment`, `ServiceMonitoringTargetReportContact`) and their factories that reference them.
+- Delete the tests that cover the inactive (pre-migration) legacy-target branch, since that branch no longer exists once the flag is removed:
+    - `app-modules/service-management/tests/Tenant/Jobs/ServiceMonitoringReportJobTest.php` — `it('still dispatches for confidential targets, since delivery suppression happens per recipient')` should keep asserting against `ServiceMonitoringReportConfiguration`, but delete `it('falls back to legacy targets when the feature is inactive')` and `it('does not dispatch for legacy targets with inactive reporting when the feature is inactive')`.
+    - Each also needs its `use App\Features\ServiceMonitoringReportConfigurationsFeature;` import removed if no longer referenced.
+- Delete this file (`.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md`) once all the above cleanup is complete.
--- a/app-modules/service-management/src/Actions/SaveServiceMonitoringReportConfigurationsAction.php
+++ b/app-modules/service-management/src/Actions/SaveServiceMonitoringReportConfigurationsAction.php
@@ -1,0 +1,80 @@
+<?php
+
+/*
+<COPYRIGHT>
+
+    Copyright © 2016-2026, Canyon GBS Inc. All rights reserved.
+
+    Aiding App® is licensed under the Elastic License 2.0. For more details,
+    see <https://github.com/canyongbs/aidingapp/blob/main/LICENSE.>
+
+    Notice:
+
+    - You may not provide the software to third parties as a hosted or managed
+      service, where the service provides users with access to any substantial set of
+      the features or functionality of the software.
+    - You may not move, change, disable, or circumvent the license key functionality
+      in the software, and you may not remove or obscure any functionality in the
+      software that is protected by the license key.
+    - You may not alter, remove, or obscure any licensing, copyright, or other notices
+      of the licensor in the software. Any use of the licensor’s trademarks is subject
+      to applicable law.
+    - Canyon GBS Inc. respects the intellectual property rights of others and expects the
+      same in return. Canyon GBS® and Aiding App® are registered trademarks of
+      Canyon GBS Inc., and we are committed to enforcing and protecting our trademarks
+      vigorously.
+    - The software solution, including services, infrastructure, and code, is offered as a
+      Software as a Service (SaaS) by Canyon GBS Inc.
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Actions;
+
+use AidingApp\ServiceManagement\Enums\ServiceMonitoringReportFrequency;
+use AidingApp\ServiceManagement\Models\ServiceMonitoringTarget;
+
+class SaveServiceMonitoringReportConfigurationsAction
+{
+    /**
+     * @param array<string, array{is_active?: bool, report_users?: array<int, string>, report_departments?: array<int, string>, report_contacts?: array<int, string>, report_channels?: array<int, string>}> $reportConfigurationsData
+     */
+    public function __invoke(ServiceMonitoringTarget $serviceMonitoringTarget, array $reportConfigurationsData): void
+    {
+        foreach (ServiceMonitoringReportFrequency::cases() as $frequency) {
+            $frequencyData = $reportConfigurationsData[$frequency->value] ?? [];
+
+            $attributes = ['is_active' => (bool) ($frequencyData['is_active'] ?? false)];
+
+            // The Recipients section (channels + recipients) is hidden while inactive, so Filament
+            // omits its fields entirely: only touch them when present, to keep the prior selections
+            if (array_key_exists('report_channels', $frequencyData)) {
+                $channels = $frequencyData['report_channels'];
+                $attributes['is_reported_via_email'] = in_array('email', $channels, true);
+                $attributes['is_reported_via_database'] = in_array('database', $channels, true);
+            }
+
+            $configuration = $serviceMonitoringTarget->reportConfigurations()->updateOrCreate(
+                ['frequency' => $frequency],
+                $attributes,
+            );
+
+            if (array_key_exists('report_users', $frequencyData)) {
+                $configuration->reportUsers()->sync($frequencyData['report_users']);
+            }
+
+            if (array_key_exists('report_departments', $frequencyData)) {
+                $configuration->reportDepartments()->sync($frequencyData['report_departments']);
+            }
+
+            if (array_key_exists('report_contacts', $frequencyData)) {
+                $configuration->reportContacts()->sync($frequencyData['report_contacts']);
+            }
+        }
+    }
+}
--- a/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
+++ b/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
@@ -40,6 +40,7 @@ use AidingApp\Contact\Models\Contact;
 use AidingApp\Department\Models\Department;
 use AidingApp\ServiceManagement\Enums\ServiceMonitoringReportFrequency;
 use App\Filament\Forms\Components\UserSelect;
+use App\Models\Scopes\WithoutAnyAdmin;
 use App\Models\User;
 use Filament\Forms\Components\CheckboxList;
 use Filament\Forms\Components\Select;
@@ -67,7 +68,7 @@ class ReportFrequencySection
                             ->label('Users')
                             ->multiple()
                             ->preload()
-                            ->options(fn (): array => User::query()->orderBy('name')->pluck('name', 'id')->all()),
+                            ->options(fn (): array => User::query()->tap(new WithoutAnyAdmin())->orderBy('name')->pluck('name', 'id')->all()),
                         Select::make("{$prefix}.report_departments")
                             ->label('Departments')
                             ->multiple()
--- a/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
+++ b/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
@@ -68,25 +68,26 @@ class ReportFrequencySection
                             ->label('Users')
                             ->multiple()
                             ->preload()
-                            ->options(fn (): array => User::query()->tap(new WithoutAnyAdmin())->orderBy('name')->pluck('name', 'id')->all()),
+                            ->options(fn (): array => once(fn (): array => User::query()->tap(new WithoutAnyAdmin())->orderBy('name')->pluck('name', 'id')->all())),
                         Select::make("{$prefix}.report_departments")
                             ->label('Departments')
                             ->multiple()
                             ->preload()
                             ->searchable()
-                            ->options(fn (): array => Department::query()->orderBy('name')->pluck('name', 'id')->all()),
+                            ->options(fn (): array => once(fn (): array => Department::query()->orderBy('name')->pluck('name', 'id')->all())),
                         Select::make("{$prefix}.report_contacts")
                             ->label('Contacts')
                             ->multiple()
                             ->preload()
                             ->searchable()
-                            ->options(fn (): array => Contact::query()->orderBy('full_name')->pluck('full_name', 'id')->all()),
+                            ->options(fn (): array => once(fn (): array => Contact::query()->orderBy('full_name')->pluck('full_name', 'id')->all())),
                         CheckboxList::make("{$prefix}.report_channels")
                             ->label('Channels')
                             ->options([
                                 'email' => 'Email',
                                 'database' => 'Application',
                             ])
+                            ->required(fn (Get $get): bool => (bool) $get($isActiveField))
                             ->columnSpanFull(),
                     ])
                     ->columns(3)
--- a/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
+++ b/app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
@@ -36,72 +36,61 @@
 
 namespace AidingApp\ServiceManagement\Filament\Components;
 
+use AidingApp\Contact\Models\Contact;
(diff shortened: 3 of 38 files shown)


app-modules/service-management/src/Jobs/ServiceMonitoringReportJob.php
```diff
@@ -75,7 +79,33 @@ public function handle(): void
...
+    }
+
+    private function dispatchForConfigurations(): void
+    {
+        ServiceMonitoringReportConfiguration::query()
+            ->where('frequency', $this->frequency)
+            ->where('is_active', true)
+            // Report delivery is confidentiality-aware per recipient, so confidential targets must not be filtered out here
+            ->with(['serviceMonitoringTarget' => fn (Relation $query) => $query->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class)])
+            ->chunkById(100, function (Collection $configurations) {
```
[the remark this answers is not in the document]
mittal-canyon: Fixed

app-modules/service-management/src/Jobs/ServiceMonitoringReportNotifyJob.php
```diff
@@ -55,11 +57,13 @@ class ServiceMonitoringReportNotifyJob implements ShouldQueue, ShouldBeUnique
     use Queueable;
     use SerializesModels;
 
-    public function __construct(public ServiceMonitoringTarget $serviceMonitoringTarget) {}
+    public function __construct(public ServiceMonitoringTarget|ServiceMonitoringReportConfiguration $reportable) {}
```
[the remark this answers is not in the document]
mittal-canyon: Fixed

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -66,13 +69,17 @@ class ServiceMonitoringReportNotification extends BaseNotification implements Sh
      **/
     private ?array $statistics = null;
 
-    public function __construct(public ServiceMonitoringTarget $serviceMonitoringTarget, public string $channel) {}
+    public function __construct(public ServiceMonitoringTarget $serviceMonitoringTarget, public ServiceMonitoringReportFrequency $frequency, public string $channel) {}
```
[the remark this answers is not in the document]
mittal-canyon: Fixed

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ConfidentialitySection.php
```diff
@@ -68,6 +71,20 @@ public static function make(string $notifiedUsersField, string $notifiedDepartme
...
+                        ...(ServiceMonitoringReportConfigurationsFeature::active()
+                            ? collect(ServiceMonitoringReportFrequency::cases())
+                                ->filter(fn (ServiceMonitoringReportFrequency $frequency): bool => (bool) $get("report_configurations.{$frequency->value}.is_active"))
+                                ->map(fn (ServiceMonitoringReportFrequency $frequency): ServiceMonitorReportRecipientsMustHaveConfidentialAccess => new ServiceMonitorReportRecipientsMustHaveConfidentialAccess(
+                                    frequency: $frequency,
+                                    reportedUserIds: Arr::wrap($get("report_configurations.{$frequency->value}.report_users")),
+                                    reportedDepartmentIds: Arr::wrap($get("report_configurations.{$frequency->value}.report_departments")),
+                                    confidentialUserIds: Arr::wrap($get('confidentialUsers')),
+                                    confidentialDepartmentIds: Arr::wrap($get('confidentialDepartments')),
+                                    creatorId: $record?->getAttribute('created_by_id') ?? auth()->id(),
```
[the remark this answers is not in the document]
mittal-canyon: Fixed

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -66,13 +69,30 @@ class ServiceMonitoringReportNotification extends BaseNotification implements Sh
...
+     * Restore notifications queued by the previous release, which had no `frequency` property.
+     * The private cache properties are left at their default (null) and recomputed lazily.
+     *
+     * @param array<string, mixed> $values
+     */
+    public function __unserialize(array $values): void
+    {
+        $this->serviceMonitoringTarget = $values['serviceMonitoringTarget'];
+        $this->frequency = $values['frequency'] ?? ServiceMonitoringReportFrequency::Monthly;
+        $this->channel = $values['channel'];
```
payal-canyon: This prevents all queued report notifications from being restored in the worker, so Daily, Weekly, and Monthly reports are never delivered. The override skips Laravel’s `SerializesModels` restoration and assigns the `ModelIdentifier` directly to the typed `ServiceMonitoringTarget` property, causing a `TypeError`.

Queued notification jobs fail in the worker, with no emails, in-app notifications, or `failed_jobs` entries. The current test doesn’t catch this because it passes a real model directly to `__unserialize()`.

Could you please delegate to the parent restore first and only fill the missing `frequency`, matching `ServiceMonitoringReportNotifyJob::__unserialize()`? Please also replace the test with a real `unserialize(serialize($notification))` round trip so this is covered.

app-modules/service-management/src/Filament/Components/ReportFrequencySection.php
```diff
@@ -0,0 +1,98 @@
...
+                    ->live()
+                    ->default(false)
+                    ->columnSpanFull(),
+                Section::make('Recipients')
+                    ->schema([
+                        UserSelect::make("{$prefix}.report_users")
+                            ->label('Users')
+                            ->multiple()
+                            ->preload()
+                            ->options(fn (): array => once(fn (): array => User::query()->tap(new WithoutAnyAdmin())->orderBy('name')->pluck('name', 'id')->all())),
```
payal-canyon: This is the only `UserSelect` in the project using a fixed `->options()` list instead of `->relationship()`. That bypasses the existing logic that preserves selected users and respects `filter_admins_from_selection`.

As a result, an existing recipient who is later made an admin can make the Edit page fail validation because their ID is no longer in the options. It also ignores `USER_SELECT_FILTER_ADMINS=false` and breaks email searching.

Could we use the existing `relationship()` approach here, or otherwise preserve the selected IDs and existing admin-filter/search behaviour? An Edit test with an existing admin recipient would also help cover this.

app-modules/service-management/src/Jobs/ServiceMonitoringReportJob.php
```diff
@@ -75,7 +80,35 @@ public function handle(): void
...
+
+    private function dispatchForConfigurations(): void
+    {
+        ServiceMonitoringReportConfiguration::query()
+            ->where('frequency', $this->frequency)
+            ->where('is_active', true)
+            // A soft-deleted target leaves its configuration row behind; skip it rather than dispatch a notify job with no target
+            ->whereHas('serviceMonitoringTarget', fn (Builder $query) => $query->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class))
+            // Report delivery is confidentiality-aware per recipient, so confidential targets must not be filtered out here
+            ->with(['serviceMonitoringTarget' => fn (Relation $query) => $query->withoutGlobalScope(ServiceMonitoringTargetVisibilityScope::class)])
```
payal-canyon: `serviceMonitoringTarget` is eager loaded here but never used, as `ServiceMonitoringReportNotifyJob::handle()` fetches the target again with `find()`.

Could we remove the `with()` and its comment? `whereHas` already excludes soft-deleted targets, and removing it also avoids the extra relation query when the job is restored.

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceMonitorings/Pages/ViewServiceMonitoringTest.php
```diff
@@ -128,30 +133,32 @@
             ]
         )
         ->assertSee('Automated Reporting')
-        ->assertSee('Frequency')
-        ->assertSee($serviceMonitoringTarget->report_frequency->getLabel())
+        ->assertSee('Weekly Reporting')
```
payal-canyon, line 136: The `ReportFrequencyInfolistSection` entries use `state()`/ `visible()` closures, but the test only asserts `Weekly Reporting` is visible. Could we also assert the Daily and Monthly entries are hidden when Weekly is active? That’s the core behaviour this ticket changes.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ReportFrequencySection.php
```diff
@@ -0,0 +1,98 @@
...
+                            ->label('Channels')
+                            ->options([
+                                'email' => 'Email',
+                                'database' => 'Application',
+                            ])
+                            ->required(fn (Get $get): bool => (bool) $get($isActiveField))
+                            ->columnSpanFull(),
+                    ])
+                    ->columns(3)
+                    ->visible(fn (Get $get): bool => (bool) $get($isActiveField)),
```
payal-canyon, line 109: The Recipients section has a `visible()` closure, but there’s no visible/hidden assertion on the Create or Edit page. Could we add coverage for this behaviour?

app-modules/service-management/tests/Tenant/Jobs/ServiceMonitoringReportJobTest.php
```diff
@@ -134,3 +132,69 @@
...
+    ]);
+
+    $target->delete();
+
+    (new ServiceMonitoringReportJob(ServiceMonitoringReportFrequency::Daily))->handle();
+
+    Queue::assertNotPushed(ServiceMonitoringReportNotifyJob::class);
+});
+
+// The following tests cover the pre-migration path, kept only until ServiceMonitoringReportConfigurationsFeature is cleaned up
```
payal-canyon: Comments like `The following tests cover…` don’t add much here. Could we use a `describe()` block for grouping, or remove the comments altogether?

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ReportFrequencySection.php
```diff
@@ -0,0 +1,98 @@
...
+use App\Filament\Forms\Components\UserSelect;
+use App\Models\Scopes\WithoutAnyAdmin;
+use App\Models\User;
+use Filament\Forms\Components\CheckboxList;
+use Filament\Forms\Components\Select;
+use Filament\Forms\Components\Toggle;
+use Filament\Schemas\Components\Section;
+use Filament\Schemas\Components\Utilities\Get;
+
+class ReportFrequencySection
```
payal-canyon, line 52: Both new components are only used by the ServiceMonitoring pages. Could we move them to `Resources/ServiceMonitorings/Schemas/Components/` to keep them with the other resource-specific components, like `ConfidentialitySection`?

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -213,6 +232,23 @@ private function getStatistics(): array
         return $this->statistics;
     }
 
+    // Reports are queued and run without an authenticated user, so confidentiality must be re-checked per recipient
+    private function notifiableCanViewTarget(object $notifiable): bool
```
payal-canyon: `notifiableCanViewTarget()` is almost identical to the logic in `ServiceMonitoringNotification`. Could we extract this into a shared helper, similar to `RecipientsMustHaveConfidentialAccess`, so report and alert suppression stay in sync?

app-modules/service-management/src/Actions/SaveServiceMonitoringReportConfigurationsAction.php
```diff
@@ -0,0 +1,80 @@
...
+
+class SaveServiceMonitoringReportConfigurationsAction
+{
+    /**
+     * @param array<string, array{is_active?: bool, report_users?: array<int, string>, report_departments?: array<int, string>, report_contacts?: array<int, string>, report_channels?: array<int, string>}> $reportConfigurationsData
+     */
+    public function __invoke(ServiceMonitoringTarget $serviceMonitoringTarget, array $reportConfigurationsData): void
+    {
+        foreach (ServiceMonitoringReportFrequency::cases() as $frequency) {
+            $frequencyData = $reportConfigurationsData[$frequency->value] ?? [];
```
payal-canyon: `SaveServiceMonitoringReportConfigurationsAction` performs multiple `updateOrCreate` / `sync` operations, but the panel doesn’t enable `databaseTransactions()`. A failure partway through could leave some frequencies saved and others not.

Could we wrap the action in `DB::transaction()`, similar to the existing multi-write actions in other modules?

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Pages/ViewServiceMonitoring.php
```diff
@@ -113,37 +124,49 @@ public function infolist(Schema $schema): Schema
...
-                            ->visible(fn (ServiceMonitoringTarget $record): bool => $record->is_reporting_active)
-                            ->columns(3),
+                        ...(ServiceMonitoringReportConfigurationsFeature::active() ? [
+                            Section::make('Automated Reporting')
+                                ->schema([
+                                    ReportFrequencyInfolistSection::make(ServiceMonitoringReportFrequency::Daily),
+                                    ReportFrequencyInfolistSection::make(ServiceMonitoringReportFrequency::Weekly),
+                                    ReportFrequencyInfolistSection::make(ServiceMonitoringReportFrequency::Monthly),
+                                ])
+                                ->visible(fn (ServiceMonitoringTarget $record): bool => $record->reportConfigurations()->where('is_active', true)->exists())
```
payal-canyon: `reportConfigurations` is already loaded here, so could we use `$record->reportConfigurations->contains('is_active', true)` instead of running another `exists()` query?

payal-canyon (changes requested):

app-modules/service-management/tests/Tenant/Notifications/ServiceMonitoringReportNotificationTest.php
```diff
@@ -171,6 +178,70 @@
...
+    $notification = new ServiceMonitoringReportNotification($target, ServiceMonitoringReportFrequency::Daily, MailChannel::class);
+
+    expect($notification->via($contact))->toBe(['mail']);
+});
+
+it('restores a notification serialized under the previous payload shape with no frequency property', function () {
+    $target = ServiceMonitoringTarget::factory()->create();
+
+    // Mirrors the payload shape queued by the previous release: no `frequency` key, and the target
+    // serialized as a full model rather than a ModelIdentifier, since that class had no SerializesModels
```
payal-canyon: The test for old queued payloads builds the target as a full model. The comment on line 222 says the previous release “had no `SerializesModels`”, but `Illuminate\Notifications\Notification` has always used it, so queued payloads contain a `ModelIdentifier`.

Could we build the values from the real serializer, like `ServiceMonitoringReportNotifyJobTest.php:277-288`?

- Start with `$notification->__serialize()`, then `unset($values['frequency'])` to mimic the old payload.
- Restore it and assert the target, frequency, and channel.
- Add a plain `unserialize(serialize($notification))` round trip for the current payload.
- Correct or remove the comment.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ReportFrequencySection.php
```diff
@@ -0,0 +1,113 @@
...
+                    ->default(false)
+                    ->columnSpanFull(),
+                Section::make('Recipients')
+                    ->schema([
+                        UserSelect::make("{$prefix}.report_users")
+                            ->label('Users')
+                            ->multiple()
+                            ->preload()
+                            ->options(function (UserSelect $component): array {
+                                $alreadySelected = array_map('strval', array_filter((array) $component->getState()));
```
payal-canyon: The new `Users` options closure fixes the admin-recipient validation failure, but it isn’t covered by any tests. The `user` picker on these pages already covers the same cases ( `EditServiceMonitoringTest.php` , `CreateServiceMonitoringTest.php`):

- Admins are filtered out.
- An already-selected admin remains in the list.
- All users are listed when `filter_admins_from_selection` is `false`.

Could we add the same three cases for `report_configurations.daily.report_users`? Since these options aren’t relationship-based, `getSearchResults()` returns `[];` could we assert `$field->getOptions()` inside `assertFormFieldExists()` instead?

Could we also add an Edit test where an existing Daily report user is an admin and saving still succeeds?

app-modules/service-management/tests/Tenant/Notifications/ServiceMonitoringReportNotificationTest.php
```diff
@@ -171,6 +178,70 @@
...
+    $values = [
```
payal-canyon: this test manually builds a payload with `serviceMonitoringTarget` as a full model, but real queued notifications use `SerializesModels` via `Illuminate\Notifications\Notification`, so the payload contains a `ModelIdentifier`. The comment on line 223 is also incorrect.

Because the test uses a full model, it still passes with the old broken implementation and doesn't cover the actual serialization path.

Could we build the payload using the real serializer, similar to `ServiceMonitoringReportNotifyJobTest? Start from `$notification->__serialize()`, `unset($values['frequency'])`, restore, and assert. It would also be good to add a `unserialize(serialize($notification))` test and update the comment.

This would ensure the test covers the actual queued-notification serialization path and catches this regression.

app-modules/service-management/src/Filament/Resources/ServiceMonitorings/Schemas/Components/ReportFrequencySection.php
payal-canyon, line 73: There’s currently no test coverage combining `report_users` with an admin, role assignment, or `filter_admins_from_selection`. The `user` picker on the same pages covers all three cases (`EditServiceMonitoringTest.php`, `CreateServiceMonitoringTest.php`).

Could we add the same three cases for `report_configurations.daily.report_users`, plus an Edit test where an existing Daily report user is an admin and saving succeeds? Since these options aren't relationship-based, `getSearchResults()` returns `[]`; we can assert the options using `$field->getOptions()` inside `assertFormFieldExists()`.

This would ensure the backfill doesn't introduce admin recipients that break the Edit page.

.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
```diff
@@ -0,0 +1,18 @@
...
+
+## Temporary Migrations
+
+- app-modules/service-management/database/migrations/2026_09_08_090400_tmp_backfill_service_monitoring_report_configurations.php
+
+## Additional Cleanup
+
+- Once every tenant has migrated, remove the legacy reporting columns, pivot tables, and model compatibility that are no longer needed.
+- Remove all inactive pre-migration compatibility tests after the legacy branch is removed.
+- Delete this file (`.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md`) once all the above cleanup is complete.
```
payal-canyon, line 53: Could you please update the cleanup file with the exact cleanup and the files that need to be changed? That way, when the feature-flag cleanup ticket is assigned to a developer, they can easily understand what needs to be updated or removed and where.

app-modules/service-management/src/Rules/RecipientsMustHaveConfidentialAccess.php
```diff
@@ -0,0 +1,157 @@
...
+     */
+    public function validate(string $attribute, mixed $value, Closure $fail): void
+    {
+        if (! $value) {
+            return;
+        }
+
+        $unreachable = $this->unreachableDepartmentNames()
+            ->concat($this->unreachableUserNames())
+            ->concat($this->unreachableContactNames());
```
payal-canyon: The confidentiality error currently merges department, user, and contact names into one unlabelled list ( `unreachableDepartmentNames()->concat(unreachableUserNames())->concat(unreachableContactNames())`). Since contacts are now included, it's no longer clear which picker the recipient came from.

For example, the Daily user was “Dr. Lucio Bailey” while the Daily contact was “Bailey Skiles”. The similar names made it easy to check the wrong picker.

Could we label each non-empty group? For example:

> …so they would never receive its automated report: Users: Dr. Lucio Bailey. Grant them confidential access below, or remove them from the Daily reporting settings.

This should make it clearer which recipient type needs to be updated.

payal-canyon (changes requested):

app-modules/service-management/src/Notifications/ServiceMonitoringReportNotification.php
```diff
@@ -66,13 +72,29 @@ class ServiceMonitoringReportNotification extends BaseNotification implements Sh
...
+    /**
+     * Restore notifications queued by the previous release, which had no `frequency` property
+     *
+     * @param array<string, mixed> $values
+     */
+    public function __unserialize(array $values): void
+    {
+        $values['frequency'] ??= ServiceMonitoringReportFrequency::Monthly;
+
+        $this->unserializeModels($values);
```
payal-canyon, line 87: The `__unserialize()` shim currently defaults legacy queued notifications to `Monthly`:

`$values['frequency'] ??= ServiceMonitoringReportFrequency::Monthly;`

The previous release derived the frequency from the target, so a legacy notification for a Daily/Weekly target can now use the wrong reporting window and statistics.

Could we restore the models first and then fall back to the target's frequency?
```php
$this->unserializeModels($values);

if (! isset($this->frequency)) {
    $this->frequency = $this->serviceMonitoringTarget->report_frequency
        ?? ServiceMonitoringReportFrequency::Monthly;
}
```
Also, this method depends on `report_frequency`, which will be removed as part of the feature-flag cleanup. Could you please check whether the entire `__unserialize()` method should be removed during cleanup, or if only part of it needs to change, and document the expected cleanup in the cleanup file? Otherwise, it could silently fall back to `Monthly` after `report_frequency` is removed.

payal-canyon (changes requested):

.cleanup-tasks/2026_09_08_service_monitoring_report_configurations_feature.md
```diff
@@ -0,0 +1,46 @@
...
+
+- Drop the legacy per-target report recipient pivot tables and their models:
+    - `service_monitoring_target_report_user` (`app-modules/service-management/database/migrations/2026_08_05_074634_create_service_monitoring_target_report_user_table.php`, model `app-modules/service-management/src/Models/ServiceMonitoringTargetReportUser.php`)
+    - `service_monitoring_target_report_department` (`app-modules/service-management/database/migrations/2026_08_05_074703_create_service_monitoring_target_report_department_table.php`, model `app-modules/service-management/src/Models/ServiceMonitoringTargetReportDepartment.php`)
+    - `service_monitoring_target_report_contact` (`app-modules/service-management/database/migrations/2026_08_05_074706_create_service_monitoring_target_report_contact_table.php`, model `app-modules/service-management/src/Models/ServiceMonitoringTargetReportContact.php`)
+
```
payal-canyon: A few cleanup items seem to be missing or incomplete:

1. **`ServiceMonitoringTarget`** - `$fillable` / `$casts` also contain `is_reported_via_database` and `is_reported_via_email`, so these should be removed along with `report_frequency` / `is_reporting_active`.

2. **`ServiceMonitoringTargetFactory`** - it sets all four legacy attributes (`is_reporting_active`, `report_frequency`, `is_reported_via_database`, `is_reported_via_email`), not just two.

3. **Legacy pivot factories** - these should also be removed when the three pivot models are deleted:

   * `database/factories/ServiceMonitoringTargetReportUserFactory.php`
   * `database/factories/ServiceMonitoringTargetReportDepartmentFactory.php`
   * `database/factories/ServiceMonitoringTargetReportContactFactory.php`

4. **`ServiceMonitoringReportNotifyJob`** - the cleanup should also cover the `__unserialize()` override and the `unserializeModels` alias. These only support legacy payloads and should be removed with the old `serviceMonitoringTarget` path.

payal-canyon, line 22: A few cleanup items seem to be missing or incomplete:

1. `ServiceMonitoringTarget` - `$fillable` / `$casts` also contain `is_reported_via_database` and `is_reported_via_email`, so these should be removed along with `report_frequency` / `is_reporting_active`.

2. `ServiceMonitoringTargetFactory` - it sets all four legacy attributes (`is_reporting_active`, `report_frequency`, `is_reported_via_database`, `is_reported_via_email`), not just two.

3. Legacy pivot factories - these should also be removed when the three pivot models are deleted:

   * `database/factories/ServiceMonitoringTargetReportUserFactory.php`
   * `database/factories/ServiceMonitoringTargetReportDepartmentFactory.php`
   * `database/factories/ServiceMonitoringTargetReportContactFactory.php`

4. `ServiceMonitoringReportNotifyJob` - the cleanup should also cover the `__unserialize()` override and the `unserializeModels` alias. These only support legacy payloads and should be removed with the old `serviceMonitoringTarget` path.

5. `ServiceMonitoringReportNotifyJobTest` - this isn't currently included in the cleanup list. There are eight tests still creating the job with a legacy `ServiceMonitoringTarget`.

Once `$reportable` is changed to `ServiceMonitoringReportConfiguration`, the first seven tests will need to be updated to use the configuration path instead. Similar tests already exist. The `__serialize()` shim test should be removed along with the shim.

app-modules/service-management/src/Concerns/ChecksServiceMonitoringTargetVisibility.php
```diff
@@ -0,0 +1,62 @@
...
+    - Use of this software implies agreement to the license terms and conditions as stated
+      in the Elastic License 2.0.
+
+    For more information or inquiries please visit our website at
+    <https://www.canyongbs.com> or contact us via email at legal@canyongbs.com.
+
+</COPYRIGHT>
+*/
+
+namespace AidingApp\ServiceManagement\Concerns;
```
payal-canyon: this is currently under `src/Concerns/`, which this branch introduced. Since the module already has `src/Notifications/Concerns/` ( `FetchServiceRequestTemplate`, `SetsServiceRequestEmailHeaders`) and only the two notifications use this trait, would it make more sense to place it there? Cosmetic only; the implementation looks correct.

app-modules/service-management/tests/Tenant/Filament/Resources/ServiceMonitorings/Pages/ViewServiceMonitoringTest.php
```diff
@@ -128,30 +133,34 @@
...
```
payal-canyon, line 136: It has no `assertSchemaStateSet` calls, even though each `ReportFrequencyInfolistSection` entry resolves its value through a `state()` closure. The section visibility is covered, but the actual values aren't. Could you please add assertions for these values as well?

payal-canyon (changes requested):

payal-canyon (approved):

PR #1450, opened 2026-09-11
[AIDAPP-1483]: Replace the Service Request edit experience with individual field edits

jayushi-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1483

### Technical Description

> Introduce individual field edits for service request.
> Remove page level edits.
> Introduce the ability to upload, delete files.
> Ensure that all the changes of service request changes are tracked properly in timeline tab

> No

### Cleanup Tasks

> N/A

---


mxm1070 (changes requested):
Overall this looks good to me. Four things I'd like changed before merge, and one smaller one.

## 1. A file between 10,000,001 and 10,485,760 bytes returns a 500

The form's size bound and the collection's disagree. `ServiceRequestMediaTable.php:170` sets `maxSize($maxFileSizeInMb * 1024)`, and Filament's `maxSize` is in kilobytes, so that permits 10,485,760 bytes. `UploadsMediaCollection` accepts `<= $mb * 1000 * 1000`, or 10,000,000. A file in between passes validation and then throws `FileUnacceptableForCollection` out of the `addMediaFromDisk(...)` chain at `:199-202`, uncaught, so the user gets an error page rather than a message.

Livewire's temporary-upload rule doesn't catch it first — that defaults to `max:12288`, or 12MB, and there's no `config/livewire.php` overriding it.

The staged S3 object is also left behind, because the media library deletes the source key only on success.

Bounding the form to the collection and handling the rejection covers both:

```php
$fileUpload->maxSize((int) floor($maxFileSizeInMb * 1000 * 1000 / 1024));

try {
    $media = $this->record
        ->addMediaFromDisk($path, 's3')
        ->usingName(pathinfo($path, PATHINFO_FILENAME))
        ->usingFileName(basename($path))
        ->toMediaCollection($collectionName);
} catch (FileCannotBeAdded $exception) {
    Storage::disk('s3')->delete($path);

    Notification::make()
        ->title('That file could not be uploaded.')
        ->body($exception->getMessage())
        ->danger()
        ->send();

    return;
}
```

A 10,200,000-byte upload gets a field error and leaves the bucket empty with this applied, and an uncaught `FileUnacceptableForCollection` without it.

The arithmetic can't be exact, since the rule is whole kilobytes — `floor()` lands 640 bytes under the collection limit. Erring under avoids reopening the gap. The catch is worth keeping independently of the bound, because a rejected mime type throws the same exception from the same place.

## 2. Pending uploads are written to the S3 bucket root under the submitted filename

`FileUpload::make('file')->disk('s3')->preserveFilenames()` has no `->directory()`, so the staged key is the submitted filename at the bucket root, unsanitised and not uniquified. Two people uploading `resume.pdf` at the same time share one key, and whichever add succeeds first deletes it.

A per-upload prefix fixes it without affecting the stored filename:

```php
->directory(fn (): string => 'tmp/' . Str::uuid())
```

Each upload then gets its own prefix rather than competing for one key at the root, and the media record still comes out as `resume.pdf`.

Worth raising as a question rather than a request: this is the only raw `Filament\Forms\Components\FileUpload` writing to S3 in `app-modules/`, and the other admin uploads — including the page this replaces, on this same collection — use `SpatieMediaLibraryFileUpload`. That component stages nothing on the target disk at all, since it adds the file with `addMediaFromString($file->get())`, so switching would remove this point entirely rather than patch it. Was there a reason to go the other way here? Point 1 stands either way, since that component's `toMediaCollection(...)` call isn't wrapped either.

## 3. A file attached to an update creates a timeline entry on the service request

`RecordServiceRequestFileUploadHistory` and `RecordServiceRequestFileDeletionHistory` resolve a `ServiceRequestUpdate` to its parent and write the history row there. Since `service_request_histories` has no reference to the update, the entry is indistinguishable from a file uploaded to the request itself — while the request's own Files tab shows nothing.

Whether a file on an update should appear on the request's timeline at all, and how it would be distinguished if so, reads like a Product call rather than something to settle in this PR. I'd suggest not shipping the behaviour in this increment and raising it as its own ticket. Both actions are new on this branch, and the widget recorded no history before it, so removing this doesn't regress anything.

Dropping the `ServiceRequestUpdate` arm from the `match` in both actions, plus the import, is enough — each `match` already has `default => null`, which the existing `if (! $serviceRequest) { return; }` guard handles. Two tests then fail, both asserting the behaviour being removed: `ServiceRequestMediaTableTest.php:805` and `:828`. Those want inverting to assert no history row, rather than deleting.

## 4. A soft-deleted status or priority leaves the field uneditable

A request whose status or priority has been soft-deleted shows an empty field with no edit action: the `belongsTo` resolves to null, and a `TextEntry` with blank state returns before it renders suffix actions. The value is gone and there's no way to set a new one from the record.

The page this replaces handled exactly this — `whereKey($record->status_id)->orWhereNull('deleted_at')` for status, and `HideDeletedExceptSelectedFromSelectOptions` for priority — so the record's own value stayed selectable once trashed. The toggle-button components list live records only, so that parity is lost.

Both delete policies refuse to remove a status or priority a request still references, so this is hard to reach going forward. Restoring what the previous page did still seems worth it, since it's a small change and the failure mode is a record nobody can fix from its own page.

## 5. `ServiceRequestFeedbackObserver` has no `deleted()` (minor)

`ServiceRequestUpdateObserver` and `ServiceRequestHistoryObserver` both dispatch `TimelineableRecordDeleted` on delete, which removes the `Timeline` row and clears the sync cache key. Feedback soft-deletes and is a timeline model, but nothing deletes one through Eloquent today, so this is future-proofing — worth adding while the three observers are otherwise symmetrical.

mxm1070 (approved):

PR #1451, opened 2026-09-11
[AIDAPP-1492]: Enhance the styling on project grouping to match

amit-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1492

### Technical Description

> Enhance the styling on project grouping to match

> No

### Cleanup Tasks

> N/A

---

### DIFF
--- a/app-modules/project/resources/views/filament/tables/groups/milestone.blade.php
+++ b/app-modules/project/resources/views/filament/tables/groups/milestone.blade.php
@@ -44,17 +44,19 @@
     CSS selector in resources/css/filament/admin/theme.css.
 --}}
 <span class="flex w-full items-center justify-between gap-x-3">
-    @if ($canManageMilestone)
+    @if ($milestone && $canManageMilestone)
         <x-filament::link
             tag="button"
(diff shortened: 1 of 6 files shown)


mxm1070:
I pushed a one-line change to `theme.css`. The group heading rule now uses the global `.fi-ta-group-heading` selector again. It is no longer scoped to `.project-work-pipeline-widget`. The global scope is intentional. The gray must apply to every table group heading in the app, not only to this widget.

I checked the following before I pushed:

- Only one other grouped table exists. It is the table select in the **Associate** slide-over on the SLA edit page. It has the same `gray-50` header background, so the contrast is 4.62 and still passes AA.
- The pipeline widget does not change in light mode or dark mode. The rule that hides the heading still works.
- `prettier --check` passes.

The rest looks good. The progress tracker for unassociated tasks is clean, and the grays are correct. The contrast is 4.62 in light mode and 5.93 in dark mode.

mxm1070 (approved):

PR #1457, opened 2026-09-14
[AIDAPP-1479]: Deprecate division impacts on service requests

rebecca-canyon:
### Ticket(s) or GitHub Issue

- https://canyongbs.atlassian.net/browse/AIDAPP-1479

### Technical Description

Remove division from service requests

No

### Cleanup Tasks

.cleanup-tasks/2026_09_14_remove_division_from_service_request_histories.md

---


app-modules/service-management/database/migrations/2026_09_14_194558_drop_division_id_from_service_requests_table.php
```diff
@@ -0,0 +1,55 @@
...
+use Illuminate\Database\Migrations\Migration;
+use Tpetry\PostgresqlEnhanced\Schema\Blueprint;
+use Tpetry\PostgresqlEnhanced\Support\Facades\Schema;
+
+return new class () extends Migration {
+    public function up(): void
+    {
+        Schema::table('service_requests', function (Blueprint $table) {
+            $table->dropConstrainedForeignId('division_id');
+        });
```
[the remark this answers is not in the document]
rebecca-canyon: The jobs that affect service requests won't be affected by this change

payal-canyon (changes requested):
In `DepartmentsTable.php`, the `division.name` column is still referencing the `division()` relationship:

It looks like the last remaining division reference in the service-management module. Could we remove this column as part of this work?

app-modules/service-management/database/migrations/2026_09_14_220000_tmp_remove_division_from_service_request_histories.php
```diff
@@ -0,0 +1,83 @@
...
+
+                            continue;
+                        }
+
+                        DB::table('service_request_histories')
+                            ->where('id', $history->id)
+                            ->update([
+                                'original_values' => json_encode($originalValues),
+                                'new_values' => json_encode($newValues),
+                            ]);
```
payal-canyon, line 77: `service_request_histories` are mirrored into `timelines`, but the migration deletes the history rows with `DB::table(...)`, so the Eloquent `deleted()` observer doesn't run. This leaves orphaned `timelines` rows, and `timeline.blade.php` can then hit `Call to a member function timeline() on null`, causing the entire View Service Request page to 500.

Could we delete the related `timelines` rows before deleting the histories, using the existing chunking pattern?

```php
if ($idsToDelete !== []) {
    DB::table('timelines')
        ->where('timelineable_type', 'service_request_history')
        ->whereIn('timelineable_id', $idsToDelete)
        ->delete();

    DB::table('service_request_histories')
        ->whereIn('id', $idsToDelete)
        ->delete();
}
```

`service_request_history` is the registered morph alias, so this should target the correct timeline records. This also keeps both deletes batched inside the existing transaction.

Could we also update the migration test to assert that no `timelines` rows remain for the deleted histories? The current test only verifies `service_request_histories`.

payal-canyon (changes requested):

app-modules/service-management/database/migrations/2026_09_14_220000_tmp_remove_division_from_service_request_histories.php
rebecca-canyon: Changes captured in ServiceRequestHistory rows can contain more than just one change, so deleting any row that contains a Division change could delete information we want to retain, which is why this migration doesn't follow the existing pattern exactly. But thank you for pointing out the timelines connection; I've updated the migration to include that.

app-modules/service-management/src/Filament/Resources/ServiceRequests/Pages/EditServiceRequest.php
```diff
@@ -0,0 +1,181 @@
...
+use Filament\Schemas\Components\Livewire;
+use Filament\Schemas\Components\Section;
+use Filament\Schemas\Components\Utilities\Get;
+use Filament\Schemas\Schema;
+use Illuminate\Database\Eloquent\Builder;
+use Illuminate\Database\Eloquent\Model;
+use Illuminate\Support\Collection;
+
+class EditServiceRequest extends EditRecord
+{
```
payal-canyon: `EditServiceRequest.php` and its `EditServiceRequestTest.php` are showing as new files against `main`. The edit page was previously removed when the edit experience was replaced with per-field actions, and nothing references it now ( `getPages()` has no `edit` route, and panel discovery only scans `app/Filament`). This looks like a modify/delete rebase resolution - should both files be deleted?

payal-canyon (changes requested):

payal-canyon (approved):

